PR #2, opened 2015-03-16
update extremes signatures to take prefetched token

thongsav-usgs:
Needed this signature to pass in token. We should leave it up to the end user to use the authenticateUser pm their own.

### DIFF
--- a/R/extremes.R
+++ b/R/extremes.R
@@ -28,7 +28,7 @@ setMethod("extremes", signature = c("character", "character", "character"),
           definition = function(data, output, token) {
             cat('this function uses the token to get data, and then calls the other one\n')
            
-            authHeader <- paste("Bearer" , token)
+            authHeader <- paste0("Bearer " , token)
             ts_list <- getJSON(url = data, auth = authHeader)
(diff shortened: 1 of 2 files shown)


R/extremes.R
```diff
@@ -22,13 +22,20 @@ setMethod("extremes", signature = c("list", "character"),
...
-          definition = function(data, output,...) {
-            cat('this function authenticates, goes out and gets data, and then calls the other one\n')
-            
-            token <- paste0('Bearer ', authenticateUser(...))
-            ts_list <- getJSON(url = data, auth = token)
+setMethod("extremes", signature = c("character", "character", "character"), 
+          definition = function(data, output, token) {
+            cat('this function uses the token to get data, and then calls the other one\n')
+           
+            authHeader <- paste("Bearer" , token)
```
jordansread: need to use `paste0` here, as the result of `paste('a','b')` is `[1] 'a', 'b'` not `'ab'`

jordansread: whoops, was wrong about this. Either would work.

PR #50, opened 2015-05-29
Added dynamically colored points based on approvals and dynamic legends

thongsav-usgs:


R/uvhydrograph-fnc.R
```diff
@@ -1,133 +1,131 @@
 
 #'@export
 uvhydrographPlot <- function(data){
-  uv_pts <- getUvHydro(data, "primarySeries" )
   layout_uvhydro()
   
+  #legend vector, needs to be dynamically built up with plots
+  primary_legend <- getNewLegendFrame()
+  addToPrimaryLegend <- function(newText, newSymbol, newColor, newLine) { 
+    primary_legend <<- rbind(primary_legend, data.frame(text = newText, symbol = newSymbol, color = newColor, line = newLine, stringsAsFactors = FALSE))
```
jiwalker-usgs: Looking into this a bit more, this is actually not as bad as I first thought.  This is an interesting pattern that might be useful elsewhere.

ldecicco-USGS: No, not too bad. 

Here's far more information than needed, if interested:

http://adv-r.had.co.nz/Functional-programming.html#closures
http://adv-r.had.co.nz/Environments.html
and finally:
http://adv-r.had.co.nz/OO-essentials.html#rc

thongsav-usgs:
Note that the dynamic paging is really slow right now. That for loop and the repeated subsetting of multiple data frames is likely the cause. Will optimize later on.

R/uvhydrograph-fnc.R
```diff
@@ -1,133 +1,141 @@
...
-  tertiary_lbl <- getUvLabel(data, "effectiveShifts")
-  
-  uv2_pts <- getUvHydro(data, "secondarySeries")
-  shift_pts <- getUvHydro(data, "effectiveShifts")
-  
-  secondary_lims <- getUvhLims(uv2_pts)
-  add_uvhydro_axes(secondary_lims, ylab = secondary_lbl, ylog = FALSE)
+  #for pagination by month, get all of the month strings for the primary series
+  all_primary_pts <- getUvHydro(data, "primarySeries" )
+  months <- unique(all_primary_pts$month, incomparables = FALSE)
```
jordansread, line 6: we assuming months to be continuous? Any issue if we miss a month? I.e., Jan and March data only? report still good?

R/uvhydrograph-fnc.R
```diff
@@ -1,133 +1,141 @@
...
-    # NOTE: this is a third plot, so this has to come at the end of this method.
-    #third axis
-    measuredShifts <- getFieldVisitErrorBarsShifts(data)
-    add_third_axes(secondary_lims = secondary_lims, tertiary_pts = shift_pts, tertiary_lbl = tertiary_lbl, measured_shift_pts = measuredShifts)
-    add_shift_measurements(measuredShifts)
+  for (month in months){  
+    #legend vector, needs to be dynamically built up with plots
+    primary_legend <- getNewLegendFrame()
+    addToPrimaryLegend <- function(newText, newSymbol, newColor, newLine) { 
+      primary_legend <<- rbind(primary_legend, data.frame(text = newText, symbol = newSymbol, color = newColor, line = newLine, stringsAsFactors = FALSE))
```
jordansread, line 14: why are you making these returns global?

jiwalker-usgs, line 14: That isn't actually what that does.  It works its way up in scope and assigns it when it finds it.  If it doesn't find that variable declared anywhere, it will then assign it to global.  This is the way to accumulate the legend in a more functional way.

R/uvhydrograph-fnc.R
```diff
@@ -1,133 +1,141 @@
...
+  points(pts$x, pts$y, type = 'l', col = 'orange', lty = 5)
+  addToLegend(paste("Estimated UV ", label), NA, "Orange", 5)
 }
 
-add_working_dv <- function(times, points){
-  pch = c(4, 15) # for x and box
-  col = 'yellow'
-  type = 'p'
+add_dv <- function(points, approvals, pch, label, addToLegend){
+  approvalColors = c("red", "yellow", "blue")
```
jordansread, line 103: personally, I prefer connecting these kinds of things explicitly, like 

``` R
approvals <- list(
   'working' = list('name' = 'Working", 'color' = "red"),
   'review' = list('name' = "In-review", color = 'yellow'))
```

or even

``` R
approvals = c("Working" = "red", "In-review"="yellow", "Approved"= "blue")
approvals[["In-review"]] # or
approvals[[2]]
names(approvals)[2]
```

jordansread, line 103: didn't realize you already have `approvals`, but point is the same.

R/uvhydrograph-fnc.R
```diff
@@ -157,11 +165,11 @@ combineLims<- function(lims1, lims2, ...) {
   return(list(xlim = xlim, ylim = ylim))
 }
 
-add_uvhydro_axes <- function(lims, ylog = TRUE, ylab){
+createPlot <- function(lims, ylog = TRUE, ylab) {
```
jordansread, line 168: kind of a generic name given creates a specific type of plot? or is this meant to be a generic plot set up? If the latter, should probably be exported and documented?

R/uvhydrograph-get.R
```diff
@@ -2,8 +2,18 @@
 getUvHydro <- function(ts, field){
   y <- ts[[field]]$points[['value']]
   x <- ts[[field]]$points[['time']]
-  time = as.POSIXct(strptime(x, "%FT%T"))
-  return(data.frame(x=time, y=y))
+  time <- as.POSIXct(strptime(x, "%FT%T"))
+  month <- format(time, format = "%y%m") #for subsetting later by month
+  return(data.frame(x=time, y=y, month=month, stringsAsFactors = FALSE))
```
jordansread, line 7: cool. I like the mm subsetting.

inst/uvhydrograph/uvhydrograph.Rmd
```diff
@@ -1,9 +1,12 @@
 ---
 output: landscape_document
```
jordansread, line 2: can this be in the yaml? or does it have to be here in the Rmd?

jordansread:
@thongsav-usgs looks good to me. I made some comments, but none are pressing. If you don't have responses, I can merge this if you want.

R/uvhydrograph-fnc.R
thongsav-usgs, line 6: they might not be continuous, but they will be in order, and missing a month should be OK

R/uvhydrograph-fnc.R
thongsav-usgs, line 14: Yeah, this is a "javascript pattern", trying to pass in a "closure" to the functions so that they can call a function which updates the variable in the environment above. Given what @jiwalker-usgs said above, and some discussion with @ldecicco-USGS we think this is actually not global.

R/uvhydrograph-fnc.R
thongsav-usgs, line 168: It's generic to the UV hydro (top/bottom plots), so I should rename it.

thongsav-usgs:
Go ahead and merge. This is still a WIP into the next Sprint so I'll change the little things then.

PR #51, opened 2015-06-01
Adding more detail for vdiagram report readme

mhines-usgs:
because I ran into this, figured it would be good to add.


thongsav-usgs:
I agree, I only knew to do it because I wrote the java that does it.

jordansread:
yep

PR #68, opened 2015-06-09
adjust vdiagram to show legend and not ugly page break

mhines-usgs:
adding new line to end of Rmd legend and removing it from the function itself AQCU-307

### DIFF
--- a/R/vdiagram-fnc.R
+++ b/R/vdiagram-fnc.R
@@ -199,7 +199,7 @@ pagingVdiagram <- function(rmd_dir, data, output, wd){
   rmdName <- 'vdiagram.Rmd'
   rmd_file <- file.path(rmd_dir, rmdName)
   
-  newPage = ifelse(output == "pdf", '$\\pagebreak$', '------')
+  #newPage = ifelse(output == "pdf", '$\\pagebreak$', '------')
   tempRmd <- tempfile(pattern = 'vdiagram', fileext = '.Rmd', tmpdir = wd)
   
   con <- file(rmd_file)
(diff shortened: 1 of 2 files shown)


R/vdiagram-fnc.R
```diff
@@ -219,7 +219,7 @@ pagingVdiagram <- function(rmd_dir, data, output, wd){
     pageText <- rawText
     pageText[pageText == replacePlot] <- sprintf('vdiagram(metaData[[%s]])', i)
     pageText[pageText == replaceTable] <- sprintf('vdiagramTable(metaData[[%s]], output)', i)
-    cat(c(pageText, newPage), file = tempRmd, sep = '\n', append = TRUE)
+    cat(c(pageText), file = tempRmd, sep = '\n', append = TRUE)
```
jordansread, line 222: this works?

inst/vdiagram/vdiagram.Rmd
```diff
@@ -12,4 +12,5 @@ vdiagram(data)
 vdiagramTable(data, output)
 ```
 Horizontal red bars at max and min gage height for the period shown.   
-Unlabeled blue points are historical measurements from the last 2 years.  
+Unlabeled blue points are historical measurements from the last 2 years.
+\
```
jordansread, line 16: gotcha

mhines-usgs:
@jread-usgs I wasn't sure this was the best way, but at least it doesn't seem to add a page break until after the legend for that report. Without this change, it was paginating weirdly, with the second reports station ID and name on top of page 2, and then a break, and the rest of the second report on the 3rd page.

PR #63, opened 2015-06-08
starting on rating diagram

jordansread:

PR #72, opened 2015-06-15
Revert "starting on rating diagram"

thongsav-usgs:
Reverts USGS-R/repgen#63


jordansread:
ugg https://travis-ci.org/USGS-R/repgen/builds/66932446#L1206

jordansread:
do you have to yank out the ratings work? Shouldn't conflict w/ UV

thongsav-usgs:
There was a commit that tried to pull out some generic plotting that has thrown off the ylims. It also appears the GW and QW plotting was lost somehow. 

We don't have to yank it out, but keeping it in reduces the amount of demo-able material.

thongsav-usgs:
let me take a look at this tomorrow. rather than yank everything out, I should probably just selectively revert UV hydro.

PR #76, opened 2015-07-10
small tweaks to grid and the legend

mhines-usgs:
Adding in two grids on the plots
Tweaking the legend so that it shows up closer to the graph

### DIFF
--- a/R/vdiagram.R
+++ b/R/vdiagram.R
@@ -88,7 +88,8 @@ plotVdiagram <- function(data){
   vplot <- gsplot() %>%
     points(NA,NA, ylab='Stage, in feet', xlab='Shift, in feet') %>%
     addMinMax(getMinStage(data, required = TRUE), getMaxStage(data, required = TRUE), col = 'red', lwd = 3) %>%
-    grid(lty = "solid") %>%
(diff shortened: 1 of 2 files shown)


R/vdiagram.R
```diff
@@ -88,7 +88,8 @@ plotVdiagram <- function(data){
   vplot <- gsplot() %>%
     points(NA,NA, ylab='Stage, in feet', xlab='Shift, in feet') %>%
     addMinMax(getMinStage(data, required = TRUE), getMaxStage(data, required = TRUE), col = 'red', lwd = 3) %>%
-    grid(lty = "solid") %>%
+    grid(lty = "dotted", nx=25) %>%
+    grid(lty = "solid", nx=5, col="black") %>%
```
jordansread, line 92: I don't know that this is robust against different y and x ticks. The ticks may be 5 divisions, which would line up fine, but they also may be more or less which would clash. Should probably tie grid to axis as default in gsplot

jordansread, line 92: BTW - we fixed the error bars, but not the callouts.

jordansread:
I will add `gsplot` to the travis scripts.

PR #79, opened 2015-07-14
unchanging pagination edits historically made

mhines-usgs:
and adding the (questionable?) right yaxis labels.


mhines-usgs:
The pagination is still wonky here, removing my hacky page breaks from the yaml.

PR #83, opened 2015-08-07
uvhydrograph using gsplot and new header (no image in header yet)

lindsayplatt:


mhines-usgs:
@ldecicco-USGS would you want to check this out sometime as a review? we're going to do a demo tomorrow and could suggest improvements moving forward though. There will be opportunities already where we are going to be making tweaks/additions to the plots. Thanks!

ldecicco-USGS:
Just a quick glance looks fine (no testing...just looking). 

I'll make an issue in `gsplot` for this: "expand y limits if error bars bleed over"...you shouldn't need to do that.

If there are other things you think you are doing that seem to be fixing `gsplot`...def put in an Issue.

If you want a more thorough code review, I could maybe do it later this week. 

It would be great to try to figure out how to upgrade your travis to their new workflow. I think if you dig into the travis documentation you should be able to get it to run. I didn't look at why Appveyor isn't building.

lindsayplatt:
@mhines-usgs I've committed the image and a few additional changes. It's not having a successful check on my computer -- saying that the "author" field is missing (which is a new value that needs to be used in the uvhydrograph call), but I don't see where it is missing.

mhines-usgs:
@lindsaycarr -- I'll try pulling this and testing..I'm free for a half hour.

ldecicco-USGS:
It probably means the author field in the DESCRIPTION file. I think they need to use the Authors@R object now. See:

https://github.com/USGS-R/gsplot/blob/master/DESCRIPTION#L6

mhines-usgs:
I had checked the Author field in Description and that's already there. I think it's somewhere else.. I'll look more shortly.. :/

lindsayplatt:
We have added a variable called `author` so that the report headers can have a users name on them, so it's treated as a variable and I don't think would be influenced by the DESCRIPTION file. 

See https://github.com/USGS-R/repgen/pull/83/files#diff-476d73c85e3d9ef8de5a74b15762dbc6R29

ldecicco-USGS:
Ah, OK. Just fyi...CRAN does prefer the Authors@R in description vs old fashion Author. I wouldn't be surprised if it became a warning/note for an as-cran check in the near term if it's not already.

lindsayplatt:
Test errors were because it wasn't able to follow the image path (even though it works locally...thought I had fixed it) and because of the Vdiagram `zero-length 'labels' specified` issue @mhines-usgs and I were having before. I added the image URL for a temporary fix for the report header, but we will still need to fix the vdiagram stuff. Should be good to go with this PR for now.

mhines-usgs:
Awesome!

PR #94, opened 2015-08-18
Adding median field

dpattermann-usgs:
This should add the functionality for a median field, I think.


mhines-usgs:
@dpattermann-usgs pulling this to check it out locally.

PR #104, opened 2015-08-28
Improve repgen for vdiagram

thongsav-usgs:
- Small progress on VDiagram code reorganization
- Reports now have unified reportMetadata structure


R/sig-vdiagram.R
```diff
@@ -54,19 +50,3 @@ setMethod("vdiagram", signature = c("character", "missing"),
           }
 )
 
-
-#'@aliases vdiagram
-#'@rdname vdiagram
-setMethod("vdiagram", signature = c("list", "missing"), 
-          definition = function(data) {
```
jordansread, line 61: Not sure why this signature is toast, it is the dev site. I.E. how I run things locally without output.

R/vdiagram-render.R
```diff
@@ -1,30 +1,59 @@
...
-  if (any(!is.na(obsShift)) && any(!histFlag)){
-    vplot <- callouts(vplot,x = obsShift[!histFlag], y = obsGage[!histFlag], labels=obsCallOut[!histFlag], cex=0.6)
+  nPages <- length(data$pages)
+  metaData <- vector(mode = 'list', length = nPages) #lol
+  # creates multi Rmd pages for output, truncates plots and tables. Returns metaData list globally, which would be nice to avoid
+  # should probably break this up into two calls. One that returns Rmd handle, the other w/ data
+  for (i in 1:nPages){
+    pageName <- names(data$pages)[i]
+    pageData <- data$pages[[pageName]]
+    metaData[[i]] <- pageData
```
jordansread, line 32: Can we avoid this is pattern unless needed? The <<- ?

R/vdiagram-styles.R
```diff
@@ -1,3 +1,22 @@
+vplot <- function(vdiagramData) {
+  vplot <- gsplot() %>%
+    points(NA,NA, ylab='Stage, in feet', xlab='Shift, in feet') %>%
+    callouts(x=c(0,0),y=c(getMinStage(data, required = TRUE), maxStage), labels="", col = 'red', lwd = 3, angle=0, legend.name="Max and min gage height for the period shown") %>%
```
jordansread, line 4: Would be nice to not do a get in here. Seems hidden relative to the new pattern

jordansread:
A few minor comments but :+1:

inst/extremes/extremes.Rmd
```diff
@@ -1,27 +1,27 @@
...
-# formTable <- padTable(tbl)
-# cat(formTable)
-kable(tbl)
-```
+
+![](http://www.usgs.gov/newsroom/images/2015-04-07/usgs_logo.png)
+
+## Extremes Report  
+
+**Location:**  `r getReportMetadata(ts,'locationNumber',required=TRUE)` `r getReportMetadata(ts,'locationName',required=TRUE)`
```
lindsayplatt, line 6: Why do we use getReportMetadata for the extremes, but not the other two? It's not a new feature, but I was wondering the same thing before these revisions. Should we be consistent? Especially, if we want to combine the header into one .Rmd file

R/vdiagram-render.R
thongsav-usgs, line 32: I was going to look into that. I think I can incorporate some java/javascript patters to isolate that a little.

inst/extremes/extremes.Rmd
thongsav-usgs, line 6: Yeah, I noticed the same too. It was a remnant of multiple people doing it simultaneously for 2 reports I believe. Consolidation/unification of both is definitely the goal.

PR #113, opened 2015-09-01
Round 2 of improvements

thongsav-usgs:


R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
 }
 
 createVdiagram <- function(data) {
-  data <- parseVDiagramData(data)
-  vplot(data)
+  styles <- getVDiagramStyle()
+  
+  vdiagramData <- parseVDiagramData(data)
```
jordansread, line 59: :+1:

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+  
+  vplot <- gsplot() %>%
+    points(NA,NA, ylab=styles$labels$y, xlab=styles$labels$x) %>%
```
jordansread: just a suggestion, but it might make sense to have the field names in `styles` match the arg names. So this would be `styles$ylab` instead of `styles$labels$y`. But I get that you have more things in `labels`

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+    callouts(y=c(vdiagramData$minStage, vdiagramData$maxStage), styles$callouts) %>%
```
jordansread, line 63: does `styles$callouts` unpack multiple arguments? Otherwise this is getting assigned to `x`, right? (since y is named and x is the first argument)

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+    grid(styles$grid) %>%
+    axis(styles$axis);
```
jordansread, line 65: same thing here. Are these expanded in gsplot or assumed to be the first arg? I would assume the former

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+  
+  vplot <- addMeasurementsAndError(vplot, vdiagramData, styles)
+  vplot <- addRatingShifts(vplot, vdiagramData, styles)
+  
+  if (any(!is.na(vdiagramData$obsShift)) && any(!vdiagramData$histFlag)){
```
jordansread: It would be nice if the return type of `histFlag` was such that we didn't need this logic, and instead were able to just call the following lines no matter what.

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+    vplot <- callouts(vplot,x = vdiagramData$obsShift[!vdiagramData$histFlag], y = vdiagramData$obsGage[!vdiagramData$histFlag], 
+                      labels=vdiagramData$obsCallOut[!vdiagramData$histFlag], cex=0.6)
+  }
+  
+  par(mar=c(7, 3, 4, 2))
```
jordansread: Not sure if y'all have tried this, but the way we hope to bake in the high level par calls in gsplot is through the initial call. So up on line 61, it would be 

``` R
vplot <- gsplot(mar=c(7, 3, 4, 2)) %>%
  # bla bla
```

jordansread: let us know if there is some reason not to do it that way

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+  print(vplot) 
+}
+
+addMeasurementsAndError <- function(vplot, vdiagramData, styles) {
+  histFlag <- vdiagramData$histFlag
+  if (any(histFlag)){
+    vplot <- do.call(arrows, append(list(object=vplot, x0=vdiagramData$minShift[histFlag], y0=vdiagramData$obsGage[histFlag],
```
jordansread, line 80: curious why `error_bars` isn't used here. Does `arrows` offer something needed in `error_bars`?

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+  }
+  
+  return(vplot)
+}
+
+echo <- function(string) {
+  print(string, quote=FALSE)
+}
+
+percentError <- function(MeasurementGrade) {
```
jordansread: belongs in `data`?

R/vdiagram-render.R
```diff
@@ -54,6 +54,129 @@ renderVDiagram <- function(data){
...
+                       col=ID), styles$rating_shift$to_segment))
+    }
+    
+    vplot <- do.call(lines, append(list(object=vplot, x=x, y=y, type="o", col=ID), styles$rating_shift$shift_segment))
+  }
+  
+  return(vplot)
+}
+
+echo <- function(string) {
```
jordansread: delete?

jordansread:
Some comments in there (and a borked travis build) to resolve or refute @thongsav-usgs or @mhines-usgs

R/vdiagram-render.R
thongsav-usgs: Actually I think it's just the two, in this case I think we should do that. I could change "labels" to "plotLabels" to keep it specific.

R/vdiagram-render.R
thongsav-usgs: That sounds good, is that already implemented in gsplot?

R/vdiagram-render.R
thongsav-usgs, line 80: I just had a hard time converting to error_bars and simply punted. I had thought I put a comment in there to eventually do it, must have been overzealous in my house cleaning.

R/vdiagram-render.R
thongsav-usgs: :+1:

R/vdiagram-render.R
thongsav-usgs, line 63: it unpacks

R/vdiagram-render.R
jordansread: yep @thongsav-usgs

R/vdiagram-render.R
jordansread, line 63: ok, then why do you need the do.call later on down? consider me confused...

R/vdiagram-render.R
thongsav-usgs, line 80: FYI, I could not get this to work using the standalone error_bar call. When I passed in via the error_par field in the points method, I DID get it to render but the bars did not appear correct.

Punting, but added comment in.

R/vdiagram-render.R
thongsav-usgs, line 63: the calls that don't use do.call aren't combining lists (they are explicitly pulling out single args into the method call).

thongsav-usgs, line 63: ACTUALLY.... I think the fix was naming all the parameters. So we probably DON'T have to do do.call... I will look into changing that and putting in another PR

thongsav-usgs, line 63: Gave it a try, but no go. When vplot is wrapped in a list, the list type causes the graphics version of the methods to invoke.

PR #118, opened 2015-09-04
Added qualifier prefixes to all values.

thongsav-usgs:


R/extremes-data.R
```diff
@@ -106,7 +108,61 @@ extremesTable <- function(data){
...
+      
+      if(! is.null(x$min$relatedDischarges)) {
+        x$min$relatedDischarges <- applyQualifiersToValues(x$min$relatedDischarges, consolidatedQualifiers$discharge)
+      }
+    }
+    return(x)
+  }))
+}
 
+applyQualifiersToValues <- function(points, qualifiers) {
```
jordansread, line 131: how about:

``` R
if(is.null(points)) return()
```

instead of doing an `!is.null` on those 4 lines above?

R/extremes-data.R
```diff
@@ -2,7 +2,9 @@
 #'@param ts a timeseries list that comes from valid extremes json
```
jordansread: do you need `#' @importFrom dplyr mutate` here?

thongsav-usgs: Will that prevent the need to call "library('dplyr')" ahead of time?

R/extremes-data.R
thongsav-usgs, line 131: good idea, will do

PR #126, opened 2015-09-10
UV hydro incorrect dimensions error fix

lindsayplatt:


R/uvhydrograph-render.R
```diff
@@ -45,11 +47,25 @@ createPrimaryPlot <- function(data, month){
...
+      }
+      primaryStyles[which(notPlot)] <- NULL
+      primaryPlotTypes <- primaryPlotTypes[which(!notPlot)]
+    } else {
+      primaryStyles <- getUvStyle(primaryData[i], primaryInfo, x, y, primaryApprovals, correctionLabels, "primary")
+      primaryPlotTypes <- getPlotType(primaryData[i], "primary")
+    }
+    
+    if (names(primaryData[i]) %in% c("series_corr", "meas_Q", "max_DV", "min_DV", "median_DV", "mean_DV", "UV_series")) {
+      for (j in seq(length(primaryStyles))) {
```
jordansread: I think you want `seq_len(length(primaryStyles))`, because seq(0) is `1 2` vs seq_len(0) is NULL (and won't enter the loop)

lindsayplatt: Ah good catch, I do this in a couple of places so I'll make sure to change it. Thanks!

lindsayplatt:
Don't merge this quite yet - submitting one more change based on Jordans comment

lindsayplatt:
Should be good to go now :+1:

PR #127, opened 2015-09-11
working with new versions. Need to fix axes=F

jordansread:
trying this out w/ new `gsplot` and `rmarkdown`

consider this a WIP


thongsav-usgs:
This has been open for awhile, is it still in play?

jordansread:
Kill it. This was migrating to the new version of `gsplot` and updating the dependencies.

PR #134, opened 2015-09-23
fixed squishy grids & tests shouldn't have errors now

lindsayplatt:


lindsayplatt:
Tests pass the checks locally for me, not sure why they are not doing the same here.

PR #137, opened 2015-09-25
adding sublocation to json and rmd

mhines-usgs:


inst/extdata/sitevisitpeak-example.json
```diff
@@ -51,6 +51,7 @@
     "altitudeDatumCode": "NGVD29    ",
     "siteNumber": "07105490       ",
     "stationName": "CHEYENNE CREEK AT EVANS AVE AT COLORADO SPRINGS,CO",
+    "sublocation": "Sub-Location",
```
mhines-usgs, line 54: @dpattermann-usgs is this right? (Minus the baloney example of Sub-Location...?)

dpattermann-usgs, line 54: Yeah, these should match up with the actual JSON that java puts out.

PR #147, opened 2015-10-06
adding handling for empty dataset to sitevisitpeak

mhines-usgs:
@thongsav-usgs working locally with the empty site visit peak example JSON file

### DIFF
--- a/R/sitevisitpeak-data.R
+++ b/R/sitevisitpeak-data.R
@@ -14,88 +14,100 @@
 #'@return string table
 #'@export
 #'
+
+
 sitevisitpeakTable <- function(data){
-  columnNames <- c("Date",
-                   "Time",
-                   "Party",
-                   "Sublocation",
-                   "Verification Method",
-                   "Reading",
-                   "Uncertainty",
-                   "Estimated Date",
-                   "Verification Comments",
-                   "Corrected Value",
-                   "Qualifier",
-                   "Date",
-                   "Time",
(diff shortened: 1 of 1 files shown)


mhines-usgs:
image

mhines-usgs:
image

R/sitevisitpeak-data.R
```diff
@@ -14,88 +14,100 @@
...
+                     "Date",
+                     "Time",
+                     "Difference from Peak Verification Reading")
+    
+    #Sends in list of readings, and gets pack the formatted data.frame
+    results <- formatSVPData(data$readings,columnNames)
+    
+    return(results)
+  }
+  if (length(data)==0) {
```
thongsav-usgs: else statement instead?

R/sitevisitpeak-data.R
```diff
@@ -14,88 +14,100 @@
...
-              nullMask(listElements$associatedIvValue),
-              quals,
-              ivDate,
-              ivTimeFormatting,
-              diff)
-    
-    toRet <- rbind(toRet, data.frame(t(toAdd),stringsAsFactors = FALSE))
+    return(toRet)
+  }  
+  if(length(data) == 0) {
```
thongsav-usgs: same comment, use else statement here

mhines-usgs: on else it complains and won't build. but else if works.. is that acceptable or should I see why the else alone is failing? @thongsav-usgs

PR #148, opened 2015-10-07
slight formatting adjustments for svp

mhines-usgs:
This isn't ideal but I'm not sure how much more time to spend on it.


mhines-usgs:
html version now looks kind of better... using a new package though. 

image

mhines-usgs:
image

PR #152, opened 2015-10-19
Sensor Reading Summary

mhines-usgs:
Not perfect (needs some rounding and formatting love) but functional


lindsayplatt:
:+1: renders for PDF and HTML for me!

Three things I notice (which you probably already know):
1. PDF is missing top-level column headers ("Field Visit Information", "Reference Readings", etc)
2. PDF has some column text overlapping (especially, "Date" and "Time")
3. HTML appears to get cutoff on the right side even though it actually does not

mhines-usgs:
Yes, the PDF rendering is (still) a problem. We have a ticket that should address this in the queue for the future, essentially ditching pander and writing raw latex out. That way we have granular control of things. I am not sure where that will shake out in the priorities yet though.

mhines-usgs:
Working on the HTML cutoff, I want to put a margin in or something but.. struggling to make the markdown behave with me.

R/sensorreading-data.R
```diff
@@ -0,0 +1,231 @@
...
+    
+    toRet <- rbind(toRet, data.frame(t(toAdd),stringsAsFactors = FALSE))
+  }
+  colnames(toRet) <- columnNames
+  rownames(toRet) <- NULL
+  
+  return(toRet)
+}
+
+nullMask <- function(val) {
```
lindsayplatt, line 130: What is this function doing?

mhines-usgs, line 130: @lindsaycarr  -- I carried that over from sitevisitpeak and didn't mess with a good thing. I believe it's supposed to accept all the values from the json and see if they're null and if not make the value the value, but if it is, make an empty string the value?

lindsayplatt, line 130: Ah, ok. I just hadn't seen it at all and was curious why it existed :)

inst/uvhydrograph/uvhydrograph.Rmd
```diff
@@ -28,6 +28,7 @@ output:
 ***
 
 ```{r echo=FALSE, warning=FALSE, message=FALSE, results='asis'}
+library(lubridate)
```
lindsayplatt: We should probably investigate more to see why this is needed. It shouldn't have to be explicitly stated here

mhines-usgs: ugh, sorry. Something must be funky with my environment. Before I merge I'll change that back (again..)

R/sensorreading-data.R
```diff
@@ -0,0 +1,231 @@
...
+  if(length(inQualifiers) < 1) return("");
+  q <- inQualifiers[[1]]
+  
+  if(is.null(q) || length(q) < 1) return("");
+  
+  qualifiers <- q[time>q$startDate & q$endDate>time,]
+  
+  builtQualifiers <- ""
+  if(nrow(qualifiers) > 0) {
+    for(i in 1:nrow(qualifiers)) {
```
lindsayplatt: I think instead of having both the if and for statement, you can just use `for(i in seq_len(nrow(qualifiers)))`. It will automatically not do anything in the loop since there is nothing to loop through. 

``` r
qualifiers <- data.frame(1:5)
qualifiers2 <- data.frame()

for(i in seq_len(nrow(qualifiers))){
  print("stuff")
}

for(i in seq_len(nrow(qualifiers2))){
  print("stuff")
}
```

inst/uvhydrograph/uvhydrograph.Rmd
lindsayplatt: Oh I forgot it was just you're environment -- so bizarre

R/sensorreading-data.R
mhines-usgs: Ah, ok. I'll have to update sitevisitpeak too with that..

lindsayplatt: Not a big deal - just something that saves a few lines and confusion with which { goes to which

Issue #45, opened 2015-04-29
Fix package check NOTEs and WARNINGS


jordansread:
Should be able to see when running `check` from RStudio

jordansread:
@mhines-usgs interested?

mhines-usgs:
From Travis CI
- [x] Undocumented code objects: ‘getSimsUrl’ ‘json’ ‘startUvhydrographRender’ ‘uvhydrographPlot’
  All user-level objects in a package should have documentation entries. See chapter ‘Writing R documentation files’ in the ‘Writing R Extensions’ manual.
- [ ] checking examples ... ERROR
  Running examples in ‘repgen-Ex.R’ failed
  The error most likely occurred in:
  
  > base::assign(".ptime", proc.time(), pos = "CheckExEnv")
  > 
  > ### Name: sensorreadingsummary
  > 
  > ### Title: sensorreadingsummary report
  > 
  > ### Aliases: sensorreadingsummary
  > 
  > ### sensorreadingsummary,character,character-method
  > 
  > ### sensorreadingsummary,list,character-method
  > 
  > ### sensorreadingsummary,list,missing-method
- [ ] Error in (function (toc = FALSE, toc_depth = 2, number_sections = FALSE,  : 
  unused argument (self_contained = FALSE)
  Calls: sensorreadingsummary ... render -> create_output_format -> do.call -> <Anonymous>
  Execution halted
- checking for unstated dependencies in ‘tests’ ... OK
- checking tests ...
  Running ‘testthat.R’ [31s/31s]
  ERROR
  Running the tests in ‘tests/testthat.R’ failed.
  Last 13 lines of output:
           pandoc = pandoc_options(to = "latex", from = from_rmarkdown(fig_caption, 
               md_extensions), args = args, keep_tex = keep_tex), clean_supporting = !keep_tex, 
           pre_processor = pre_processor, intermediates_generator = intermediates_generator)
   })(self_contained = FALSE)

**This error has to do with the newer rmarkdown package. Travis CI uses .8.1 or so and we are still using .6.1 -- when I upgrade locally to that version, I encounter the same message.**
- [ ]  testthat results ================================================================
  OK: 9 SKIPPED: 0 FAILED: 2
  1. Error: example data vdiagram 
  2. Error: examples work 
  
  Error: testthat unit tests failed

mhines-usgs:
From my local check:
- [x] checking installed package size ... NOTE
  installed size is 56.6Mb
[...]

PR #159, opened 2015-10-25
build check fixes

mhines-usgs:


mhines-usgs:
Even though Travis is failing, this cleans up all the warnings on my local machine. See https://github.com/USGS-R/repgen/issues/45 for other issues seen on the Travis build. One seems related to the rmarkdown version (we are staying at 0.6.1 for now and it is using 0.8.1) and the others are not happening locally. Will continue to investigate other Errors from Travis.

PR #165, opened 2015-11-10
Better HTML styling of tabular reports

mhines-usgs:


inst/sensorreading/sensorreading.Rmd
```diff
@@ -28,16 +30,12 @@ tbl <- sensorreadingTable(data)
...
-  } else {
-    htmlTable(tbl$toRet, 
-              cgroup=c("Field Visit Information","Reference Readings","Recorder Readings","Correction Information","Nearest IV"),
-              n.cgroup=c(4,4,4,4,4),
-              col.rgroup = c("none", "#F7F7F7"), css.cell = "padding-left: 1em; padding-right: 1em; padding-top:.5em;padding-bottom:.5em;",
-              align="|rrrr|rrrr|rrrr|rrrr|rrrr|",
-              rnames=FALSE)
-  }
-  pander(tbl$comments_table, split.cell=25, split.table=Inf)
+htmlTable(tbl$toRet,
```
thongsav-usgs, line 33: So this will break now if they ask for PDF?

mhines-usgs, line 33: Yes.. Because of the "pandoc error 67" black hole @lindsaycarr and I spent hours trying to figure out, which comes with the newer rmarkdown upgrade. I tried reverting back to 0.6.1 with the HTML changes I just made and I get pandoc error 1!

mhines-usgs, line 33: @thongsav-usgs upping the version of rmarkdown will cause ALL the pdf generation to fail with pandoc 43 or 67 or whichever. Should I hold off on this until we have that conversation?

PR #174, opened 2015-12-03
dvhydro improvements

mhines-usgs:
for axes and corrected json field name


mhines-usgs:
@lindsaycarr I can't figure out where the ugly $dvhplot etc are coming from, any ideas?

image

lindsayplatt:
Are you stopping at a breakpoint and then pressing continue?

mhines-usgs:
@lindsaycarr no, not really. the dvhydrographPlot calls each createPlot function for each of the plots, then returns the list of plots to the function call in the Rmd..

PR #183, opened 2015-12-16
Made extremes report more general, using report metadata to for

thongsav-usgs:
headings/labels now.


thongsav-usgs:
closed, expanding generalness to any stat derived timeseries (Max/min/median)

PR #187, opened 2015-12-16
Changed 5yr GW legend to match GW specifications

dpattermann-usgs:
Max is the smallest number, because numbers are being measured as distance below ground level (so the smallest number is the highest up).
Also, fixed fiveyrgwsum(data) so that it works.


dpattermann-usgs:
@lindsaycarr @thongsav-usgs Think this should do it. If you think I should change it lower down rather than directly in the legend, say so.

lindsayplatt:
I think directly in the legend is good

R/fiveyeargwsum-styles.R
```diff
@@ -5,9 +5,9 @@ getFiveyearStyle <- function(data) {
     styles <- switch(names(data), 
                      stat = list(lines = list(x=x, y=y, type="o", col="black", pch=20, cex=0.5, lwd=0.8, legend.name=data[[1]]$legend.name)),
                      est_stat = list(lines = list(x=x, y=y, col="red", lty=2, legend.name=data[[1]]$legend.name)),
-                     max_iv = list(points = list(x=x, y=y, col="red", pch=8, cex=2, legend.name=paste("Max. Instantaneous Water Level:", y)),
+                     max_iv = list(points = list(x=x, y=y, col="red", pch=8, cex=2, legend.name=paste("Min. Instantaneous (Lowest) Water Level:", y)),
```
lindsayplatt: @dpattermann-usgs could you have it say "Min. (Lowest) Instantaneous..." instead? Same for max - just move where the word in parentheses is?

dpattermann-usgs: Ah! That's a good idea. Will do.

PR #190, opened 2015-12-17
Modified uvHydro rmd so that it uses the full identifier in the header

dpattermann-usgs:
Also changed the header to use the function to access reportMetadata rather than direct access.:

### DIFF
--- a/inst/uvhydrograph/uvhydrograph.Rmd
+++ b/inst/uvhydrograph/uvhydrograph.Rmd
@@ -2,9 +2,9 @@
 
 ## `r data[['title']]`  
 
-**Location:**  `r getReportMetadata(data,stationId)` - `r getReportMetadata(data,stationName)`
+**Location:**  `r getReportMetadata(data,'stationId')` - `r getReportMetadata(data,'stationName')`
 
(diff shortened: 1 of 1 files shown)


inst/uvhydrograph/uvhydrograph.Rmd
```diff
@@ -2,13 +2,13 @@
 
 ## `r data[['title']]`  
 
-**Location:**  `r data[['reportMetadata']][['stationId']]` - `r data[['reportMetadata']][['stationName']]`
+**Location:**  `r getReportMetadata(data,stationId)` - `r getReportMetadata(data,stationName)`
```
thongsav-usgs: Tests are failing, I think you are supposed to pass in strings to the getReportMetadta function. EG: getReportMetadata(ts,'locationNumber',required=TRUE)

dpattermann-usgs: Weird. It ran locally. I can change it to pass in strings

dpattermann-usgs:
@thongsav-usgs This should fix it. Tests pass this time.

PR #194, opened 2015-12-21
Correction Information columns were empty.

dpattermann-usgs:
Methods were never executed because it looked for something in data, but data didn't exist. This wasn't noticed because data is a function so it didn't throw an "object ' ' not found" exception.
Also, updated the JSON file to use current data format and which has all data members.


dpattermann-usgs:
Travis CI keeps failing because the old JSON example file for SRS uses a format that we don't use anymore. I'll update the JSON and it should pass.

PR #206, opened 2015-12-29
Uv hydro profiling

dpattermann-usgs:
Using RProf(), I found out that datetime parsing was taking a while. Fasttime is very fast and helps handle this problem.


dpattermann-usgs:
I'm having trouble with Travis builds. It doesn't seem to want to install fasttime when testing.

thongsav-usgs:
@dpattermann-usgs 
I think you need to make sure you do an import before the function that uses it and also add it to the DESCRIPTION file. We've been using roxygen to update that automatically I think.

@ldecicco-USGS ?

R/uvhydrograph-data.R
```diff
@@ -226,8 +226,8 @@ getUvHydro <- function(ts, field, estimatedOnly = FALSE){
   x <- ts[[field]]$points[['time']]
   
   if(!is.null(y) & !is.null(x)){
-    
-    time <- as.POSIXct(strptime(x, "%FT%T"))
+
+    time <- fastPOSIXct(x,tz = "UTC")
```
ldecicco-USGS: Yeah, so up in the roxygen part of this file, add:
#' importsFrom fasttime fastPOSIXct
Then, make sure you redo the roxygen documentation, and then you should notice that:
importsFrom(fasttime, fastPOSIXct)
is in the NAMESPACE file

ldecicco-USGS:
Also, I haven't done any benchmarking, but since you are already using the `lubridate` package, you might want to check if the function `fast_strptime` is comparable (in case you don't want to depend on yet another package). It's definitely faster than base, but I haven't compared it to `fastPOSIXct`.

dpattermann-usgs:
@ldecicco-USGS I hadn't done any benchmarking for lubridates parse_date_time either. I'm comparing the two right now. Seems like lubridate is nearly just as fast. I chose fasttime because it's very lightweight (it only has the one function) and it's pretty simple to use. But if we're already using lubridate, I'll just use that. Thanks for the help Laura!

PR #208, opened 2015-12-29
AQCU-512, AQCU-511, and suggestions from 12/29 demo

lindsayplatt:


lindsayplatt:
*\* for error_bars to be fixed, https://github.com/USGS-R/gsplot/pull/297 should be merged

PR #210, opened 2016-01-04
Optimizing uvHydrograph and vdiagram

dpattermann-usgs:
Added in Lubridate parse_date_time again. Going to investigate more how timezones are assigned, looking at dev and changing system timezones.

### DIFF
--- a/R/uvhydrograph-data.R
+++ b/R/uvhydrograph-data.R
@@ -241,7 +241,8 @@ getUvHydro <- function(ts, field, estimatedOnly = FALSE){
   
   if(!is.null(y) & !is.null(x)){
 
-    time <- as.POSIXct(strptime(x, "%FT%T"))
+    format <- "Ymd HMOS z"
+    time <- parse_date_time(x,format,tz = "America/Chicago",quiet=TRUE)
     
     month <- format(time, format = "%y%m") #for subsetting later by month
     uv_series <- data.frame(x=time, y=y, month=month, stringsAsFactors = FALSE)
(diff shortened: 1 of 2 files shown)


R/uvhydrograph-data.R
```diff
@@ -241,7 +241,8 @@ getUvHydro <- function(ts, field, estimatedOnly = FALSE){
   
   if(!is.null(y) & !is.null(x)){
 
-    time <- as.POSIXct(strptime(x, "%FT%T"))
+    format <- "Ymd HMOS z"
+    time <- parse_date_time(x,format,tz = "America/Chicago",quiet=TRUE)
```
dpattermann-usgs, line 245: I'm considering subbing in Sys.timezone() for "America/Chicago" if the server is assigning timezones.

R/uvhydrograph-data.R
```diff
@@ -241,7 +241,8 @@ getUvHydro <- function(ts, field, estimatedOnly = FALSE){
   
   if(!is.null(y) & !is.null(x)){
 
-    time <- as.POSIXct(strptime(x, "%FT%T"))
+    format <- "Ymd HMOS z"
+    time <- parse_date_time(x,format,tz = Sys.timezone(),quiet=TRUE)
```
dpattermann-usgs: I'm fairly sure the location of the R server is how the data gets its UTC Offset, but want to test on dev.

R/uvhydrograph-data.R
thongsav-usgs, line 245: Don't think we want to set any time zone, the timezone has to be based on the data, not the system running the parsing.

PR #216, opened 2016-01-06
cleanup for shared functions + corrections-at-a-glance

lindsayplatt:
This initial work for corrections-at-a-glance is done. More work to come in a new PR!


R/utils-shared.R
```diff
@@ -0,0 +1,48 @@
...
+  
+  return(date_index)
+}
+
+
+############ used in dvhydrograph-data and correctionsataglance-data ############ 
+
+formatDates <- function(char_date){
+  as.POSIXct(strptime(char_date, "%FT%T"))
+}
```
dpattermann-usgs, line 48: Should we consider switching this to the lubridate function as we did for uvhydro? I'm wondering if dvhydro/5-year would see speed ups from it. Something to think about considering we have the function made up in uvhydro already.

lindsayplatt, line 48: Good idea! We'll keep in mind :smile: `lubridate::parse_date_time()`

lindsayplatt:
Going to merge this one and start a new PR, it'll be too much code to review if I don't

PR #218, opened 2016-01-08
more corrections-at-a-glance + dvhydro error fix

lindsayplatt:
Currently a work in progress!


R/correctionsataglance-data.R
```diff
@@ -2,83 +2,208 @@
 
 parseCorrectionsData <- function(data){
 
-  allDataRange <- c(formatDates(data$primarySeries$requestedStartTime), 
-                    formatDates(data$primarySeries$requestedEndTime))
+  dateData <- formatDateRange(data$primarySeries$requestedStartTime, data$primarySeries$requestedEndTime)
+  apprData <- formatDataList(data$primarySeries$approvals, 'APPROVALS', datesRange = dateData$dateSeq) #top bar = primary series approvals
+  fieldVisitData <- list(startDates = formatDates(data$fieldVisits$startTime),
+                         dataNum = length(data$fieldVisits$startTime)) #points on top bar = field visits
+  preproData <- formatDataList(data$corrections, "PRE_PROCESSING") #lane one = pre-processing
```
thongsav-usgs, line 9: I'm starting to think we should have given the JSON preparsed from the java side. The java service layer should insulate the rendering layer from changes to the data. Just a comment for now, should be a direct swap in later if we decide to do it.

lindsayplatt, line 9: :+1:

PR #219, opened 2016-01-11
DV/5yr report updates

thongsav-usgs:


R/fiveyeargwsum-styles.R
```diff
@@ -1,12 +1,20 @@
...
 
     x <- data[[1]]$time
     y <- data[[1]]$value
+    
+    #semantics for min/max are swapped on inverted plots
+    minMaxLabels = c("Max. Instantaneous", "Min. Instantaneous");
+    if(invertedPlot) {
+      minMaxLabels = c("Min. Instantaneous", "Max. Instantaneous");
+    }
+
```
lindsayplatt, line 7: I think we should keep data parsing/munging stuff in the -`data.R` script. Then, you can have L1-2 be:

```
getFiveyearStyle <- function(data, info=NULL, ...) {
     args <- list(...)
```

and then the max/min `legend.name = paste(args$minMaxLabels[1], info$type, ":", y)`

lindsayplatt, line 7: this would also go up in `dvhydrograph-data.R`

thongsav-usgs, line 7: I'm not sure I consider this data munging, it's strictly a labeling change. That is unless we consider the legend title to be part of the data, is that how we are treating it else where?

lindsayplatt, line 7: I guess that's just where I've been changing it so far, but yeah it's not really data munging bc it's the title https://github.com/USGS-R/repgen/blob/master/R/dvhydrograph-data.R#L92-L112

thongsav-usgs, line 7: For consistency, I'll switch it to the way you've been doing it.

thongsav-usgs, line 7: Actually, I can't completley use that pattern because the legend name needs to be mixed with y (measurement number), so I'll use the first pattern you posted.

lindsayplatt:
Ok, I'd say we are good to go on this now. Also, sorry I started a bad trend with the legend name stuff :sob:

thongsav-usgs:
Not at all, it's still in the "render" layer, so it still makes sense. Better to be consistent than not.

PR #226, opened 2016-01-23
Changes to guarantee we avoid any knitr collisions

thongsav-usgs:


thongsav-usgs:
@lindsaycarr may want to take a look at this too.

I'm still having the problem of having concurrent report generations overwrite each other but have done extensive testing and proved they collide before any rendering even occurs (at the environment level).

Merging now to test on dev, it might be an issue with Rserve on windows not supporting concurrent sessions. See https://rforge.net/Rserve/rserve-win.html

Testing if the knitr changes fixes on dev.

thongsav-usgs:
A lot of this was undone in https://github.com/USGS-R/repgen/pull/227

PR #229, opened 2016-01-26
axis flipping for uvhydro + most of truncated labels framework

lindsayplatt:
truncated labels commented out -- waiting for a gsplot fix (multiple callouts accepted)


lindsayplatt:
@thongsav-usgs this should move the uvhydro flip ticket along. I can't remote in, so I can't get to JIRA to do that - can you move it to test?

thongsav-usgs:
Will do! Thanks for squeezing this in during travel!

PR #234, opened 2016-02-04
truncating labels for uvhydro & vdiagram round 2

lindsayplatt:
updated how `width_char` is calculated. Example illustrating truncated labels from @thongsav-usgs  was because it thought that the label was getting cutoff on the right axis b/c the character width used was enormous, so it rotated the callout. Since it was near the left axis, it was truncated from the left. 

Example from PT: https://cida-eros-aqcudev.er.usgs.gov:8443/aqcu-front-end/service/reports/vdiagram/?primaryTimeseriesIdentifier=209c5e8130244dcc995a0b1321cd74eb&upchainTimeseriesIdentifier=925c02693cc247b4a00ac0d605ce4fc0&ratingModelIdentifier=Gage+height-Discharge.STGQ%4006934500&station=06934500&lastMonths=12

### DIFF
--- a/R/utils-shared.R
+++ b/R/utils-shared.R
@@ -79,7 +79,7 @@ testCallouts <- function(plot_obj, xlimits){
   
   png('TMP_PLOT')
   print(plot_obj)
-  width_char <- par("cxy")[1]*xrange
+  width_char <- par("cxy")[1]
   #When you're done
   dev.off()
   #Delete the plot you just generated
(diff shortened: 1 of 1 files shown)


lindsayplatt:
once checks pass this should be good to merge

thongsav-usgs:
:+1:

thongsav-usgs:
Looks like test coverage went down slightly.

lindsayplatt:
wow...0.2% and it throws a red X

PR #238, opened 2016-02-08
Added rounding to account for border comparison on calculated number

thongsav-usgs:
ranges.

### DIFF
--- a/R/sensorreading-data.R
+++ b/R/sensorreading-data.R
@@ -140,8 +140,8 @@ getRecorderWithinUncertainty <- function(uncertainty, value, recorderValue) {
     ref <- as.numeric(value)
     unc <- as.numeric(uncertainty)
     rec <- as.numeric(recorderValue)
-    val1 <- ref+unc
-    val2 <- ref-unc
+    val1 <- round(ref+unc, getSrsPrecision())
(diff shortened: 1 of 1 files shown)


thongsav-usgs:
@jiwalker-usgs and @lindsaycarr not a huge fan of this, but the other option was adding an error range to each calculation.

We were already using 2 as a known precision for these numbers (specified by the source) so I opted to just go with rounding as the quickest implementation.

lindsayplatt:
@thongsav-usgs I like that you made that a function so it's very easy to change. Although this might be the answer to your problem: https://cran.r-project.org/doc/FAQ/R-FAQ.html#Why-doesn_0027t-R-think-these-numbers-are-equal_003f

Use `all.equal()` and it has a built in tolerance. You can also adjust the tolerance using the `tolerance` argument

thongsav-usgs:
I don't think that includes less than or equal, but we could do that in the boolean logic itself.

lindsayplatt:
Could you do two separate statements?

```
if(x < y | all.equal(y,x)){
     #do something
}
```

thongsav-usgs:
This conversation is bleeding into #240

PR #240, opened 2016-02-09
When no uncertainty exists, evaluate if the two values are equal

thongsav-usgs:


lindsayplatt:
@thongsav-usgs did you see my last comment on the merged PR?

thongsav-usgs:
@lindsaycarr: yes, using an uncertainty number is slightly less appealing in that it's a little more arbitrary than the precision number of 2. Precision of 2 seems to have actual significance from the source data. Of course, using a small enough buffer is more than safe, do you have a preference?

lindsayplatt:
Nope, I just wanted to share something else that I had found and see if you thought it was a better method. No point in changing something that you've already fixed :)

PR #241, opened 2016-02-11
Moved comments to their own rows. Added some questionable logic to

thongsav-usgs:
exclude redundant comments for the same day.

### DIFF
--- a/R/sensorreading-data.R
+++ b/R/sensorreading-data.R
@@ -16,12 +16,10 @@ sensorreadingTable <- function(data){
                    "Reading Type",
                    "Reading",
                    "Uncertainty",
-                   "Comments",
                    "Method",
                    "Reading Type",
                    "Reading",
                    "Uncertainty",
-                   "Comments",
                    "Recorder w/in Uncertainty?", 
                    "Indicated Correction",
                    "Applied Correction",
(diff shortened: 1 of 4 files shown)


lindsayplatt:
So what exactly is the goal? Why do comments get there own row? Trying to get the big picture.

thongsav-usgs:
I don't think it matches the current data structure, but the way the old report looked (pre-aqcu) was you would have the comments listed once per "field visit". Multiple rows in the table may be from the same visit. This change was requested so you don't have to see redundant comments accross multiple rows. For this to truly be corrected we probably need to do some row grouping. Soliciting feedback from the POs now, not sure we should merge. I have feeling we'll double back on this.

lindsayplatt:
If it spans across multiple rows, would it make sense to try and do an excel type "merge cells"? I can try investigating to find that

thongsav-usgs:
Getting it out on DEV, but we need to follow up. As far as just rendering/organizing, I think it's where the POs want.

thongsav-usgs:
@lindsaycarr definitely want it to span multiple rows, go ahead and take a look at it now. I'm using a css hack to make it look like it does.

mhines-usgs:
The way I got that to work in the headers was with the htmlTable package...

https://cran.r-project.org/web/packages/htmlTable/index.html

PR #242, opened 2016-02-15
dvhydro: fixing issues with splitting & re-connecting the data

lindsayplatt:
split data to eliminate lines connecting gaps & connect non-estimated data with estimated data -- added more logic to handle a split on a data point at the end + had logic for which ts extended to meet the other backwards (estimated extends to connect with non-estimated)

### DIFF
--- a/R/dvhydrograph-data.R
+++ b/R/dvhydrograph-data.R
@@ -128,66 +128,71 @@ splitDataGaps <- function(data, gapData_nm, ignore_nm){
   notIgnore <- which(!names(data) %in% ignore_nm)
   ignore <- which(names(data) %in% ignore_nm)
   
-  data_split <- list()
+  if(length(notIgnore) != 0){
   
-  for(i in notIgnore){
-    t <- data[[i]][[gapData_nm]]
-    t_dates <- as.Date(format(t, "%Y-%m-%d"))
-    diff_days <- diff(t_dates)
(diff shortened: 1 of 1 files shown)


R/dvhydrograph-data.R
```diff
@@ -106,74 +106,104 @@ getStatDerived <- function(data, chain_nm, legend_nm, estimated){
...
-  data_split <- connectTS(data_split)
-  
-  data_split <- append(data_split, data[ignore])
 }
 
 connectTS <- function(data_split){
   
   len_data <- length(data_split)
-  if(len_data == 1){  # you can't connect data that is only of length one
+  if(len_data <= 1){  # you can't connect data that is only of length one
```
lindsayplatt, line 201: this change also fixed AQCU-595 -- because it stopped empty data

R/dvhydrograph-data.R
```diff
@@ -106,74 +106,104 @@ getStatDerived <- function(data, chain_nm, legend_nm, estimated){
...
     return(data_split)
   } else {
   
     first_dates <- lapply(data_split, function(d) {d$time[1]})
-    ranks <- rank(unlist(first_dates))
+    ranks <- order(unlist(first_dates))
```
lindsayplatt, line 206: this change also fixed AQCU-595 -- because it was giving the wrong index values

PR #248, opened 2016-02-25
Making VDiagram Taller and wider

mhines-usgs:
Per https://internal.cida.usgs.gov/jira/browse/AQCU-615

Adjusting from 6 to 11


mhines-usgs:
image

instead of

image

mhines-usgs:
image

PR #249, opened 2016-02-26
Removed unneeded complexity introduced by bad concurrency theory

thongsav-usgs:


thongsav-usgs:
Also FYI, @lindsaycarr

jiwalker-usgs:
:shipit:

PR #251, opened 2016-02-26
Edits to Extremes Report so that the report removes duplicate max/mins 

mhines-usgs:
If they are the same value. Keeps only the first record regardless of time.

https://internal.cida.usgs.gov/jira/browse/AQCU-605

@lindsaycarr can you check this out? Original and new examples attached to the JIRA ticket. 

I did a hacky fix on the column names because in the example I was working with at least, the column names need to be used in the !duplicated function, and in the example (see the attached json) two of the column names are the same and it hates that. So, I change the names temporarily, do the !duplicated call and then change them back to the originals..

### DIFF
--- a/R/extremes-data.R
+++ b/R/extremes-data.R
@@ -118,7 +118,11 @@ extremesTable <- function(rawData){
 
     toRet <- rbind(toRet,toAdd)
   }
-  
+  #Get rid of duplicates for same date, time, primary and upchain params
+  #Hack because we can't use the duplicated function on columns with the same name
+  colnames(toRet) <- c("","Date","Time","Test","Test2") 
+  toRet <- toRet[!duplicated(toRet[,c("Date","Test","Test2")]),]
+  colnames(toRet) <- columnNames
   return(toRet)
 }
 
--- a/inst/extremes/extremes.Rmd
+++ b/inst/extremes/extremes.Rmd
@@ -26,5 +26,5 @@ cat(getLogo())
 tbl <- extremesTable(data)
 # formTable <- padTable(tbl)
 # cat(formTable)
-kable(tbl, align='r')
+kable(tbl, align='r', row.names=FALSE)
 ```


lindsayplatt:
If those data columns are literally the exact same name, why are they both on there...

mhines-usgs:
I think it isn't always the case but I can't say. It's coming from two different variables in the json though, perhaps they just happen to be the same in this scenario?

mhines-usgs:
I keep thinking about this (the labels being the same). Here's a screenshot of the metadata of the data example I'm working with, @thongsav-usgs can you verify that in the java-json conversion our labels aren't getting mixed up somewhere?

image

jordansread:
:shipit:

thongsav-usgs:
What's the exact desired behavior? The issue is the parameter might be the exact same for primary and upchain time series, but they still come from different sources time series. I don't know what the goal of the request is, removing duplicate labels seems  wrong if the two time series are different and the actual request should be to make the labels more meaningful.

thongsav-usgs:
Wait, are we talking about duplicate row entries?

mhines-usgs:
From the JIRA ticket, "The report currently provides all inst min & max values in a given day no matter how many times the min or max occurs. It would be fine just to list the FIRST inst max/min in a day if multiple exist. This produces a much cleaner output."

The desired behavior is to remove redundant row entries from the same date, just to keep the first instance. E.g. here is the original with redundant rows for 2/1/2015 and -0.1 / -0.1 

image

thongsav-usgs:
Sigh...  I'd get the PO/RCs to triple think that request, they explicitly asked for this behavior (it used to only display 1 instance). It was a pretty explicit request since it was 3 points to refactor the data layer to provide all instances of the mins/maxes. I even remember bumping out other high priority tasks for this.

mhines-usgs:
Hm Ok. I already tagged Laura to review it so hopefully she'll see that. The ticket was just made Feb 11th...

PR #254, opened 2016-03-02
CORR: order data in lanes + fiveyr: approval info labeled correctly

lindsayplatt:
Fixing AQCU-593 and part of AQCU-619/620


lindsayplatt:
sometimes there were multiple approval date ranges for a single approval type, this new code should take care of that.

PR #255, opened 2016-03-07
adding ablines at each value on the axes. 

mhines-usgs:
simplest addition of more gridlines per Gary's comment on AQCU-615 without implementing minor tickmarks and reference lines
https://internal.cida.usgs.gov/jira/browse/AQCU-615?focusedCommentId=100976&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-100976


mhines-usgs:
@lindsaycarr do you mind peeking at this? Example plots are attached to the JIRA ticket. 

https://internal.cida.usgs.gov/jira/browse/AQCU-615

lindsayplatt:
@mhines-usgs Code looks good. Seems like we should add more to the x-axis? Would also be pretty simple to add major grid lines like he suggested by adding a style called `majorabline` or something. Or wait and see what Gary says

PR #267, opened 2016-03-28
Fix for AQCU-639

mhines-usgs:
Was getting an error in vDiagram, running into cases where value of a callout x variable was NA, and the math couldn't be done correctly to test the callout length/angle. So, check to see first if the variable is NA instead of numeric value desired, and skip that test if so.
https://internal.cida.usgs.gov/jira/browse/AQCU-639

### DIFF
--- a/R/utils-shared.R
+++ b/R/utils-shared.R
@@ -191,7 +191,7 @@ testCallouts <- function(plot_obj, xlimits){
   testCalloutsByView <- function(plot_obj, callouts_index, view_num, xlimits_real, width_char, xrange){
     for(i in callouts_index){
         callout_args <- plot_obj[[view_num]][[i]]
-        if (!is.na(callout_args$x[i] | !is.na(callout_args$y[i]))) {  
+        if (!is.na(xtfrm(callout_args$x[i]) | !is.na(xtfrm(callout_args$y[i])))) {  
           text_len <- nchar(callout_args$labels)
             
             len <- ifelse(is.null(callout_args$length), 0.1, callout_args$length)
--- a/R/utils-shared.R
+++ b/R/utils-shared.R
@@ -190,21 +190,23 @@ testCallouts <- function(plot_obj, xlimits){
   
   testCalloutsByView <- function(plot_obj, callouts_index, view_num, xlimits_real, width_char, xrange){
     for(i in callouts_index){
-      callout_args <- plot_obj[[view_num]][[i]]
-      text_len <- nchar(callout_args$labels)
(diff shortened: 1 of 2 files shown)


mhines-usgs:
Testing other example files locally for other plots to see what's up with that check failure. When I run a check locally I don't get any issues.

mhines-usgs:
looks like 
Error in Ops.POSIXt(callout_args$x[i], !is.na(callout_args$y[i])) : 
  '|' not defined for "POSIXt" objects
Calls: uvhydrograph ... createPrimaryPlot -> testCallouts -> testCalloutsByView -> Ops.POSIXt

Will test more here.

R/utils-shared.R
```diff
@@ -190,21 +190,23 @@ testCallouts <- function(plot_obj, xlimits){
...
-      tooLong <- xnew > xlimits_real[2]
-      
-      if(any(tooLong)){
-        out <- which(tooLong)
-        notout <- which(!tooLong)
-        plot_obj[[view_num]][[i]]$angle[notout] <- NA
-        plot_obj[[view_num]][[i]]$angle[out] <- 150
-      }
+        callout_args <- plot_obj[[view_num]][[i]]
+        if (!is.na(xtfrm(callout_args$x[i]) | !is.na(xtfrm(callout_args$y[i])))) {
```
thongsav-usgs, line 194: Echoing @lindsaycarr, but want '&' instead of '|'. Also do we want to go ahead and throw in null checks to be safe?

mhines-usgs, line 194: I'm not sure exactly why, but with an & it reports the original error message. I can throw in null checks, sure. Not really sure why the & doesn't seem to work.

thongsav-usgs, line 194: Sorry, didn't see conversation in main thread.

mhines-usgs, line 194: So with the ampersand, the logic comes out TRUE. with the pipe it comes up FALSE. Hmm.

thongsav-usgs:
Can we start adding utility functions for empty check. I'd propose a "isNotEmpty" function that checks for NA/Null/ and "" (empty string). We also should move "nullMask" function at https://github.com/USGS-R/repgen/blob/master/R/sitevisitpeak-data.R#L95 to the same area (utils-shared.R)

lindsayplatt:
I kind of have a similar thing going with `anyDataExist` but I agree that we need something like that  https://github.com/USGS-R/repgen/blob/master/R/utils-json.R#L70-74

PR #268, opened 2016-03-29
adding specific versions to install R packages to "lock down" depende…

mhines-usgs:
…ncies.

Used console output from yesterday's functioning deploy on dev to get versions, or grabbed latest version from cran if the version wasn't specified in the console (because that is what it would be grabbing).

https://internal.cida.usgs.gov/jira/browse/AQCU-589


thongsav-usgs:
Is there a way to fully explode the upward dependencies? I think the way R works is it doesn't guarantee the versions of any packages that these packages depend on.

mhines-usgs:
Ah you mean beyond what we have here, e.g. if magrittr depends on something else, fix that version too? (currently a "hidden" dependency?)

PR #273, opened 2016-03-30
updating json that have waterlevel data

lindsayplatt:
image


lindsayplatt:
@thongsav-usgs here's a quick table I put together of the status for example data. Need to wait and merge until we merge in your groundwater example, so that I can pull those changes

jiwalker-usgs:
:+1:

PR #274, opened 2016-03-30
Dependency redux

mhines-usgs:
[One more time..](https://www.youtube.com/watch?v=FGBhQbmPwH8)
also made this [sweet graphic](https://www.mindmeister.com/676221733/repgen)


lindsayplatt:
that mind map...holy moly

PR #276, opened 2016-03-31
stringi stringr heartbreaker

mhines-usgs:
movin on uppppp

stringi has to be installed before stringr.


thongsav-usgs:
:shipit:

PR #286, opened 2016-04-07
UvHydrograph handle timezones and plot last day.

dpattermann-usgs:


inst/uvhydrograph/uvhydrograph.Rmd
```diff
@@ -22,6 +22,8 @@ cat(getLogo())
 ***
 
 ```{r echo=FALSE, warning=FALSE, message=FALSE, dev='svg', results='hide'}
+Sys.setenv(TZ=data$reportMetadata$timezone)
+
```
dpattermann-usgs, line 26: I'm unsure if this is a good way to do it or if I should do this in the java layer (when setting up the Rserve request). I didn't want to do it for all reports so I put it in this specific report, but it can be tucked back into the java layer if we need to.

dpattermann-usgs:
Note: This will break all the UvHydrograph examples if it's pulled in (because it requires the timezone attribute to be in reportMetadata).

lindsayplatt:
IMO, Java layer should probably be setting the timezone. But does that mean it will be applied to all of the reports? Pulling some R veterans to weigh in on using `Sys.setenv` in our package @ldecicco-USGS @jread-usgs

PR #289, opened 2016-04-12
Uvhydro Do the same axis/last day change for secondary plots as primary plots.

dpattermann-usgs:
Also, PR for PT's capitalization.


lindsayplatt:
:+1:

PR #290, opened 2016-04-12
Per https://internal.cida.usgs.gov/jira/browse/AQCU-670

mhines-usgs:
This adds time to the last date in endDate to rectify the missing days bugs in dvhydro and fiveyeargwsum

Not sure if it is necessary in these reports, because the time window is so large, it may be hard to notice

### DIFF
--- a/R/fiveyeargwsum-data.R
+++ b/R/fiveyeargwsum-data.R
@@ -40,7 +40,7 @@ parseFiveYrSupplemental <- function(data, parsedData){
   horizontalGrid <- signif(seq(from=seq_horizGrid[1], to=seq_horizGrid[2], along.with=seq_horizGrid), 1)
   
   startDate <- formatDates(data$reportMetadata$startDate, "fiveyr", "start")
-  endDate <- formatDates(data$reportMetadata$endDate, "fiveyr", "end") + hours(23) + minutes(45)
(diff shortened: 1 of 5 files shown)


R/fiveyeargwsum-data.R
```diff
@@ -40,7 +40,7 @@ parseFiveYrSupplemental <- function(data, parsedData){
   horizontalGrid <- signif(seq(from=seq_horizGrid[1], to=seq_horizGrid[2], along.with=seq_horizGrid), 1)
   
   startDate <- formatDates(data$reportMetadata$startDate, "fiveyr", "start")
-  endDate <- formatDates(data$reportMetadata$endDate, "fiveyr", "end")
+  endDate <- formatDates(data$reportMetadata$endDate, "fiveyr", "end") + hours(23) + minutes(45)
```
mhines-usgs: tracing this to the formatDates function, seems like it's ending the month already on the 30th, so this may be really unnecessary? @dpattermann-usgs -- i'd like your feedback

dpattermann-usgs: Talked to @lindsaycarr about this and she thinks the hours 23 minutes 45 will still make a difference. I think we should look at the formatDates function for fiveYr a bit though.

PR #292, opened 2016-04-13
corr: changing capitalization back so it matches json...

lindsayplatt:
@dpattermann-usgs please make sure this is correct!!!


dpattermann-usgs:
:+1: Niceeee

PR #294, opened 2016-04-15
Added Q measurements to DV Hydrographs. Also consolidated data parsing

thongsav-usgs:
for ground water data between DV/UV hydrograph and 5yr reports. NOTE
that the json is changing from "groundWater" to "gwLevel".

### DIFF
--- a/R/utils-shared.R
+++ b/R/utils-shared.R
@@ -36,8 +36,8 @@ startRender <- function(data, output, author, reportName){
 ############ used in dvhydrograph-data, fiveyeargwsum-data, uvhydrograph-data ############ 
 
 getGroundWaterLevels<- function(ts, ...){
-  y <- as.numeric(ts$gwLevel[['groundWaterLevel']])
-  x <- ts$gwLevel[['dateString']]
+  y <- as.numeric(ts$gwlevel[['groundWaterLevel']])
+  x <- ts$gwlevel[['dateString']]
   time = as.POSIXct(strptime(x, "%Y%m%d"))
   month <- format(time, format = "%y%m") #for subsetting later by month
   return(data.frame(time=time, value=y, month=month, stringsAsFactors = FALSE))
--- a/R/utils-shared.R
+++ b/R/utils-shared.R
@@ -35,6 +35,70 @@ startRender <- function(data, output, author, reportName){
 
 ############ used in dvhydrograph-data, fiveyeargwsum-data, uvhydrograph-data ############ 
 
+getGroundWaterLevels<- function(ts, ...){
+  y <- as.numeric(ts$gwLevel[['groundWaterLevel']])
+  x <- ts$gwLevel[['dateString']]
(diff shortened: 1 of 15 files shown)


R/utils-shared.R
```diff
@@ -35,6 +35,70 @@ startRender <- function(data, output, author, reportName){
 
 ############ used in dvhydrograph-data, fiveyeargwsum-data, uvhydrograph-data ############ 
 
+getGroundWaterLevels<- function(ts, ...){
+  y <- as.numeric(ts$gwLevel[['groundWaterLevel']])
+  x <- ts$gwLevel[['dateString']]
+  time = as.POSIXct(strptime(x, "%Y%m%d"))
```
lindsayplatt, line 41: you should use a different var name. `time` is a stats function in R

lindsayplatt, line 41: oh wait, looks like that's how it has been...better to rename it if you've got the time

lindsayplatt:
@thongsav-usgs there are huge chunks that are shown as added/deleted, so I am having troubling figuring out what has changed. You added: `getGroundWaterLevels` and `getFieldVisitMeasurementsQPoints`? Both are used in dvhydro, but only `getGroundWaterLevels` is used in fiveyrgwsum. Is that correct?

thongsav-usgs:
@lindsaycarr getGWL is used in all 3 (DV/UV/5yr), the getFVM used in UV and DV.

Pretty much any function in uvhydro-data that was pulling common data elements, I moved into util-shared.R

Do we want a data specific shared util file?

lindsayplatt:
hmm well the utils-shared.R I think could be named better. All the util-[].R files are for shared utilities, so that name is pointless (I named it so I feel okay pointing out its uselessness). I think the data parsing functions are currently in utils-json. I would rename that to utils-data and consider migrating stuff from utils-shared into that.

thongsav-usgs:
Let's hold off and do a non-functional refactor later without any feature changes involved so we can isolate that.

R/utils-shared.R
```diff
@@ -35,6 +35,70 @@ startRender <- function(data, output, author, reportName){
 
 ############ used in dvhydrograph-data, fiveyeargwsum-data, uvhydrograph-data ############ 
 
+getGroundWaterLevels<- function(ts, ...){
+  y <- as.numeric(ts$gwlevel[['groundWaterLevel']])
+  x <- ts$gwlevel[['dateString']]
+  time = as.POSIXct(strptime(x, "%Y%m%d"))
+  month <- format(time, format = "%y%m") #for subsetting later by month
+  return(data.frame(time=time, value=y, month=month, stringsAsFactors = FALSE))
+}
```
lindsayplatt, line 44: only other comment: should we have a NULL check on this? getWaterQualityMeasurements has one, but some others don't. Or should that be addressed when we make that common null checking function?

thongsav-usgs, line 44: Let's also wait on that, we need to find a sprint to elevate all of our code improvement tasks and do them.

PR #298, opened 2016-04-19
Add timezones to all Created fields in Rmd.

dpattermann-usgs:
Also looks like there are different line endings in sig-uvhydrograph? I don't know how to swap them back though in RStudio.


dpattermann-usgs:
@lindsaycarr Can you help me out on why this is failing?

PR #299, opened 2016-04-22
More elegant labeling logic

mhines-usgs:


R/correctionsataglance-data.R
```diff
@@ -101,12 +101,14 @@ formatDateRange <- function(startD, endD){
...
+#   #don't print Month Year in plot if there isn't enough room inside the rectangle
   dateSeq <- startSeq
   startSpan <- as.numeric(endSeq[1] - startSeq[1])
   endSpan <- as.numeric(tail(endSeq, 1) - tail(startSeq, 1))
-  if(startSpan <= 16){dateSeq[1] <- NA}
-  if(endSpan <= 14){dateSeq[length(dateSeq)] <- NA}
+  startLabelText <- format(startD, "%b %Y")
+  endLabelText <-  format(endD, "%b %Y")
+  if(isTextLong(labelText=startLabelText,dateLim=NULL,startD,endD,totalDays=numdays)) {dateSeq[1] <- NA}
+  if(isTextLong(labelText=endLabelText,dateLim=NULL,startD,endD,totalDays=numdays)) {dateSeq[length(dateSeq)] <- NA}
```
mhines-usgs: @lindsaycarr Here you had suggested I use startD=startSpan and endD=endSpan, but I run into errors on that, and it looks like the isTooLong function at line 350, is looking for the difference between two dates, not the number # of days. Anyway, it's still plotting the end month/year for some reason. Can you look at this with me sometime today?

R/correctionsataglance-data.R
```diff
@@ -80,13 +80,13 @@ formatDateRange <- function(startD, endD){
...
   firstOfMonth_end <- day(endD) == 1
+  numdays <- as.numeric(difftime(strptime(endD, format="%Y-%m-%d"), strptime(startD,format="%Y-%m-%d"), units="days"))
   
   if(firstOfMonth){
     startSeq <- seq(startD, endD, by="1 month")
   } else {
     fromDate <- as.POSIXct(format(seq(startD, length=2, by="month")[2], "%Y-%m-01"))
-    numdays <- as.numeric(difftime(strptime(endD, format="%Y-%m-%d"), strptime(startD,format="%Y-%m-%d"), units="days"))
-    if (numdays<=27) {
+    if (fromDate >= endD) {
```
mhines-usgs, line 89: I think this is logical and works better than the arbitrary is it less than a full month (27).

R/correctionsataglance-data.R
```diff
@@ -342,13 +344,15 @@ findTextLocations <- function(dataIn, isDateData = FALSE, ...){
...
+    late <- which(endD > dateLim[2])
+    startD[early] <- dateLim[1]
+    endD[late] <- dateLim[2]
+    totalDays <- difftime(dateLim[2], dateLim[1], units="days")
+  } 
   
-  totalDays <- difftime(dateLim[2], dateLim[1], units="days")
   widthOfChar <- (1/365)*totalDays*3 #each character will be 1/365 * num days in the range * buffer
   widthOfLabel <- nchar(labelText)*widthOfChar
   widthOfRect <- difftime(endD, startD, units="days")
```
mhines-usgs, line 357: the way I'm passing parameters in, for example with example4.json, I don't want Apr 2016 to plot because the rectangle for that month is too small at the end of the sequence. The values coming in are 

labelText = Apr 2016
dateLim = NULL
startD = 2015-04-01
endD = 2016-04-11
totalDays = 376

widthOfChar for the labelText = 3.0904...
widthofLabel = 24.7232
widthOfRect = 376
so, when widthOfRect is compared to widthOfLabel, even on the last one, it fits, because it looks like it's comparing to the full length rather than the short rect available... I think.

mhines-usgs, line 357: So, need to figure out how to pass it the start and end dates for that specific section of the rectangle.. I think. Rather than the entire range for the request.

mhines-usgs, line 357: Talked and figured this out, I think :)

lindsayplatt:
you already merged this but I finally got a chance to look at it :+1:

R/correctionsataglance-data.R
```diff
@@ -101,13 +101,14 @@ formatDateRange <- function(startD, endD){
...
+#   #don't print Month Year in plot if there isn't enough room inside the rectangle
   dateSeq <- startSeq
   startSpan <- as.numeric(endSeq[1] - startSeq[1])
   endSpan <- as.numeric(tail(endSeq, 1) - tail(startSeq, 1))
-  if(startSpan <= 16){dateSeq[1] <- NA}
-  if(endSpan <= 14){dateSeq[length(dateSeq)] <- NA}
-  
+  startLabelText <- format(startD, "%b %Y")
+  endLabelText <-  format(endD, "%b %Y")
+  #if(isTextLong(labelText=startLabelText,dateLim=NULL,startD,endD,totalDays=numdays)) {dateSeq[1] <- NA}
```
mhines-usgs, line 110: Yeah I couldn't decide on if this is actually used to not plot the first month so I commented it out. Do you remember why you would want to shorten the start of the approval span? maybe that is the problem I'm seeing in AQCU-664, where it's shifting the labels on single month ranges... hmm. @lindsaycarr

R/correctionsataglance-data.R
lindsayplatt, line 357: is this out on dev? Just ran a report that has lots of months but they are all labeled, and its still crazy. I think this might be a diff ticket

https://cida-eros-aqcudev.er.usgs.gov:8443/aqcu-webservice/service/reports/correctionsataglance/?primaryTimeseriesIdentifier=1343310484b0434c83acacbe56fca350&station=01047200&startDate=2010-10-01&endDate=2013-09-30

mhines-usgs, line 357: so I'm only applying this check to the last rectangle right now. thankfully that's working, so it shouldn't be too complicated to check the others. is there a normal length of time they'll run this report for? does it have to work for any number of years?

R/correctionsataglance-data.R
lindsayplatt, line 110: Isn't dateSeq the labels for the rectangles (poor naming, it should be more clear - sorry that was my fault!)? I think the points is to make any label that is bigger than the rectangle not show up. If startD and endD are vectors, I think you might want to do this?

``` r
dateSeq[isTextLong(labelText=startLabelText,dateLim=NULL,startD,endD,totalDays=numdays)] <- NA
```

R/correctionsataglance-data.R
lindsayplatt, line 357: Ah ok, gotcha. And good question, I don't know. Seems like it, but dealing with tons of years where the labels overlap definitely is a different ticket's scope.

mhines-usgs, line 357: Ok, I'll see what I can do based on your other feedback. I'll inquire about the typical time range, at least we could change label styles (say 2 dig mo/2dig year?) if the range is too tight or something.

lindsayplatt, line 357: yeah, that would be good

PR #301, opened 2016-04-25
uvhydro: series_corr "N/A" label for when comments are missing

lindsayplatt:


lindsayplatt:
@mhines-usgs struggling with why the table won't show up for the series corrections. It happened before this PR, but I think we just did not notice? Or it's because of the new gsplot version on my computer. Could you test on yours?

lindsayplatt:
erg, I'll check out why those tests aren't passing on the plane

mhines-usgs:
@lindsaycarr sure, I'll test it out in a few minutes and let you know

those tests on Travis have been failing for a bit, I haven't paused yet on what I'm working on to figure out why

mhines-usgs:
@lindsaycarr I don't see any tables with corrections either, my gsplot is at 0.5.4 - I tested with local json, (uvhydro-example, uvhydro-hawaii)

mhines-usgs:
@lindsaycarr I figured out that the results='asis' versus results='hide' was hiding the table. I put in a new PR for that. previously the table was coming through in the report as NULL so I had added the results='hide' to hide that. Apparently we don't want that, though.

lindsayplatt:
ahhh good find! So we still need to figure out a different way to stop the NULL from displaying

lindsayplatt:
PR to change it back: https://github.com/USGS-R/repgen/pull/307

PR #302, opened 2016-04-25
applies isTooLong to full dateSeq

mhines-usgs:
going to merge this but will push up changes to render the shorter date format in another PR. I can't seem to get the logic right on what I want to do, though.

for some reason, the isTextLong on its own I could never actually get to apply to all the items in the vector. probably something I'm doing wrong, but at least this amazing for loop works. 

right now, if you put in a time range that's over 13 months, you're going to run out of room in the rectangles and see no dates at all.


mhines-usgs:
here is what I want to do with the render function but I'm getting errors on it. will keep poking at it.

https://github.com/mhines-usgs/repgen/commit/97791ce9df7516448b5f1c4e7ebfd150e65b90b7

PR #306, opened 2016-04-27
moving labels to left

mhines-usgs:
rather than right. AQCU-664 addressing an issue where if a point is on the end date of a time range, the footnote label displays of the chart.


lindsayplatt:
Just getting to glance at all of these. Won't this now create a problem if the labels are at the start date of a time range? Won't the get cut off the on the other side?

PR #309, opened 2016-04-28
Check for primarySeriesCorrections before parsing them, otherwise thi…

mhines-usgs:
…s appears to fail for UVhydrograph, https://internal.cida.usgs.gov/jira/browse/AQCU-678


mhines-usgs:
not sure why there are still test errors, even locally. the tests they are running actually create reports.

PR #310, opened 2016-04-28
getting tests to pass again

lindsayplatt:
utils-shared: determine if the list is empty before proceeding in getCorrections


lindsayplatt:
@mhines-usgs yay! it works now :) I think this will fix the new error that came up and made it into AQCU-703. Plus, AQCU-678 should be good to go now.

mhines-usgs:
Nice!

PR #311, opened 2016-04-29
Test structure refactor

thongsav-usgs:


mhines-usgs:
:+1:

mhines-usgs:
I just pulled this and will add to it

scratchpad.R
```diff
@@ -0,0 +1,6 @@
+library(jsonlite)
```
mhines-usgs, line 1: did you mean to check this in?

scratchpad2.R
```diff
@@ -0,0 +1,4 @@
+library(repgen)
```
mhines-usgs, line 1: same ? did you mean to check this in?

thongsav-usgs, line 1: Yes to both, they are helpful for reference when not building with RStudio. I can move them if there is a better place for them, but they were supposed to just be helper files during development. I could be convinced to drop them also if @lindsaycarr or @jiwalker-usgs don't like them either.

PR #312, opened 2016-04-29
move label position

mhines-usgs:
depending on whether the xyText and the dateRange have overlapping values.


lindsaycarr:
you should checkout the testCallouts function. Might be able to generalize it to move the position or callout angle. It is used to flip the callout angle if the text goes outside of the plotting region

https://github.com/USGS-R/repgen/blob/master/R/utils-shared.R#L253

PR #314, opened 2016-04-29
if xyText for a lane falls on start/end date for footnote item, move it to left or right

mhines-usgs:
depending


lindsayplatt:
moved this comment over here from a PR you recently closed (didn't refresh my page, so I did not know it was closed!):

you should checkout the testCallouts function. Might be able to generalize it to move the position or callout angle. It is used to flip the callout angle if the text goes outside of the plotting region

https://github.com/USGS-R/repgen/blob/master/R/utils-shared.R#L253

mhines-usgs:
Ok @lindsaycarr, do you think this is OK to merge this for now and then I can see if I can figure out a way to edit that function to work for this need?

R/correctionsataglance-render.R
```diff
@@ -116,14 +116,28 @@ plotLanes <- function(gsplotObject, laneData, whichCol,
              labels = laneData[[labelName]], cex=1) 
     }
     
-    if(any(names(laneData) %in% 'numText')){  
+    if(any(names(laneData) %in% 'numText')){   
       i <- which(!is.na(laneData$numText))
+      if (any(laneData$moveText)) { 
+        pos <- NA
+        dateRange <- format(addData$dateData$dateRange, "%m/%d/%Y")
+        xyText <- format(laneData$xyText$x, "%m/%d/%Y")
```
lindsayplatt: can you rename this so it is more obvious that it is the x position of the xyText? I thought it was that actual text

R/correctionsataglance-render.R
```diff
@@ -116,14 +116,28 @@ plotLanes <- function(gsplotObject, laneData, whichCol,
...
+          if (any(xyText == dateRange[1])) {
```
lindsayplatt: shouldn't we be checking if it is <= the start date? And similarly, if it is >= the end date on line 128?

lindsayplatt:
yeah, I think that's a good plan. Just put some comments on the actual code though, but maybe merge after that?

R/correctionsataglance-render.R
mhines-usgs: changed to labelDate

R/correctionsataglance-render.R
mhines-usgs: changed this as suggested, but I don't think we _should_ see any dates > or < , but you never know!

PR #322, opened 2016-05-04
add isEmpty utility that checks for !is.null() and !is.na()

kmschoep-usgs:


R/utils-shared.R
```diff
@@ -171,3 +171,14 @@ formatComments <- function(comments){
   table_comments <- do.call(paste0, htmlbreaks_end)
   return(table_comments)
 }
+
+############ used in various places ############ 
+
+isEmpty <- function(val){
+  if(!is.null(val) && !is.na(val)){
```
jiwalker-usgs, line 178: I think this would read easier if it was negated `if(is.null(val) || is.na(val)) result <- TRUE`

mhines-usgs:
@kmschoep-usgs -- I merged for now to test locally within a function or two. We can always change it if it makes more sense to do it the way @jiwalker-usgs suggests

kmschoep-usgs:
I tried it the way @jiwalker-usgs suggested, and the tests still work out as I expected, so I'll change it for improved readability.

PR #324, opened 2016-05-04
Adding in the comparison site label into the legend per AQCU-683

mhines-usgs:
https://internal.cida.usgs.gov/jira/browse/AQCU-683


R/uvhydrograph-data.R
```diff
@@ -94,7 +94,7 @@ parseUVSupplemental <- function(data, plotName, pts) {
     
     primary_lbl <- getUvLabel(data, "primarySeries")
     date_lbl <- paste(lims_UV$xlim[1], "through", lims_UV$xlim[2])
-    comp_UV_lbl <- data[['comparisonSeries']]$name
+    comp_UV_lbl <- data$reportMetadata$comparisonStationId
```
mhines-usgs, line 97: I stole this var since it wasn't being used in its old form.. not sure why it was there in the first place

PR #326, opened 2016-05-04
add isEmptyOrBlank to test for is.null() and empty string.

kmschoep-usgs:


R/utils-shared.R
```diff
@@ -175,10 +175,21 @@ formatComments <- function(comments){
...
   } else {
+    result <- FALSE
+  }
+  return(result)
+}
+
+############ used in various places ############ 
+
+isEmptyOrBlank <- function(val){
+  if(is.null(val) || val==""){
```
dpattermann-usgs: Does this catch the is.na() case?

kmschoep-usgs: no, just null.  If it needs to handle all three cases, I can add that.  I was just going by the informal-ish requirements in slack.

dpattermann-usgs: I honestly don't know if it needs to. Just was confused because I wasn't sure if isEmptyOrBlank should do everything that isEmpty does and more.

R/utils-shared.R
```diff
@@ -175,10 +175,21 @@ formatComments <- function(comments){
 ############ used in various places ############ 
 
 isEmpty <- function(val){
-  if(!is.null(val) && !is.na(val)){
-    result <- FALSE
+  if(is.null(val) || is.na(val)){
```
thongsav-usgs: You can assign what's in the if clause directly to the result variable. You don't have to do an if/else if all it does is assign true/false.

R/utils-shared.R
```diff
@@ -175,10 +175,21 @@ formatComments <- function(comments){
...
   } else {
+    result <- FALSE
+  }
+  return(result)
+}
+
+############ used in various places ############ 
+
+isEmptyOrBlank <- function(val){
+  if(is.null(val) || is.na(val) || val==""){
```
thongsav-usgs: Same comment, can assign directly to result.

PR #328, opened 2016-05-05
replace cases of (!is.null() && !is.na()) with (!isEmpty()).

kmschoep-usgs:


mhines-usgs:
looks good so far!! I made two old/new and compared side by side, looks legit. I just committed two more small edits for sensorreading, will check out sitevisitpeak.

PR #329, opened 2016-05-05
more isEmpty goodness

mhines-usgs:
@kmschoep-usgs -- just made a few more small changes for sensorreading

### DIFF
--- a/R/sensorreading-data.R
+++ b/R/sensorreading-data.R
@@ -47,7 +47,7 @@ formatSensorData <- function(data, columnNames){
     listElements <- data[listRows,]
     
     if ("displayTime" %in% names(data)) {
-      if(!is.na(listElements$displayTime) || is.null(listElements$time)) {
+      if(!isEmpty(listElements$displayTime) || isEmpty(listElements$time)) {
         dateTime <- (strsplit(listElements$displayTime, split="[T]"))
         date <- strftime(dateTime[[1]][1], "%m/%d/%Y")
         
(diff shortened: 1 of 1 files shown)


R/sensorreading-data.R
```diff
@@ -47,7 +47,7 @@ formatSensorData <- function(data, columnNames){
     listElements <- data[listRows,]
     
     if ("displayTime" %in% names(data)) {
-      if(!is.na(listElements$displayTime) || is.null(listElements$time)) {
+      if(!isEmpty(listElements$displayTime) || isEmpty(listElements$time)) {
```
lindsayplatt, line 50: not sure this is doing what we would expect...especially for listElements$time. There might be a case where we can continue if its NA, but not NULL? Hence, it was only checking for NULL, but isEmpty checks for both. Which we had tests that looked at this unique condition...

lindsayplatt, line 50: @mhines-usgs @thongsav-usgs @dpattermann-usgs what are your thoughts on my previous comment? Should we be concerned?

mhines-usgs, line 50: I think you're right @lindsaycarr, I overlooked that this was operating on different variables!

lindsayplatt, line 50: could you update in your next PR? I'll likely not be merging anything anytime soon

lindsayplatt, line 50: oh wait, I see it here https://github.com/USGS-R/repgen/pull/333 !!!

mhines-usgs:
yep, i'm waiting on it right now https://cida-eros-aqcudev.er.usgs.gov:8443/jenkins/job/Repgen%20DEV%20Deployer/381/console

PR #331, opened 2016-05-05
Allow report to render for Aquarius-only training sites by removing

kmschoep-usgs:
requirement that stationName be present in json.

### DIFF
--- a/R/extremes-data.R
+++ b/R/extremes-data.R
@@ -7,6 +7,8 @@
 extremesTable <- function(rawData){
   
   data <- applyQualifiers(rawData)
+  l_qual <- data$dv$qualifiers
+  if (length(l_qual)==0) return ("The dataset requested is empty.")
   
   primaryLabel <- getReportMetadata(rawData,'primaryLabel')
   primaryParameter <- getReportMetadata(rawData,'primaryParameter')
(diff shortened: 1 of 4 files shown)


R/extremes-data.R
```diff
@@ -7,6 +7,8 @@
 extremesTable <- function(rawData){
   
   data <- applyQualifiers(rawData)
+  l_qual <- data$dv$qualifiers
+  if (length(l_qual)==0) return ("The dataset requested is empty.")
```
thongsav-usgs, line 11: The data does not have to have qualifiers, this is cutting off rendering of valid data sets.

inst/extremes/extremes.Rmd
```diff
@@ -26,5 +26,8 @@ cat(getLogo())
 tbl <- extremesTable(data)
 # formTable <- padTable(tbl)
 # cat(formTable)
-kable(tbl, align='r', row.names=FALSE)
+if(!identical(tbl, "The dataset requested is empty.")){
```
thongsav-usgs, line 29: I would nest this boolean logic a function. Right now you are relying on two hardcoded strings from two different places to be equal. A dev may change it in one place but not the other.

kmschoep-usgs, line 29: this logic is all over repgen.  I can change it here, but we should put in a tech debt task to apply the change everywhere.

thongsav-usgs, line 29: Yes, there are a lot of instances where we need to consolidate shared logic, that's why utils-shared was created.

R/extremes-data.R
kmschoep-usgs, line 11: I'll need an example data set with no qualifiers but other valid data.

PR #334, opened 2016-05-10
check to see if data exist in environment

mhines-usgs:


lindsayplatt:
@mhines-usgs do you want to add tests to make sure the isEmptyOrBlank functions as we expect? I think we could just copy and paste the very very simple stuff I wrote for now

mhines-usgs:
@lindsaycarr sure, I can add those from slack. Thanks.

PR #335, opened 2016-05-13
appropriately check for missing extremes data

kmschoep-usgs:


inst/extdata/extremes/extremes-example.json
```diff
@@ -112,8 +112,7 @@
     "locationNumber": "06933500",
     "locationLon": "-91.9773333333333",
     "dvUnit": "ft^3/s",
-    "startDate": "2014-12-16T00:00:00-06:00",
-    "isInverted": false
```
mhines-usgs, line 116: is this not needed? (isInverted?) @kmschoep-usgs

R/sig-extremes.R
```diff
@@ -7,7 +7,7 @@
 #'@importFrom rmarkdown render
 #'@examples
 #'library(jsonlite)
-#'data <- fromJSON(system.file('extdata','extremes','extremes-example.json',package = 'repgen'))
```
mhines-usgs, line 10: you could leave the original file (unless it's no longer a valid dataset to test with) and just have two examples here?

inst/extdata/extremes/extremes-example.json
kmschoep-usgs, line 116: It was a duplicate isInverted attribute

mhines-usgs, line 116: :+1:

PR #337, opened 2016-05-13
removing the # signs here, shouldn't be necessary, printing out weird…

mhines-usgs:
… ## at the bottom of report in the table

@kmschoep-usgs don't think we need those # symbols in there, they were causing the pound signs to print on the bottom of the table, like the r if statements above, just need to be wrapped in tick marks instead. double pound at the beginning is actually R Markdown syntax for header 2 (syntax to format text) which you effectively get if the command is true. But if it's not true, we see two # signs in the last row of the table that prints out.

### DIFF
--- a/inst/extremes/extremes.Rmd
+++ b/inst/extremes/extremes.Rmd
@@ -30,4 +30,4 @@ if(!identical(tbl, "The dataset requested is empty.")){
   kable(tbl, align='r', row.names=FALSE)
 }
 ```
-#`r if(identical(tbl, "The dataset requested is empty.")) {paste(tbl)}`#
+`r if(identical(tbl, "The dataset requested is empty.")) {paste(tbl)}`


mhines-usgs:
image

kmschoep-usgs:
Ah, thanks!  I was pulling that from another report.  Good to know.

mhines-usgs:
@kmschoep-usgs we might have weird # too there, which report if you remember?

kmschoep-usgs:
sitevisitpeak

mhines-usgs:
Hmm, I wonder why they don't print by default for that report. I just ran a few locally and expected them to show up at the bottom of the table like extremes..

kmschoep-usgs:
Oh, I see now why I put that line between the #'s in extremes, without doing that the empty dataset string is not formatted like on the other reports.  Will have to do more exploring with styles to get it working.

mhines-usgs:
Yeah, check this out: https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwjwu8ufv9fMAhULymMKHUHVATUQFggcMAA&url=https%3A%2F%2Fwww.rstudio.com%2Fwp-content%2Fuploads%2F2015%2F02%2Frmarkdown-cheatsheet.pdf&usg=AFQjCNHTjjlYFQjxyljotaYc6U4Wagw-CQ&bvm=bv.122129774,d.cGc

or google r markdown cheat sheet

I'm not sure why they (extremes and svp) behave differently, though...

PR #339, opened 2016-05-17
Add more test cases for extremes report, change site metadata source to

kmschoep-usgs:
stationName, attempt to change style of no dataset output.


mhines-usgs:
I'm changing the existing test json to include the stationName, I see a lot of locationNames there and that's why it's not showing up when I test locally :)  

I wonder if that change was missed on the java side too, because I see it on a report I just generated. I'll pull the json and look.

mhines-usgs:
Yup, still using locationName, boo. I can find that in the java and edit it and have @dpattermann-usgs check it out :D

PR #340, opened 2016-05-17
Change in Approval Description would cause approval not to show

ee-usgs:
This was caused by using the description as the 'key' - now using the numeric level
- Levels 0, 1 & 2 now matched instead of description text
- New/future approval types (levels other than 0-2) will default to grey
- new correctionsataglance-example5.json used to validate (how to wire up for testing??)


mhines-usgs:
@ee-usgs Lindsay mentioned on JIRA too, but this works fine (your changes) there are other issues we'll have to work out, though. Good job!

PR #342, opened 2016-05-18
Addressing the CORR colors MIA

mhines-usgs:
https://internal.cida.usgs.gov/jira/browse/AQCU-715

### DIFF
--- a/R/correctionsataglance-data.R
+++ b/R/correctionsataglance-data.R
@@ -3,7 +3,7 @@
 parseCorrectionsData <- function(data){
 
   dateData <- formatDateRange(data$primarySeries$requestedStartTime, data$primarySeries$requestedEndTime)
-  apprData <- formatDataList(data$primarySeries$approvals, 'APPROVALS', datesRange = dateData$dateSeq) #top bar = primary series approvals
+  apprData <- formatDataList(data$primarySeries$approvals, 'APPROVALS', datesRange = dateData$realSeq) #top bar = primary series approvals
   fieldVisitData <- list(startDates = formatDates(data$fieldVisits$startTime),
                          dataNum = length(data$fieldVisits$startTime)) #points on top bar = field visits
   
@@ -103,6 +103,7 @@ formatDateRange <- function(startD, endD){
   
 #   #don't print Month Year in plot if there isn't enough room inside the rectangle
   dateSeq <- startSeq
+  realSeq <- dateSeq #so we're not passing NA into other places of the report
   labelSeq <- format(dateSeq, " %m/%Y ")
   for (i in 1:length(dateSeq)) { 
     if (isTextLong(labelText=labelSeq[i],dateLim=NULL,startD=startSeq[i],endD=endSeq[i],totalDays=numdays))
@@ -111,6 +112,7 @@ formatDateRange <- function(startD, endD){
   
   return(list(dateRange = c(startD, endD),
               dateSeq = dateSeq,
+              realSeq = realSeq,
               startMonths = startSeq,
               endMonths = endSeq,
               middleDate = median(startSeq)))
--- a/R/correctionsataglance-data.R
+++ b/R/correctionsataglance-data.R
@@ -117,10 +117,21 @@ formatDateRange <- function(startD, endD){
 }
 
 formatDataList <- function(dataIn, type, ...){
+  
   args <- list(...)
   
   if(length(dataIn) == 0){
     return()
+  }
+  
+  if(type == 'APPROVALS' && length(dataIn)>0){
+    for (i in 1:length(dataIn$endTime)) {
+      if (as.Date(dataIn$endTime[i]) > "9000-12-31") {
+        endT <- format(as.Date(dataIn$endTime[i]), format="%m-%d")
+        endT <- paste0("9000-",endT,"T00:00:00.000-06:00")
+        dataIn$endTime[i] <- endT
+      }
+    }
   }
   
   start <- which(names(dataIn) %in% c('startTime', 'startDate'))
--- a/R/correctionsataglance-data.R
+++ b/R/correctionsataglance-data.R
@@ -128,9 +128,9 @@ formatDataList <- function(dataIn, type, ...){
   
   if(type == 'APPROVALS' && length(dataIn)>0){
     for (i in 1:length(dataIn$endTime)) {
-      if (as.Date(dataIn$endTime[i]) > "9000-12-31") {
+      if (as.Date(dataIn$endTime[i]) > "2050-12-31") {
         endT <- format(as.Date(dataIn$endTime[i]), format="%m-%d")
-        endT <- paste0("9000-",endT,"T00:00:00.000-06:00")
+        endT <- paste0("2050-",endT,"T00:00:00.000-06:00")
         dataIn$endTime[i] <- endT
       }
     }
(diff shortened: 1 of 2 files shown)


R/correctionsataglance-data.R
```diff
@@ -117,12 +117,23 @@ formatDateRange <- function(startD, endD){
 }
 
 formatDataList <- function(dataIn, type, ...){
+
```
ee-usgs, line 122: This formatDataList method is a bit weird.  Its called 8 times - Are those 8 usages all so similar that they should use this same method?  It looks like it might make sense to have methods like this:
formatApprovalList()
formatMetaList()
handleCommonlistFormatting() //called by the other two methods after they have done their list-specific stuff.

R/correctionsataglance-data.R
```diff
@@ -117,12 +117,23 @@ formatDateRange <- function(startD, endD){
...
 
 formatDataList <- function(dataIn, type, ...){
+  
   args <- list(...)
   
   if(length(dataIn) == 0){
     return()
   }
   
+  if(type == 'APPROVALS' && length(dataIn)>0){
```
ee-usgs, line 129: If keeping this as a single method, moving all of the APPROVALS stuff into a single if block would make this more readable.  Big picture:

if (Approvals) {
//do this
} else if (Meta) {
//do that
} else {
//do we expect this case to happen?
}
//do stuff common to all list formatting

…but overall, I think it makes sense to try to avoid having to pass in a flag (approval or meta) to tell the method what to do.

R/correctionsataglance-data.R
mhines-usgs, line 122: I'm not sure, exactly, this isn't code that I wrote. If you think there could be improvements, perhaps we should create some jiras for future work?

lindsayplatt, line 122: this is code I wrote, and I'm definitely open for improvements! I agree with @mhines-usgs: since this is working, additional work should go into a technical debt ticket.

mhines-usgs:
So to you both @ee-usgs & @lindsaycarr , I am not sure the 9999 years are the primary problem. Changing some json datasets seems to make them work properly but not all the datasets. The two on the JIRA task work now with this change, but the originals don't. Lindsay, are you willing to take a look at this? Anything else jump out at you?

mhines-usgs:
Hmmm I think this bug is due to the apprDates sometimes replacing real dates with NA from our labeling tweak changing some dates to NA (see formatDateRange dateSeq) . I think all I need to do is make sure that we don't re-use that dateSeq list for anything else in order to avoid the problem..

lindsayplatt:
@mhines-usgs nice find!

lindsayplatt:
...by nice, I mean good thing you found that :)

mhines-usgs:
So I fixed that, but.. still plotting wrong. Continuing to poke...

lindsayplatt:
documentation in R makes it sound like those years are fine: 

> For input, only years 0:9999 are accepted

https://stat.ethz.ch/R-manual/R-devel/library/base/html/strptime.html

R/correctionsataglance-data.R
```diff
@@ -128,9 +128,9 @@ formatDataList <- function(dataIn, type, ...){
   
   if(type == 'APPROVALS' && length(dataIn)>0){
     for (i in 1:length(dataIn$endTime)) {
-      if (as.Date(dataIn$endTime[i]) > "9000-12-31") {
+      if (as.Date(dataIn$endTime[i]) > "2050-12-31") {
         endT <- format(as.Date(dataIn$endTime[i]), format="%m-%d")
-        endT <- paste0("9000-",endT,"T00:00:00.000-06:00")
+        endT <- paste0("2050-",endT,"T00:00:00.000-06:00")
```
lindsayplatt, line 133: when replacing 2050 with a different year, 8944 still works but 8945 does not....WHAT

mhines-usgs, line 133: :(( what the heck. 2050 seems to work though, in _all_ cases I tried. It's very weird. I'm really not sure what to think...

lindsayplatt, line 133: if you look at the gsplot object, the rectangle positions are correct, but somehow it's showing up incorrectly?

If you just do a gsplot thing, it seems to work just fine:

``` r
s <- c(as.POSIXct("1922-01-01 00:00:00"), as.POSIXct("2015-08-01 01:00:00"))
e <- c(as.POSIXct("2015-08-01 01:00:00"), as.POSIXct("9999-08-01 00:00:00"))
gs <- gsplot() %>%
    points(x = as.POSIXct(NA), y = NA, ylim=c(0,10), 
           xlim=c(as.POSIXct("2015-01-01"), as.POSIXct("2016-01-01"))) %>%
    rect(xleft = s, ybottom = 8, xright = e, ytop = 9, 
         col = c("green", "red"), border = NA)
gs
```

mhines-usgs, line 133: Another random clue, I tried changing the order of the approvals in the JSON, and they seem to overwrite eachother. I feel like there's a mixup occurring somewhere in the rectangle code but it's so greek I can't seem to pick it apart to say. And why this 2050 seems to fix it, again, no idea.

lindsayplatt, line 133: hmmm another weird thing, when I get rid of `xlim=parseData$additionalPlotData$dateData$dateRange` on L16, the approval rectangles look correct

mhines-usgs, line 133: do you think this is some kind of weird gsplot bug?

lindsayplatt, line 133: I'm starting to, but I can't seem to recreate it

lindsayplatt:
@mhines-usgs try removing `dev='svg'` from the chunk options and then change your 2050 thing to 9999. It should work...so now I'm looking into why that matters

mhines-usgs:
Sooo, whaaatttt!!! That's insane! How on earth did you find that? It does work...

lindsayplatt:
I was able to recreate using base R plotting:

# dev='svg' + xlim specified inside points

```
{r echo=FALSE, warning=FALSE, message=FALSE, dev='svg', results='asis', fig.width=18, fig.height=9}
s <- c(as.POSIXct("1922-01-01 00:00:00"), as.POSIXct("2015-08-01 01:00:00"))
e <- c(as.POSIXct("2015-08-01 01:00:00"), as.POSIXct("9999-08-01 00:00:00"))

plot(x = as.POSIXct(NA), y = NA, ylim = c(0,10), 
     xlim = c(as.POSIXct("2015-01-01"), as.POSIXct("2016-01-01")))
rect(xleft = s, ybottom = 8, xright = e, ytop = 9, col = c("green", "red"), border = NA)
```

# no dev='svg' + xlim specified inside points

```
{r echo=FALSE, warning=FALSE, message=FALSE, results='asis', fig.width=18, fig.height=9}
s <- c(as.POSIXct("1922-01-01 00:00:00"), as.POSIXct("2015-08-01 01:00:00"))
e <- c(as.POSIXct("2015-08-01 01:00:00"), as.POSIXct("9999-08-01 00:00:00"))

plot(x = as.POSIXct(NA), y = NA, ylim = c(0,10), 
     xlim = c(as.POSIXct("2015-01-01"), as.POSIXct("2016-01-01")))
rect(xleft = s, ybottom = 8, xright = e, ytop = 9, col = c("green", "red"), border = NA)
```

PR #343, opened 2016-05-23
uvhydro error with month of all NA data values + dv at noon

lindsayplatt:
fix for [719 ](https://internal.cida.usgs.gov/jira/browse/AQCU-719) & [713](https://internal.cida.usgs.gov/jira/browse/AQCU-713)


lindsayplatt:
merging because I added a test but coverage still decreased...

PR #344, opened 2016-05-24
Fix bug in selecting comments to print.

dpattermann-usgs:
Fixes AQCU-668. We weren't printing the first comment of SRS reports because lastDate was never initialized.

### DIFF
--- a/R/sensorreading-data.R
+++ b/R/sensorreading-data.R
@@ -127,12 +127,12 @@ formatSensorData <- function(data, columnNames){
     selectedRecComm <- ''
     
     #only display comments that haven't already been displayed and are in this same date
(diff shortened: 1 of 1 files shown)


R/sensorreading-data.R
```diff
@@ -127,12 +127,12 @@ formatSensorData <- function(data, columnNames){
     selectedRecComm <- ''
     
     #only display comments that haven't already been displayed and are in this same date
-    if(date == lastDate && lastRefComm != refComm) {
+    if((date == lastDate && lastRefComm != refComm) || (isEmptyOrBlank(lastDate))) {
```
lindsayplatt, line 130: @dpattermann-usgs is `lastRefComm` also empty when it's the first go around

PR #347, opened 2016-06-02
changing where the labels come from 

mhines-usgs:
so that primary label isn't used by default. now looks at the data to get the labels

### DIFF
--- a/R/uvhydrograph-data.R
+++ b/R/uvhydrograph-data.R
@@ -8,6 +8,7 @@ getMonths <- function(data){
 }
 
 parseUVData <- function(data, plotName, month) {
+  
   if(plotName == "primary"){
     
     corr_UV <- subsetByMonth(getUvHydro(data, "primarySeries" ), month)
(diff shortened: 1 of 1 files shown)


R/uvhydrograph-data.R
```diff
@@ -20,19 +21,20 @@ parseUVData <- function(data, plotName, month) {
...
                                         plot_type="uvhydro", month=month)
-    approvals_dv_max <- getApprovals(data, chain_nm="derivedSeriesMax", legend_nm=paste("DV Max", primary_lbl),
+    approvals_dv_max <- getApprovals(data, chain_nm="derivedSeriesMax", legend_nm=paste("DV Max", data$derivedSeriesMax$type, data$derivedSeriesMax$units),
                                             appr_var_all=c("appr_approved_dv", "appr_inreview_dv", "appr_working_dv"), 
                                             plot_type="uvhydro", month=month, point_type=24)
-    approvals_dv_mean <- getApprovals(data, chain_nm="derivedSeriesMean", legend_nm=paste("DV Mean", primary_lbl),
+    approvals_dv_mean <- getApprovals(data, chain_nm="derivedSeriesMean", legend_nm=paste("DV Mean", data$derivedSeriesMean$type, data$derivedSeriesMean$units),
                                             appr_var_all=c("appr_approved_dv", "appr_inreview_dv", "appr_working_dv"), 
                                             plot_type="uvhydro", month=month, point_type=21)
     approvals_dv_median <- getApprovals(data, chain_nm="derivedSeriesMedian", legend_nm=paste("DV Median", primary_lbl),
```
lindsayplatt, line 34: I think this one needs `data$derivedSeriesMean$type` as well

mhines-usgs, line 34: dangit!

lindsayplatt, line 34: also, looks like there's a function that already does this `getUvLabel`

mhines-usgs, line 34: I did that, (first or second commit) but was causing issues in the isEmptyOrVar function (not valid for atomic vector error), so worked around that..

R/uvhydrograph-data.R
```diff
@@ -20,19 +21,23 @@ parseUVData <- function(data, plotName, month) {
     meas_Q <- subsetByMonth(getFieldVisitMeasurementsQPoints(data), month)  
     
     primary_lbl <- getUvLabel(data, "primarySeries")
+    dvmin_lbl <- getUvLabel(data, "derivedSeriesMin")
```
mhines-usgs: @lindsaycarr here is where I tried to use the getUvLabel
but was getting the "$ operator is invalid for atomic vectors" coming from the isEmptyVar function on the variable incoming, which was just "Temperature, water degC"

mhines-usgs: I then tried to wrap something around that function so it wouldn't enter it, but couldn't get that figured out, so went back to just modifying how I have in the latest.

lindsayplatt: hmmm seems to work with the uvhydro-example.json data for me. I'll test it out after the meeting I am about to go to  (only half hour)

mhines-usgs:
Ok, on the ticket itself, I have the example dataset from Laura, try that if you get a chance!

https://internal.cida.usgs.gov/jira/browse/AQCU-718

PR #348, opened 2016-06-02
one more change for aqcu-718.

mhines-usgs:
forgot the median


lindsayplatt:
moving ticket to test - will be good to go once this is merged (and I'm hoping it will be within this sprint)

PR #349, opened 2016-06-02
dynamic labels using getUvLabel

lindsayplatt:
@mhines-usgs this passes checks on my computer and I also downloaded the uvhydro-aqcu718.json file and used that as data. Seemed to work fine...


mhines-usgs:
awesome. I was trying to assign that getUvLabel to new variables above that code block, perhaps that was why...

lindsayplatt:
hmmm maybe? Seems weird though because it was the actual function that was having a fit

mhines-usgs:
I'm not sure, do you want to come by and see? very confusing... not sure why yours wouldn't error and mine does...

PR #350, opened 2016-06-03
removing discharge measurements from report for dvhydro

mhines-usgs:
only want them to display if they are selected, not display them all the time.

https://internal.cida.usgs.gov/jira/browse/AQCU-717


lindsayplatt:
I'd say this is good for removing it from the R code, but @thongsav-usgs has suggestions for the Java side on JIRA

mhines-usgs:
Yep, saw those. I'll do a new PR for that later today! Thanks!

PR #354, opened 2016-06-10
Converting to down sampled values for several series.

ee-usgs:
R changes for downsampling.
See https://github.com/USGS-CIDA/AQCU/pull/510 for Java changes and context.


lindsayplatt:
@thongsav-usgs here's the downsampled R changes. Why on earth did coverage decrease by 14%??

thongsav-usgs:
@lindsaycarr, I'm betting it counts lines of example json as code. If so, the flip to the downsampled JSON leaves all of that original dense time series json untouched.

PR #356, opened 2016-06-16
handling missing corrected data 

lindsayplatt:
@mhines-usgs can you try this out - I will send JSON via slack


mhines-usgs:
@lindsaycarr yes!

lindsayplatt:
merging my own changes since @mhines-usgs already verified the updates work!

PR #357, opened 2016-06-17
fix for AQCU-735

mhines-usgs:
applyDates was null sometimes and causing problems, so if applyDates was null, I turned them into NAs so it would properly shape into a df

### DIFF
--- a/R/correctionsataglance-data.R
+++ b/R/correctionsataglance-data.R
@@ -126,12 +126,14 @@ formatDataList <- function(dataIn, type, ...){
     return()
   }
   
-  if(type == 'APPROVALS' && length(dataIn)>0){
-    for (i in 1:length(dataIn$endTime)) {
-      if (as.Date(dataIn$endTime[i]) > "2050-12-31") {
-        endT <- format(as.Date(dataIn$endTime[i]), format="%m-%d")
-        endT <- paste0("2050-",endT,"T00:00:00.000-06:00")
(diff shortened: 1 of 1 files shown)


R/correctionsataglance-data.R
```diff
@@ -223,6 +227,11 @@ findOverlap <- function(dataList){
         dataIn$line_num <- 1
         
         #ordered by applied date for process order data
+        if (isEmptyOrBlank(dataIn$applyDates)) {
+          for (i in 1:length(dataIn)) {
```
lindsayplatt, line 231: You shouldn't need to have the for loop. Just making `dataIn$applyDates <- NA` will work because data.frame will automatically fill in the other NAs. @mhines-usgs I'll go ahead and make a fix for this since you are traveling!

Example:

``` r
dataIn <- list(values = 1:10, dates = 1:10, applyDates = NULL)
as.data.frame(dataIn)

Error in data.frame(values = 1:10, dates = 1:10, applyDates = NULL, check.names = TRUE,  : 
  arguments imply differing number of rows: 10, 0

dataIn <- list(values = 1:10, dates = 1:10, applyDates = NA)
as.data.frame(dataIn)
   values dates applyDates
1       1     1         NA
2       2     2         NA
3       3     3         NA
4       4     4         NA
5       5     5         NA
6       6     6         NA
7       7     7         NA
8       8     8         NA
9       9     9         NA
10     10    10         NA
```

PR #359, opened 2016-06-23
Changes and improvements spawned from the update of DV date formats.

thongsav-usgs:


R/utils-json.R
```diff
@@ -196,7 +194,7 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, plot_type=NULL
       d <- date_index_list[[i]]
       
       if(isDV){
-        applicable_dates <- points[['time']][d] - hours(12)
+        applicable_dates <- points[['time']][d]
```
lindsayplatt: we should pull this line out of the if-else since it happens for both DV and not DV now

R/uvhydrograph-data.R
```diff
@@ -307,9 +259,9 @@ getInverted <- function(data, renderName, plotName) {
     
   } else if (plotName == "secondary") {   
     dataName <- switch(renderName,
-                       corr_UV2 = "downsampledSecondarySeries",
-                       est_UV2 = "downsampledSecondarySeries",
-                       uncorr_UV2 = "downsampledSecondarySeriesRaw")
```
lindsayplatt, line 312: didn't we just change all of these?

scratchpad.R
```diff
@@ -3,4 +3,5 @@ library(lubridate)
 library(dplyr)
 library(gsplot)
 library(repgen)
-fiveyeargwsum(fromJSON(system.file('extdata','fiveyeargwsum','fiveyeargwsum-with-discrete-example.json', package = 'repgen')), 'html', "Author")
+vdiagram(fromJSON(system.file('extdata','vdiagram','vdiagram-example.json', package = 'repgen')), 'html', "Author")
```
lindsayplatt, line 6: did you mean commit this file??

R/utils-json.R
thongsav-usgs: derp, yep. will do.

R/uvhydrograph-data.R
thongsav-usgs, line 312: Yes, but unfortunately the java side does not attached inverted flags to the secondary downsampled stuff. I did not update the java to do that because I think we might (later) remove the inverted flags from the downsampled series all together. It's wasted processing on the backend if we continue to include the full resolution version.

scratchpad.R
thongsav-usgs, line 6: Yes/no, it was there as a helper for my eclipse IDE setup, I accidentally checked it in while ago and decided to keep it. "Accidentally" committed the update again.

lindsayplatt:
@jread-usgs we are using gsplot 0.5.5 and are getting an error with lazyeval: `Error in lazy_dots(...) : Promise has already been forced`. Looks like you might have investigated this before, any thoughts? We are thinking this might just be fixed when we upgrade, but that won't happen yet

https://github.com/hadley/lazyeval/issues/61
https://github.com/USGS-R/gsplot/pull/335

lindsayplatt:
recreated the error we were having in repgen, and tested it on gsplot 0.5.5 and gsplot 0.6.1 and the issue is indeed fixed in the newest version

``` r
gs <- gsplot() %>%
    error_bar(x = c(as.POSIXct("2013-10-18 09:23:17"), as.POSIXct("2013-10-24 12:26:20")),
              y = c(40500, 47900),
              y.low = c(3240, 3832),
              y.high = c(3240, 3832),
              col = "black",
              lwd = 0.7,
              epsilon = 0.1,
              legend.name = 'Discharge measurement and error')
```

PR #360, opened 2016-06-28
estimated date range edges are now exclusive

thongsav-usgs:


R/utils-json.R
```diff
@@ -218,6 +219,12 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
   return(approvals_all)
 }
 
+subsetByMonth <- function(pts, onlyMonth) {
```
lindsayplatt, line 219: doesn't this function already exist - https://github.com/USGS-R/repgen/blob/master/R/uvhydrograph-data.R#L165-L170? Looks like it needs to be pulled out into `utils-shared.R`

R/utils-json.R
```diff
@@ -231,21 +238,12 @@ getApprovalDates <- function(data, chain_nm, approval){
...
-      format <- "Ymd"
-    } else {
-      format <- "Ymd HMOS z"
-    }
-    
-    time <- parse_date_time(x,format, tz=ts$reportMetadata$timezone,quiet = TRUE)
-    if(isDV) { #want to render DVs at noon, convert to string to add noon and convert back to POSIXct
-      time <- time + hours(12)
-    }
+    time=flexibleTimeParse(x, ts$reportMetadata$timezone)
```
lindsayplatt: ooo I like this! Might want to follow the current convention of `<-` instead of `=` though

R/utils-shared.R
```diff
@@ -95,13 +95,17 @@ isLogged <- function(all_data, ts_data, series){
 
 ############ used in dvhydrograph-data, correctionsataglance-data, fiveyeargwsum-data, uvhydrograph-data ############ 
 
-formatDates <- function(char_date, type=NA, isDV=FALSE){
-  if(isDV) {
-    format <- "%F"
-  } else {
-    format <- "%FT%T"
-  }
+formatDates <- function(char_date, type=NA){
```
lindsayplatt, line 98: :+1:

R/utils-json.R
thongsav-usgs, line 219: oops, yes, I meant to delete from UV hydro

PR #361, opened 2016-06-30
fix for rendering failure when comments are of length 0

mhines-usgs:
This could be a bandaid for some bad json?

see https://internal.cida.usgs.gov/jira/browse/AQCU-758 for json causing issue

Update: maybe not bad JSON, just coming in to the function kind of oddly? but this seems to handle both cases.

### DIFF
--- a/R/sensorreading-data.R
+++ b/R/sensorreading-data.R
@@ -151,7 +151,7 @@ formatSensorData <- function(data, columnNames){
       '', '', ''
     )
     toRet <- rbind(toRet, data.frame(t(columnRow),stringsAsFactors = FALSE))
-  } #for loop
+  } 
   colnames(toRet) <- columnNames
   rownames(toRet) <- NULL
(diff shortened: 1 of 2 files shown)


R/sensorreading-data.R
```diff
@@ -266,7 +266,7 @@ getSRSQualifiers <- function(inQualifiers) {
 
 getComments <- function(comments) {
   comm <- unlist(comments)
-  if (!isEmpty(comm)) {
+  if (!isEmptyOrBlank(comm) && length(comm)>0) {
```
lindsayplatt: honestly, I would add this to the isEmptyOrBlank function. Replace 206 with `result <- (length(val) == 0 || isEmpty(val) || val=="")`

mhines-usgs: :+1:

mhines-usgs:
Travis build is failing but not by something I changed..

PR #362, opened 2016-07-01
Losing the isDV flag and date format coming back as NA and losing the…

mhines-usgs:
… approval bars. Fix for AQCU-760
https://internal.cida.usgs.gov/jira/browse/AQCU-760


mhines-usgs:
Travis failing on unrelated test failures (uvhydro).

PR #363, opened 2016-07-01
Fix for AQCU-744

mhines-usgs:
removing the time from the start/end times in appr_dates when the report type is dvhydro. otherwise the date index/list of dates was skipping a needed day.

https://internal.cida.usgs.gov/jira/browse/AQCU-744


mhines-usgs:
capturing the commented out tests here: https://internal.cida.usgs.gov/jira/browse/AQCU-761

PR #366, opened 2016-07-05
small changes for new version of gsplot

mhines-usgs:


mhines-usgs:
the NAMESPACE file here is bad, I'm closing and will re-open with proper files. Gar.

lindsaycarr:
strange, I didn't need to load the yaml package to render vdiagram or uvhydrograph

PR #367, opened 2016-07-05
small changes for new version of gsplot

mhines-usgs:
new Rd for flexibleTimeParse function from PT


lindsayplatt:
strange, I didn't need to load the yaml package to render vdiagram or uvhydrograph

mhines-usgs:
Why Travis, why?!

mhines-usgs:
I wonder what's up. I just undid and re-ran everything and today I'm not getting a yaml load config error. :/ No idea. Reverting and then at least I don't have to worry about Travis.

PR #369, opened 2016-07-06
Update travis.yml

mhines-usgs:
Get past a build issue on Travis I created.


mhines-usgs:
Unnecessary and will close without merging, my other builds were failing because I deleted a branch. Not sure how to resolve that problem but at least beyond this one.

PR #372, opened 2016-07-08
adding time back to points based on feedback from AQCU-744

mhines-usgs:
@lindsaycarr will this cause negative impacts on the other reports which use the getApprovals function?

For AQCU-744, in order to expand the approval lines at the bottom of the graph into the "day" rather than stop on the start of the day because it only has a date to work with that it's grabbing from the firstDownChain dataset, I'm adding in the + hours + minutes to push it into the day itself. 

I'm asking Laura to verify it looks right on the ticket, too. Just wondering if this will create erroneous representations for the other reports and if so, is there an alternative way to achieve this without impacting those.

### DIFF
--- a/R/utils-json.R
+++ b/R/utils-json.R
@@ -193,7 +193,13 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
     for(i in seq_along(date_index_list)){
       d <- date_index_list[[i]]
       
-      applicable_dates <- points[['time']][d] + hours(23) + minutes(59)
+      title <- getReportMetadata(data, 'title')
+      #result <- !objectName %in% listObjects
+      if (grepl(title, 'DV Hydrograph')) {
+        applicable_dates <- points[['time']][d] + hours(23) + minutes(59)
+      } else {
+        applicable_dates <- points[['time']][d]
+      }
         
       #use the Y values of the points, otherwise line at bottom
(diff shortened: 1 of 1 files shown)


lindsayplatt:
It will change the other reports so that they have the same behavior. If that's desired, than I would say go for it. If not, we need to throw in some logic to only do that for that report (I believe dv?).

R/utils-json.R
```diff
@@ -193,7 +193,13 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
     for(i in seq_along(date_index_list)){
       d <- date_index_list[[i]]
       
-      applicable_dates <- points[['time']][d]
+      title <- getReportMetadata(data, 'title')
+      #result <- !objectName %in% listObjects
+      if (grepl(title, 'DV Hydrograph')) {
+        applicable_dates <- points[['time']][d] + hours(23) + minutes(59)
```
thongsav-usgs, line 199: This DV specific logic should be moved up and out of the function. The function should remain generic and should not have any hidden magic gnome logic.

To do this, wherever DV hydrograph calls this function, that should be hte place that adds the hours/minutes back in if needed.

mhines-usgs, line 199: so make the changes in dvhydrograph-data before it comes to getApprovals? I can try.

PR #374, opened 2016-07-12
Tweaks to re-generalize getApprovals and still achieve AQCU744

mhines-usgs:
@thongsav-usgs 

Not sure this is the ideal way you want it but it does generalize it and still work. If you have something else in mind (going all the way back to dvhydrograph-data.R?) perhaps we should look at this together. I'm not sure your comment from slack would work there, though.

> if possible, in DV hydro code:
> 
> ```
> dates <- formatDates
> dates <- dates + hours(23) + minutes(59)
> ```


thongsav-usgs:
This is fine for now as it removes DV specifics from utils. We can look later at pushing that parameter up.

PR #376, opened 2016-07-15
A fix for AQCU-770

mhines-usgs:
A bit too aggressive for the uvhydrograph corrections, because the value for those correction comments are NA and this statement in its previous state removes them from the allVars variables

I tested this with five year gw summary and dvhydrograph, because they also use the function but it doesn't seem to affect the test datasets I have here, hoping @lindsaycarr can take a peek before I merge.

### DIFF
--- a/R/utils-shared.R
+++ b/R/utils-shared.R
@@ -219,5 +219,5 @@ isEmptyOrBlank <- function(val = NULL, listObjects = NULL, objectName = NULL){
 isEmptyVar <- function(variable){
   all(!is.null(variable), 
       nrow(variable) != 0 || is.null(nrow(variable)), 
-      length(variable$time[!is.na(variable$time)]) != 0 & length(variable$value) != 0)
+      length(variable$time[!is.na(variable$time)]) != 0)
 }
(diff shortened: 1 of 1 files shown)


R/utils-shared.R
```diff
@@ -219,5 +219,5 @@ isEmptyOrBlank <- function(val = NULL, listObjects = NULL, objectName = NULL){
 isEmptyVar <- function(variable){
   all(!is.null(variable), 
       nrow(variable) != 0 || is.null(nrow(variable)), 
-      length(variable$time[!is.na(variable$time)]) != 0 & length(variable$value[!is.na(variable$value)]) != 0)
+      length(variable$time[!is.na(variable$time)]) != 0 & length(variable$value) != 0)
```
lindsayplatt: I'm wondering if we need to check `time` and `value`? If there are non-NA times, then we would consider that variable not empty. Not really sure we need to check for length of `variable$value` as well. Are there really cases where there are non-NA times, but there are no values listed? That would seem like bad data

lindsayplatt: So, I am suggesting deleting the check for `value` and only keeping `length(variable$time[!is.na(variable$time)]) != 0`

mhines-usgs: Ok, I'll give that a whirl. I'm not sure if we would need to check both either, but can test that out and ensure at least that the test cases we have set up still appear as expected. Thanks!

mhines-usgs:
my testing on uvhydro, dvhydro and five year seem just fine with this change. Merging.

PR #377, opened 2016-07-18
uvhydro: secondary plot values were shifted compared to primary

lindsayplatt:


lindsayplatt:
@mhines-usgs review please?? Only one line changed...

mhines-usgs:
Works fine for me, :+1:

PR #378, opened 2016-07-21
Modifying hyperlinks to open in new tabs in the browser

mhines-usgs:
https://internal.cida.usgs.gov/jira/browse/AQCU-771

### DIFF
--- a/R/uvhydrograph-data.R
+++ b/R/uvhydrograph-data.R
@@ -171,7 +171,7 @@ getSimsUrl<- function(data){
   if(is.null(url) || url == '') {
     url <- "SIMS URL: NA"
   } else {
-    url <- paste("SIMS URL:", url) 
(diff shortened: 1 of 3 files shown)


lindsayplatt:
@mhines-usgs is there a specific reason why the sims url for uvhydro is changed in uvhydrograph-data, but it's changed in the markdown file for dvhydrograph and fiveyrgwsum?

mhines-usgs:
I honestly don't know, just changed what was there, eg uvhydro had the function in data instead of just reading in to the Rmd directly. Should I make uvhydro like the others instead to be consistent?

lindsayplatt:
I think that would probably be the best way to handle it. Or make a function? That way you can keep uvhydro's check for NULL urls and use it for dvhydro and fiveyr

mhines-usgs:
Ok, so make a function in utils-shared?

lindsayplatt:
:+1:

PR #385, opened 2016-08-04
AQCU-769: Corrected bug in previous patch.

ashalper-usgs:


lindsayplatt:
Yep, makes since that we need to add in the extra check :+1:

PR #386, opened 2016-08-15
AQCU-785: "Secondary" TS Nomenculture Modification

zmoore-usgs:
Modified the UV Hydrograph R code to remove all instances of "Secondary" TS and replace them with "Reference" (or "Upchain" in the case of the Q Hydrograph) and added code for rendering the "Upcahin" TS in the same manner as "Reference."

### DIFF
--- a/R/uvhydrograph-data.R
+++ b/R/uvhydrograph-data.R
@@ -42,35 +42,25 @@ parseUVData <- function(data, plotName, month) {
     approvals <- append(approvals, approvals_dv_min)
   }
   
-  if(plotName == "reference"){
+  if(plotName == "secondary"){
     
-    corr_UV2 <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeries"), month)
-    est_UV2 <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeries", estimatedOnly=TRUE), month)
-    uncorr_UV2 <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeriesRaw"), month)
-    
-    series_corr2 <- subsetByMonth(getCorrections(data, "referenceSeriesCorrections"), month)
-    
-    effect_shift <- subsetByMonth(getTimeSeries(data, "effectiveShifts"), month)
-    gage_height <- subsetByMonth(getMeanGageHeights(data), month)
-    gw_level <- subsetByMonth(getGroundWaterLevels(data), month)
-    meas_shift <- subsetByMonth(getFieldVisitMeasurementsShifts(data), month)
-    
-    ref_readings <- subsetByMonth(getReadings(data, "reference"), month)
-    csg_readings <- subsetByMonth(getReadings(data, "crestStage"), month)
-    #hwm_readings <- subsetByMonth(getReadings(data, "waterMark"), month)
-    
-    approvals <- getApprovals(data, chain_nm="downsampledReferenceSeries", legend_nm=getTimeSeriesLabel(data, "downsampledReferenceSeries"),
-                              appr_var_all=c("appr_approved", "appr_inreview", "appr_working"),
-                              subsetByMonth=TRUE, month=month, point_type=73)
-  }
+    #Reference Time Series Data
+    corr_UV_ref <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeries"), month)
+    est_UV_ref <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeries", estimatedOnly=TRUE), month)
+    uncorr_UV_ref <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeriesRaw"), month)
+    series_corr_ref <- subsetByMonth(getCorrections(data, "referenceSeriesCorrections"), month)
+    approvals_ref <- getApprovals(data, chain_nm="downsampledReferenceSeries", legend_nm=getTimeSeriesLabel(data, "downsampledReferenceSeries"),
+                                  appr_var_all=c("appr_approved", "appr_inreview", "appr_working"),
+                                  subsetByMonth=TRUE, month=month, point_type=73)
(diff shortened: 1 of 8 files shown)


R/uvhydrograph-render.R
```diff
@@ -157,22 +161,100 @@ createSecondaryPlot <- function(data, month){
...
     } else {
-      status_msg <- paste('Corrected data missing for', data$reportMetadata$secondaryParameter)
+      status_msg <- paste('Corrected data missing for', data$reportMetadata$referenceParameter)
+    }
+  } 
+  
+  return(list(plot=plot_object, table=table, status_msg=status_msg))
+}
+
+createUpchainPlot <- function(data, month){
```
lindsayplatt: seems like this entire function is a copy of the `createReferencePlot` function, other than using the word upchain rather than reference. I think that is a good indication that we should create one utils function that uses a more generic name, and I think keeping secondary plot could be fine - secondary referring to upchain or reference.

thongsav-usgs:
I'm not quite sure I wanted to go this far on the renaming @lindsaycarr and @zmoore-usgs. We just wanted to be able to consume the new incoming parameters (upchain and reference params), but still place them on either the primary and/or secondary plots. That is, in our code, we should still have two plots (primary/secondary), and our logic should pick and choose where to place upchain and reference series.

lindsayplatt:
Ok, I thought that is where we had left that discussion but I thought I might have missed a change last week. So really the only difference in our R code is where it is referencing the name of data within the JSON because that name will be different now.

thongsav-usgs:
That is correct, for now we would like to plot either upchain/reference json fields onto the secondary plot. We have upcoming work to dynamically choose the primary/seconary plot for the reference json, but that can be ignored for now.

zmoore-usgs:
Oh okay, the wording on the ticket made it sound like the renaming of all "secondary" to "reference" should continue into the R code but yeah having one function to handle both with just a little added logic definitely makes sense to me. I'll make those changes, thanks for the feedback!

lindsayplatt:
Yeah, I don't remember if the ticket was updated based on our discussion on 8/4 or not

R/uvhydrograph-data.R
```diff
@@ -44,24 +44,34 @@ parseUVData <- function(data, plotName, month) {
...
+    approvals_up <- getApprovals(data, chain_nm="downsampledUpchainSeries", legend_nm=getTimeSeriesLabel(data, "downsampledUpchainSeries"),
+                                 appr_var_all=c("appr_approved", "appr_inreview", "appr_working"),
+                                 subsetByMonth=TRUE, month=month, point_type=73)
     
     effect_shift <- subsetByMonth(getTimeSeries(data, "effectiveShifts"), month)
     gage_height <- subsetByMonth(getMeanGageHeights(data), month)
     gw_level <- subsetByMonth(getGroundWaterLevels(data), month)
     meas_shift <- subsetByMonth(getFieldVisitMeasurementsShifts(data), month)
+    approvals <- approvals_ref
+    approvals <- append(approvals, approvals_up)
```
lindsayplatt: You can just do this in one step: `approvals <- append(approvals_ref, approvals_up)`

R/uvhydrograph-render.R
```diff
@@ -167,6 +167,72 @@ createSecondaryPlot <- function(data, month){
       status_msg <- paste('Corrected data missing for', data$reportMetadata$secondaryParameter)
     }
   } 
+
+  isUpchainSeries <- any(grepl("downsampledUpchainSeries", names(data)))
```
lindsayplatt: We shouldn't need any of this because we want the upchain series to plot onto the reference plot, not make a new plot. Since the upchain series was added to parseUVdata, it should already be in the secondaryData list and will be plot automatically

zmoore-usgs: Yeah I was thinking that initially but PT thought I should use different variable names for each set of data (I.E: "corr_UV_ref" and "corr_UV_up" in place of "corr_UV2"). So then the calls for stuff like YAxisInterval need to vary based on whether we're looking at reference or upchain data because the variables we use in the call change (such as "secondaryData$corr_UV_ref" vs "secondaryData$corr_UV_up") so I wasn't sure what the best way to handle that would be aside from separating it out like this?

lindsayplatt: I think you could just pass both values in as a vector since it's determining max/min of data. We want these to be on the same plot (that's what I gathered from the ticket)

``` r
secCorr <- c(secondaryData$corr_UV_ref$value, secondaryData$corr_UV_up$value)
secUncorr <- c(secondaryData$uncorr_UV_ref$value, secondaryData$uncorr_UV_up$value)
YAxisInterval(secCorr, secUncorr)
```

R/uvhydrograph-render.R
```diff
@@ -102,9 +102,10 @@ createSecondaryPlot <- function(data, month){
   table <- NULL
   status_msg <- NULL
   
-  isSecondarySeries <- any(grepl("downsampledSecondarySeries", names(data)))
+  isReferenceSeries <- any(grepl("downsampledReferenceSeries", names(data)))
+  isUpchainSeries <- any(grepl("downsampledUpchainSeries", names(data)))
```
lindsayplatt, line 106: @thongsav-usgs if there is no reference timeseries, but there is an upchain timeseries do we plot the data? That might be a question for Laura.

lindsayplatt:
:+1: just want to get clarification on the logic for determining whether or not to render a plot based on what data is present. Then, good to go! So, I'll pass to PT since he has more knowledge about that :)

R/uvhydrograph-render.R
thongsav-usgs, line 106: Yes, sometimes (eg: for discharge hydrographs), you'll get an upchain series and not a reference series. For others, you might get both, but alsojust the reference series.

PR #387, opened 2016-08-15
Increase buffer between edge of chart and max/min of data

ashalper-usgs:
Same behavior as previous patch, with slightly better modularization and style.

### DIFF
--- a/R/uvhydrograph-render.R
+++ b/R/uvhydrograph-render.R
@@ -3,7 +3,7 @@
 #'@param data coming in to create a plot
 #'@rdname uvhydrographPlot
 uvhydrographPlot <- function(data) {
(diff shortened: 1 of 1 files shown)


lindsayplatt:
So none of the functionality changed, but should it based on Laura's comments on the JIRA ticket? Or are you implementing that in a different PR?

ashalper-usgs:
No, no feature changes in this patch. Only code coherence, and re-factoring things to be more serviceable for the next time. It's a bit of a tricky work-flow problem, and we already know the users are going to want more control over specifying the _y_-axis real interval. This patch puts things in a better position for future development.

Laura has compared the results with QA, and I think she approves.

PR #388, opened 2016-08-16
WIP for AQCU-775

mhines-usgs:
@lindsaycarr I feel like this is kind of right, but I haven't been able to get it to show up on the plot. 

I essentially branched the getApprovals function into two chunks, should the approvals show at the bottom? If not - it's daily values, so do the regular plot of daily points like before. 

If it is at the bottom, I branched that to make sure we get the date with times (note in the code about that) and then created a rectangle with that information, using your suggestion in the ticket for what it needs. However, I'm not seeing it show up. Can you take a look and suggest what's missing or wrong? Thanks!


R/uvhydrograph-styles.R
```diff
@@ -20,9 +20,9 @@ getUvStyle <- function(data, info, correctionLabels, plotName) {
                               points=list(x=x, y=y, pch = 21, bg = 'black', col = 'black', cex = .8, lwd=1),
                               callouts=list(x=x, y=y, labels = data$meas_Q$n, cex = .75, col='red', length = 0.05)),
 
-                appr_approved_uv = list(points = list(x=x, y=y, col="lightskyblue", type='p', pch=data[[1]]$point_type, cex=2.5, bg="lightskyblue", legend.name=legend.name, where='first')),
-                appr_inreview_uv = list(points = list(x=x, y=y, col="yellow2", type='p', pch=data[[1]]$point_type, cex=2.5, bg="yellow2", legend.name=legend.name, where='first')),
-                appr_working_uv = list(points = list(x=x, y=y, col="lightpink", type='p', pch=data[[1]]$point_type, cex=2.5, bg="lightpink", legend.name=legend.name, where='first')),
+                appr_approved_uv = list(rect(x0=x0, x1=x1, y0=y0, y1=y1+(info$lims_UV$ylim*0.10)), col="lightskyblue", bg="lightskyblue", legend.name=legend.name, where='first'),
```
mhines-usgs: the y1+(info$lims_UV$ylim*0.10) is just a wild guess for now, to give it some height

lindsayplatt:
I'm investigating. All of the UV approvals are getting removed from allVars on line 72 of uvhydrograph-data.R . Something funky going on in isEmptyVar? I'll look at it tomorrow, but I'm headed home now.

lindsayplatt:
Ah-ha! So at the end of `parseUVData`, we use `isEmptyVar` to remove empty list items. However, that is looking for a field called time which out approved UV lists no longer have now that they are rectangles. So, we need to change that function or have it skip anything with "uv" in the name.

https://github.com/USGS-R/repgen/blob/master/R/utils-shared.R#L218

R/utils-json.R
```diff
@@ -168,53 +168,73 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
+      
+      if (data[[chain_nm]]$approvals$description == approval && length(data[[chain_nm]]$approvals)>0) {
+        #every time I try using getApprovalDates it drops the time I need, so I'm hacking it in here in the meanwhile
+        startTime <- strftime(data[[chain_nm]]$approvals$startTime, format="%Y-%m-%dT%H:%M%S", usetz=TRUE)
+        endTime <- strftime(data[[chain_nm]]$approvals$endTime, format="%Y-%m-%dT%H:%M%S", usetz=TRUE)
+        appr_dates <- data.frame(startTime=startTime, endTime=endTime, stringsAsFactors = FALSE)
+      }
+      
+      if (length(appr_dates)>0) {
+        approval_info[[approval]] <- list(x0 = appr_dates[1,1], x1 = appr_dates[1,2], y0 = substitute(getYvals_approvals(plot_object, length(appr_dates))), y1 = substitute(getYvals_approvals(plot_object, length(appr_dates))), legend.name = paste(approval, legend_nm), time=appr_dates[1,1]) ##added a fake time var to get through a future check
```
lindsayplatt: @mhines-usgs it doesn't seem like you we need to do any of the plot object altering stuff below. Does it work if you just stick a call to addHeight in here? addHeight would already account for inverted axes (the yheight is + or - appropriately), but there would need to be some way to handle logged axes.

```
y1 = substitute(getYvals_approvals(plot_object, length(appr_dates)) + addHeight(plot_object))

addHeight <- function(object){
         yheight <- (object$side.2$lim[2]-object$side.2$lim[1])*0.03 
         return(yheight)
}
```

mhines-usgs: I can try but it was giving me an error like can't add a numeric to an object or something (presumably because it's unevaluated)

R/uvhydrograph-styles.R
```diff
@@ -1,11 +1,19 @@
 
 getUvStyle <- function(data, info, correctionLabels, plotName) {
-  x <- data[[1]]$time
-  y <- data[[1]]$value
+  if (exists('value', where=data[[1]])) { #point data
+    x <- data[[1]]$time
+    y <- data[[1]]$value
+  } else { #rectangle
```
lindsayplatt: I don't think you need any of this - you can just stick data[[1]]$x0 into the list directly rather than making an entirely new object. For the rectangles, the x and y values that already exist (x <- data[[1]]$time) will just be NULL. Which should never cause an error because the rectangle calls don't even use x and y

R/uvhydrograph-styles.R
```diff
@@ -20,9 +28,9 @@ getUvStyle <- function(data, info, correctionLabels, plotName) {
...
 
-                appr_approved_uv = list(points = list(x=x, y=y, col="lightskyblue", type='p', pch=data[[1]]$point_type, cex=2.5, bg="lightskyblue", legend.name=legend.name, where='first')),
-                appr_inreview_uv = list(points = list(x=x, y=y, col="yellow2", type='p', pch=data[[1]]$point_type, cex=2.5, bg="yellow2", legend.name=legend.name, where='first')),
-                appr_working_uv = list(points = list(x=x, y=y, col="lightpink", type='p', pch=data[[1]]$point_type, cex=2.5, bg="lightpink", legend.name=legend.name, where='first')),
+                appr_approved_uv = list(rect = list(xleft=x0, xright=x1, ybottom=y0, ytop=y1, yheight=yheight, col="lightskyblue", bg="lightskyblue", legend.name=legend.name, where='first')),
+                appr_inreview_uv = list(rect = list(xleft=x0, xright=x1, ybottom=y0, ytop=y1, yheight=yheight, col="yellow2", legend.name=legend.name, where='first')),
+                appr_working_uv = list(rect = list(xleft=x0, xright=x1, ybottom=y0, ytop=y1, yheight=yheight, col="lightpink", bg="lightpink", legend.name=legend.name, where='first')),
```
lindsayplatt: you shouldn't stick yheight in here - it will just cause a bunch of warnings about yheight not being a thing

R/uvhydrograph-render.R
```diff
@@ -72,6 +72,11 @@ createPrimaryPlot <- function(data, month){
         plot_object <- extendYaxisLimits(plot_object, primaryStyles[[err]])
       }
       
+      which_rect <- grep('rect', names(primaryStyles))
```
lindsayplatt: if the implementation of addHeight I commented above works then this and the for loop for secondary can be deleted

R/utils-json.R
lindsayplatt: hmmm if it's unevaluated it should allow you to put whatever you want, but could fail when it is actually evaluated which is in the do.calls in uvhydro render

mhines-usgs: ok, i'm trying it now!

R/uvhydrograph-styles.R
mhines-usgs: ok, trying that out

mhines-usgs: Nice, you're so spot on! :+1: x :100:

mhines-usgs:
So with Lindsay's suggested changes to the formatDate function, plus, modifying the dates in the json file, I get rectangles. 

The format that works for dates is:

```
"startTime": "2016-05-01T00:00:00Z",
"endTime": "2016-06-30T23:59:59.999999999Z"
```

this doesn't work:

```
"startTime": "2014-10-01T00:00:00-05:00",
"endTime": "9999-12-31T23:59:59.9999999Z"
```

I've been seeing this difference for a while, and I think it's a bug somewhere upstream in the json generation? @thongsav-usgs & @lindsaycarr ?

mhines-usgs:
image

mhines-usgs:
although, i seem to always get pink bars, and need to revisit the height determination.

lindsayplatt:
@mhines-usgs the startTime having -5:00 is the bug? I noticed endTime has Z in both of those, but how do you know which timezone?

R/uvhydrograph-styles.R
```diff
@@ -20,9 +21,9 @@ getUvStyle <- function(data, info, correctionLabels, plotName) {
                               points=list(x=x, y=y, pch = 21, bg = 'black', col = 'black', cex = .8, lwd=1),
                               callouts=list(x=x, y=y, labels = data$meas_Q$n, cex = .75, col='red', length = 0.05)),
 
-                appr_approved_uv = list(points = list(x=x, y=y, col="lightskyblue", type='p', pch=data[[1]]$point_type, cex=2.5, bg="lightskyblue", legend.name=legend.name, where='first')),
-                appr_inreview_uv = list(points = list(x=x, y=y, col="yellow2", type='p', pch=data[[1]]$point_type, cex=2.5, bg="yellow2", legend.name=legend.name, where='first')),
-                appr_working_uv = list(points = list(x=x, y=y, col="lightpink", type='p', pch=data[[1]]$point_type, cex=2.5, bg="lightpink", legend.name=legend.name, where='first')),
+                appr_approved_uv = list(rect = list(xleft=data[[1]]$x0, xright=data[[1]]$x1, ybottom=data[[1]]$y0, ytop=data[[1]]$y1, col="lightskyblue", bg="lightskyblue", legend.name=legend.name, where='first')),
```
lindsayplatt: just for style, you can add `border=NA` to the list of rect arguments for that black line to go away

mhines-usgs:
Thanks, yeah, it wants the times with Z instead of -whatever, but I presume that is actually local time...

mhines-usgs:
I could be wrong about the Z time. Man this is confusing.

Could it be when it's mixed it doesn't like it, but if it's consistent it might be ok?

thongsav-usgs:
Z is UTC and I think there is a line of code somewhere which converts from UTC to the timezone passed in from the reportMetadata, this is all off the top of my head from working on a previous task a few months ago though, so I don't know if it's still there.

R/utils-json.R
```diff
@@ -168,53 +168,73 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
+      
+      if (data[[chain_nm]]$approvals$description == approval && length(data[[chain_nm]]$approvals)>0) {
+        #every time I try using getApprovalDates it drops the time I need, so I'm hacking it in here in the meanwhile
+        startTime <- formatDates(data[[chain_nm]]$approvals$startTime, format_str="%Y-%m-%dT%H:%M:%S")
+        endTime <- formatDates(data[[chain_nm]]$approvals$endTime, format_str="%Y-%m-%dT%H:%M:%S")
+        appr_dates <- data.frame(startTime=startTime, endTime=endTime, stringsAsFactors = FALSE)
+      }
+      
+      if (length(appr_dates)>0) {
+        approval_info[[approval]] <- list(x0 = appr_dates[1,1], x1 = appr_dates[1,2], y0 = substitute(getYvals_approvals(plot_object, length(appr_dates))), y1 = substitute(getYvals_approvals(plot_object, length(appr_dates)) + addHeight(plot_object)), legend.name = paste(approval, legend_nm), time=appr_dates[1,1]) ##added a fake time var to get through a future check
```
mhines-usgs: this looks like it's overriding the other time frames/spans i think, so it ends up with one time frame, and working always wins. revisiting the for loop

mhines-usgs:
image

R/utils-json.R
```diff
@@ -168,53 +168,74 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
     }
+  } else { #approvals at bottom 
+    approval_info <- list()
+    appr_dates <- NULL
     
-    names(approval_info) <- rep(appr_var, length(date_index_list))
+    startTime <- formatDates(data[[chain_nm]]$approvals$startTime, format_str="%Y-%m-%dT%H:%M:%S")
+    endTime <- formatDates(data[[chain_nm]]$approvals$endTime, format_str="%Y-%m-%dT%H:%M:%S")
+    type <- data[[chain_nm]]$approvals$description
+    type <- gsub("Working","appr_working_uv", type)
```
mhines-usgs: this feels dirty and short-cutty but I can't seem to get them renamed any other way, maybe Lindsay can help.

lindsayplatt: I think a switch statement would be cleaner. This should accomplish what lines 221-224 are doing.

``` r
type <- switch(data[[chain_nm]]$approvals$description,
               "Working" = "appr_working_uv",
               "In Review" = "appr_inreview_uv",
               "Approved" = "appr_approved_uv")
```

lindsayplatt:
:clap: yay!

mhines-usgs:
@lindsaycarr yeah it's a little unsatisfying because there are time problems still in the data that we don't handle as well as odd heights, but .. it's working? making a task to revisit the heights.

lindsayplatt:
still :clap: because that was a lot of new changes, and you even managed to get coverage to increase!

PR #390, opened 2016-08-23
AQCU-756: Plot Reference TS on Primary plot with Primary TS for Q Hydrographs

zmoore-usgs:


thongsav-usgs:
Do the approvals (are they all bars) conflict with each other? I see approvals being pulled for Primary, Reference, and Upchain, but the primary plot should only ever plot the primary approvals. I think the bottom plot would plot either reference or upchain, with upchain taking priority if it exists?

zmoore-usgs:
Ah okay I had set it up so that the primary was plotting approvals from both by just appending them all together but I can change that. I haven't been seeing any approval conflicts but I also haven't been seeing any approval bars on the UV Hydrographs since merging in the latest changes to master, so it's possible there are conflicts that I'm just not seeing. Approval bars don't seem to be showing on UV Hydros on DEV either. 

So instead of plotting all approvals on the primary plot you're thinking I should plot primary approvals on the primary plot and plot the approvals bar for the reference data (but not the reference data itself) on the bottom plot if there is no upchain data there?

thongsav-usgs:
Right, it's strange, but depending on the hydrograph type, the secondary plot is most interested in different things. For discharge, the secondary plot is most interested in the Upchain series, so you would only plot the approvals for that. For ground water or water quality, the bottom plot is interested in reference time series, so the approvals would be for that.

The primary plot is always focused on the primary ts, so approvals for primary are all that need to be up top. I think the same is true for "corrections", and you do not need to pull corrections information for anything you do not display approval bars for.

zmoore-usgs:
Gotcha, that makes sense to me. I do currently have the corrections showing up in different colors on the primary plot for the primary TS and reference TS so they are somewhat differentiated but it can certainly get confusing when both series have a lot of them. I'll go ahead and make changes as you described and see if Laura provides any additional guidance.

EDIT: Hadn't checked JIRA when I posted this. I see that she posted some clarifications so I'll work with those as well.

R/uvhydrograph-data.R
```diff
@@ -166,7 +177,7 @@ correctionsTable <- function(data) {
 
 parseLabelSpacing <- function(data, info) {
   
-  if (names(data) %in% c("series_corr", "series_corr_ref", "series_corr_uv")){
+  if (names(data) %in% c("series_corr", "series_corr_ref", "series_corr_up")){
```
thongsav-usgs: Looks like series_corr_up was removed everywhere else. Is this extra?

thongsav-usgs:
Nevermind last deleted comment. This looks OK code wise but hard for me to recall business logic. Going to merge to allow Laura to verify.

PR #391, opened 2016-08-23
Multiple fixes

mhines-usgs:
CORR: changing check for year 9999 and change from 2050 to 2100.

utils-json: checking for year 9999 and writing to 2100 instead to address problem with R and rendering issue for rectangles

uvhydrograph styles: adding same colored border to rectangles so that the little white gap doesn't appear when it was

### DIFF
--- a/R/utils-json.R
+++ b/R/utils-json.R
@@ -216,19 +216,31 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
     approval_info <- list()
     appr_dates <- NULL
     
-    startTime <- formatDates(data[[chain_nm]]$approvals$startTime, format_str="%Y-%m-%dT%H:%M:%S")
-    endTime <- formatDates(data[[chain_nm]]$approvals$endTime, format_str="%Y-%m-%dT%H:%M:%S")
-    type <- data[[chain_nm]]$approvals$description
-    type <- unlist(lapply(type, function(desc) {
-      switch(desc,
-             "Working" = "appr_working_uv",
-             "In Review" = "appr_inreview_uv",
-             "Approved" = "appr_approved_uv")
-    }))
-    legendnm <- data[[chain_nm]]$approvals$description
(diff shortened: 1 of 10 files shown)


R/utils-json.R
```diff
@@ -216,19 +216,32 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
-    }))
-    legendnm <- data[[chain_nm]]$approvals$description
-    appr_dates <- data.frame(startTime=startTime, endTime=endTime, type=type, legendnm=legendnm, stringsAsFactors = FALSE)
+    if (!isEmpty(data[[chain_nm]]$approvals$startTime)) {
+      startTime <- flexibleTimeParse(data[[chain_nm]]$approvals$startTime, timezone = data$reportMetadata$timezone)
+      endTime <- flexibleTimeParse(data[[chain_nm]]$approvals$endTime, timezone = data$reportMetadata$timezone)
+      #hacky fix for 9999 year issue which prevents the rectangles from displaying on the graphs 
+      #apologies to the people of 2100 who have to revisit this
+      for (i in 1:length(endTime)) {
+        if (as.Date(endTime)[i] > "2100-12-31") {
```
lindsayplatt, line 225: would it make sense to have a helper function that did this wonky 9999 conversion for corr & this? That way, if we ever find a cleaner way of doing it, we will only have to fix it once?

lindsayplatt, line 225: it seems like the two implemented methods are pretty different - so I'm not sure that would work

mhines-usgs: right, the dates are different formats :head-explode:

mhines-usgs: I could try to wrap it into one though, basically, if it's a date only (no time) do this, if it's a date with a time, do it this way or something ... I'll think about it after I figure out this other problem

lindsayplatt: I wonder if you could pass in "time" as an arg? Then `time_str=NULL` by default. That might be cleaner?

PR #395, opened 2016-08-31
splitDataGaps for uvhydro, dvhydro, fiveyrgwsum (AQCU-808)

lindsayplatt:
replaces splitDataGaps which was exclusively for dvhydrograph (see #189) and removes connectTS (see #205)


lindsayplatt:
things before I want this merged:
- [x] ~~is handling list data instead of data frame data a hack? uvhydro has data.frame, but dvhydro & fiveyr have lists~~ this is fine - we have them this way for a reason (uvhydro needs to have data.frame to allow for subsetting by month)
- [x] ~~seems to be non-inclusive with gap dates (endDate for gap is being dropped)~~ was because date formats were different
- [x] need to break for gaps in dates that might not necessarily be in "gaps" because data exists for estimated but not corrected (dvhydro). Use gap tolerence field that @zmoore-usgs added

@thongsav-usgs

lindsayplatt:
Noticed that the legends got lost somewhere along the way....same when you render DV hydro on Andy's branch w/o my changes (#396). So I'm not sure where this was introduced

PR #396, opened 2016-08-31
Correct positioning of approval bars.

ashalper-usgs:


R/utils-json.R
```diff
@@ -263,10 +263,15 @@ subsetByMonth <- function(pts, onlyMonth) {
...
+
 getYvals_approvals <- function(object, num_vals){
   ylim <- ylim(object)$side.2[1]
   yvals <- rep(ylim, num_vals)
-  return(yvals)
+  
+  # The constant scaling factor below was reverse-engineered from analyzing the 
+  # (pandoc?) SVG in Inkscape. I'm not entirely sure why it works, and it's easy
+  # to imagine why it wouldn't in some cases.
+  return(0.908196 * yvals)
```
lindsayplatt: I would leave that function as-is and add one that calculates the user coords

``` r
getYvals_userExtent <- function(object, num_vals){
  #calculate value using 4% thingy
  usr_coords <- par("usr")[3:4]

  # ylim <- final value after calculation (taking log and reverse situations into account)
  # for logged 4% = (log(ylim(object)$side.2[2]) - log(ylim(object)$side.2[1]))*0.04
  # and par("usr")[3] = log(ylim(object)$side.2[1]) - 4%

  yvals <- rep(ylim, num_vals)
  return(yvals)
}
```

then, in `getApprovals`, 

``` r
y0 = substitute(getYvals_approvals(plot_object, 1))

#and

y1 = substitute(getYvals_userExtent(plot_object, 1))
```

lindsayplatt: until we can get a helper-type function the just retrieves the user extent for us

ashalper-usgs: What's the purpose of `num_vals`? Never understood that.

ashalper-usgs: Also, if I leave `getYvals_approvals()` as is, then the approval bar pops up well above the _x_-axis, which is not where I want it. Trying to have the bottom of the approval bar aligned with the _x_-axis line.

ashalper-usgs: Is this the way I check for a log situation?

`isLogged(data, parsedData, "firstDownChain")`

lindsayplatt: The function `rep` needs the value to repeat and then the number of times to repeat it. `num_vals` is the number of points (x values) that exist for approvals. We use this so that we create the exact same number of y values to go with it. 

The approval "line" is actually a rectangle now. `y0` is the lower y value of the rectangle and `y1` is the upper y value of the rectangle. Currently, `y0` is using the values from `getYvals_approvals()` (so the lowest data extent, not plot extent) and `y1` is that number plus something define in `addHeight()`. I'm proposing that `y0` should be the lowest plot extent (using `par("usr")` calculations), and `y1` (the upper part of the rectangle) exist at the lowest data extent. I'm not certain this will work, but that's what I imagine COULD work.

 `isLogged` is used to see if a specific timeseries is logged and if so, indicates that the plot y-axis should be logged. You'd need to check a bunch of the timeseries and determine if any are logged to use that function. The better way is to use either `primaryInfo$logAxis` (and `primaryInfo$isInverted` for reverse y axes) created in uvhydrograph-render **_OR**_ inside the plot object (which is already passed into the `getYvals_approvals()` function) by checking `object$side.2$log` (and `object$side.2$reverse` for reversed y axes).

R/utils-json.R
```diff
@@ -278,10 +281,51 @@ subsetByMonth <- function(pts, onlyMonth) {
...
+  #   times: "times" argument of R rep() function.
+  #   c: A scaling ratio to adjust top or bottom of approval bar rectangle.
+  #
+  # Returns:
+  #   Approval bar, top or bottom y-axis point, in world coordinates.
+  reverse <- object$side.2$reverse
+  
+  # no idea whether semantics of object$global$par$ylog are correct in this
+  # context or not; it just looked promising in the debugger
+  if (is.null(object$global$par$ylog)) {
```
lindsayplatt, line 293: yes, `object$global$par$ylog` is good - but I don't think it will ever be NULL based on what the `isLogged` function returns. I wonder if we should not have logic correcting for when this is NULL so that we can catch that scenario as a bug if it ever happens

R/utils-json.R
```diff
@@ -278,10 +281,51 @@ subsetByMonth <- function(pts, onlyMonth) {
   }
   return(pts)
 }
-getYvals_approvals <- function(object, num_vals){
-  ylim <- ylim(object)$side.2[1]
-  yvals <- rep(ylim, num_vals)
-  return(yvals)
+
+ApprovalInfoY <- function(object, times, c) {
```
lindsayplatt: you should rename "c" to something different because "c" is the name of a function and this could result in some weirdness later on

R/utils-json.R
```diff
@@ -278,10 +281,51 @@ subsetByMonth <- function(pts, onlyMonth) {
...
+  # Compute top and bottom vertical position of approval bars.
+  # 
+  # Args:
+  #   object: A gsplot object.
+  #   times: "times" argument of R rep() function.
+  #   c: A scaling ratio to adjust top or bottom of approval bar rectangle.
```
lindsayplatt: you should convert this documentation to Roxygen and put it right above. This is used by R to create help files for the package

```
#' compute top and bottom vertical position of approval bars.
#' @param object a gsplot object
#' @param times "times" argument of R rep() function.
#' @param sc a scaling ratio to adjust top or bottom of approval bar rectangle.
```

R/utils-json.R
ashalper-usgs: It was NULL in the debugger sometimes, but I don't know under what circumstances. It's there to defensively keep the code from terminating with an error.

lindsayplatt: Right, but I wonder if we would like it to terminate with an error if it comes through as NULL. That might flag us that something else is going wrong. We can always change that later though

R/utils-json.R
ashalper-usgs: OK. Re-named to `ratio`.

R/utils-json.R
ashalper-usgs: Done.

R/utils-json.R
ashalper-usgs: As you like. I have removed it.

lindsayplatt:
@ashalper-usgs once your changes are pushed and checks pass, I'll merge this and move to test for Laura

R/utils-json.R
ashalper-usgs: um...actually, it chokes on `uvhydro-groundwater.json`. How do you want to proceed?

lindsayplatt: nevermind then

lindsayplatt: behavior of gsplot is not what I expected! Thanks though :)

ashalper-usgs: Unsurprising here. NULL considered generally misunderstood. ;)

mhines-usgs:
Lindsay's one suggested change (dvhydro-render lines 27 & 78) yaxs='i' instead of yaxs='r' is missing from this PR

That fixes the first approval bar period. However subsequent periods/approvals get shifted up.

PR #397, opened 2016-09-01
AQCU-774: Does not display aging status for DV correctly if day has mixed aging status.

zmoore-usgs:
Issue was that the approval dots were not using the "lowest" approval status on a given day. For example, sometimes a day would be "Working" until 9:00AM and then switch to "In Review" and a point on that day might be rendered as "In Review" but it should be rendered as the lowest approval status on that day which would be "Working".

I am definitely not confident that this is the most efficient way to accomplish this as I'm still learning R so any feedback on that front would be greatly appreciated!

### DIFF
--- a/R/utils-json.R
+++ b/R/utils-json.R
@@ -169,47 +169,78 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
   approvals_all <- list()
   
   if(approvalsAtBottom==FALSE) {
-    for(approval in appr_type){
-      appr_var <- appr_var_all[which(appr_type == approval)]
+    approved_dates <- getApprovalDates(data, chain_nm, "Approved")
+    review_dates <-getApprovalDates(data, chain_nm, "In Review")
+    working_dates <- getApprovalDates(data, chain_nm, "Working")
+
+    approved_points <- list()
+    review_points <- list()
+    working_points <- list()
       
-      if(subsetByMonth){
-        points <- subsetByMonth(getTimeSeries(data, chain_nm), month)
-      } else {
-        points <- data[[chain_nm]][['points']]
-        points$time <- formatDates(points[['time']], type=NA)
-      }
-      
-      appr_dates <- getApprovalDates(data, chain_nm, approval)
-      date_index <- apply(appr_dates, 1, function(d, points){
-        which(points$time >= d[1] & points$time <= d[2])}, 
-        points=points)
-      
-      if(is.list(date_index)){
-        date_index_list <- date_index
-      } else {
-        date_index_list <- list(date_index)
-      }
-      
+    if(subsetByMonth){
+      points <- subsetByMonth(getTimeSeries(data, chain_nm), month)
+      points_no_times <- points
+      points_no_times$time <- as.POSIXct(strptime(points_no_times$time, "%F"))
+    } else {
+      points <- data[[chain_nm]][['points']]
+      points_no_times <- points
+      points_no_times$time <- as.POSIXct(strptime(points_no_times[['time']], "%F"))
+    }
+
+    working_index<- apply(working_dates, 1, function(d, points){
+      which(points$time >= d[1] & points$time <= d[2])}, 
+      points=points_no_times)
+
+    review_index<- apply(review_dates, 1, function(d, points){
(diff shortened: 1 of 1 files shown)


R/utils-json.R
```diff
@@ -169,47 +169,74 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
+
+    review_index<- apply(review_dates, 1, function(d, points){
+      which(points$time >= d[1] & points$time <= d[2])}, 
+      points=points_no_times)
+
+    approved_index<- apply(approved_dates, 1, function(d, points){
+      which(points$time >= d[1] & points$time <= d[2])}, 
+      points=points_no_times)
+
+    if(is.list(working_index)){
```
lindsayplatt: I would recommend not using `is.list()`. It's tricky -- data.frames are technically "lists" under the hood. Try `class(working_index) == "list")` instead.

Trickiness:

``` r
ex <- data.frame(a=1,b=2)
is.list(ex)
[1] TRUE
```

R/utils-json.R
```diff
@@ -169,47 +169,74 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
+    if(is.list(review_index)){
+      review_index <- unique(unlist(review_index, recursive=FALSE))
+    }
+
+    if(is.list(approved_index)){
+      approved_index <- unique(unlist(approved_index, recursive=FALSE))
+    }
+    
+    review_index <- setdiff(review_index, working_index)
+    approved_index <- setdiff(approved_index, review_index)
```
lindsayplatt, line 198: I've never seen that function - handy!!

R/utils-json.R
```diff
@@ -169,47 +169,74 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
-        points=points)
-      
-      if(is.list(date_index)){
-        date_index_list <- date_index
-      } else {
-        date_index_list <- list(date_index)
-      }
+    approved_dates <- getApprovalDates(data, chain_nm, "Approved")
+    review_dates <-getApprovalDates(data, chain_nm, "In Review")
+    working_dates <- getApprovalDates(data, chain_nm, "Working")
```
lindsayplatt: :+1: for removing that top for loop

R/utils-json.R
```diff
@@ -169,47 +169,74 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
       
+    if(subsetByMonth){
+      points <- subsetByMonth(getTimeSeries(data, chain_nm), month)
+      points_no_times <- points
```
lindsayplatt: 178 & 179 can be moved outside of the logic because they are the same as 182 & 183

lindsayplatt:
It seems like a lot of the code finding the approvals (where you do the same thing for inreview, working, and approved) can be pulled out into one function where the argument is the approval type. That way we can get rid of some duplicated code.

R/utils-json.R
```diff
@@ -169,47 +169,74 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
...
+      which(points$time >= d[1] & points$time <= d[2])}, 
+      points=points_no_times)
+
+    review_index<- apply(review_dates, 1, function(d, points){
+      which(points$time >= d[1] & points$time <= d[2])}, 
+      points=points_no_times)
+
+    approved_index<- apply(approved_dates, 1, function(d, points){
+      which(points$time >= d[1] & points$time <= d[2])}, 
+      points=points_no_times)
```
lindsayplatt: these apply statements can be in a single function

R/utils-json.R
lindsayplatt: these if statements can go in the same function as the applys starting on line 186

zmoore-usgs:
Whoops, used the wrong ticket number in my last commit message.  Should be AQCU-774.

R/utils-json.R
```diff
@@ -164,52 +164,66 @@ getEstimatedDates <- function(data, chain_nm, time_data){
   return(date_index)
 }
 
+getApprovalIndex <- function(data, points, chain_nm, approval, subsetByMonth=FALSE) {
```
lindsayplatt, line 167: :+1:

PR #399, opened 2016-09-07
AQCU-772: Allow extremes report generation for single parameter workflows

zmoore-usgs:


R/extremes-data.R
```diff
@@ -7,47 +7,73 @@
...
+
+  dataRows <- c(maxRows, minRows)
+
+  #Change column and row names to their correct forms and add them into the dataframe.
+  toRet <- data.frame()
   
-  index <- which(names(data) %in% c("upchain", "primary", "dv")) 
-  results <- list()
+  for(i in 1:length(dataRows)){
+    toAdd <- cbind(dataRows[[i]][["name"]],dataRows[[i]][,-1])
```
lindsayplatt: I think this could be simplified to `toAdd_test <- dataRows[[i]]`, right? You cbind the first column and then all but the first column. Wouldn't that be the same as keeping the df as is?

R/extremes-data.R
```diff
@@ -7,47 +7,73 @@
...
-  for (i in index) {  
-    
-    subset <- data[[i]][which(names(data[[i]])%in%c("min","max"))]
-    
-    min.max <- lapply(subset, function(x) {
+  return(toRet)
+}
+
+createDataRows <- function(data, param, rowName, isUpchain, includeRelated=TRUE, doMerge=TRUE){  
+    subset <- data[which(names(data)%in%c(param))]
```
lindsayplatt: rename subset to something else (maybe `subsetData`) because subset is a function name and you are overwriting here. You can check if the name already is a function by typing it and then hitting tab in the console - if no suggestions pop up then you're good

R/extremes-data.R
```diff
@@ -68,92 +94,50 @@ extremesTable <- function(rawData){
...
+
+        if(!is.null(relatedSet))
+        {
+          relatedValue <- relatedSet$value
         }
-        else {
-          primary <- x$points$value  
+
+        if(!isUpchain){
+          data.frame(name=rowName, date=dateTime[,1], time=timeFormatting, primary=primaryValue, related=relatedValue, stringsAsFactors = FALSE)
```
lindsayplatt: this data frame is not being saved as an object!

R/extremes-data.R
```diff
@@ -68,92 +94,50 @@ extremesTable <- function(rawData){
...
+        } else {
+          data.frame(name=rowName, date=dateTime[,1], time=timeFormatting, primary=relatedValue, related=primaryValue, stringsAsFactors = FALSE)
```
lindsayplatt: and neither is this one

R/extremes-data.R
```diff
@@ -68,92 +94,50 @@ extremesTable <- function(rawData){
...
-        
       } else {
-        if (is.null(x$points$value)) {
-          primary <- "N/A"
-        }
-        else {
-          primary <- x$points$value  
-        }
-        upchain  <- "N/A"
+        data.frame(name=rowName, date=dateTime[,1], time=timeFormatting, primary=primaryValue, stringsAsFactors = FALSE)
```
lindsayplatt: or this one

R/extremes-data.R
```diff
@@ -68,92 +94,50 @@ extremesTable <- function(rawData){
...
-      colnames(toAdd) <- columnNames
-    } 
-    toRet <- rbind(toRet,toAdd)
-    
-  }
-  
-  return(toRet)
+
+      #Replace Duplicate Names with blank names
+      if(NROW(dataRows[duplicated(dataRows["name"]),]) > 0){
```
lindsayplatt, line 139: hmm I've never seen that function capitalized before -- but it does the same thing as `nrow` so :+1:

tests/testthat/test-extremes.R
```diff
@@ -37,4 +37,25 @@ test_that("extremes examples work",{
   expect_is(extremes(data, 'html'), 'character')
 })
 
+context("testing example of no upchain")
```
lindsayplatt, line 40: yay tests!! :tada:

zmoore-usgs:
About the data frames not being saved: I based that part off of the original code for the extremes report which was not saving them either. Somehow the data frames are still being passed out through lapply but I'm not really sure how.. I just figured it was a shortcut or something since the original extremes report code did it and it seems to work somehow here as well.

lindsayplatt:
Ah it's because functions return the last thing evaluated by default but I think it's best to be explicit. Our bad for not catching it in previous could, but can you add it? Save to a var and then have one return statement at the very end.

zmoore-usgs:
Ah okay good to know, I wasn't aware of that. Makes a lot more sense now! :smile: Yep, I will make that change!

PR #403, opened 2016-09-15
Last step is detected at end of each series of points (broken up by

thongsav-usgs:
gaps). This fix only extends the step 24 hours versus all the way to end
of data set.


mhines-usgs:
checkin this out

PR #404, opened 2016-09-15
AQCU-808: Fixed an issue where gaps for DV data were too big.

zmoore-usgs:
Also includes a temporary fix for the issue where DV data was being rendered at noon instead of on the date lines. See comment on line 264.

### DIFF
--- a/R/utils-shared.R
+++ b/R/utils-shared.R
@@ -333,10 +333,12 @@ applyDataGaps <- function(data, relevantData, isDV=FALSE){
   if(!isEmptyOrBlank(gapData)){
     pattern <- paste0("(", paste(names(relevantData), collapse="|"), ")")
     names(gapData) <- regmatches(names(gapData), m=regexpr(pattern, names(gapData)))
+    #add data back together
+    relevantDataWithGaps <- append(relevantData[!haveField], gapData)
+  } else {
+    relevantDataWithGaps <- relevantData
   }
   
-  #add data back together
-  relevantDataWithGaps <- append(relevantData[!haveField], gapData)
   return(relevantDataWithGaps)
 }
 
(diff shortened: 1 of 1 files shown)


R/utils-shared.R
```diff
@@ -261,7 +261,8 @@ splitDataGaps <- function(data, ts, isDV){
       endGaps <- c(endGaps, endEstimated)
     }
     
-    if(isDV){ ts$time <- flexibleTimeParse(ts$time, timezone = data$reportMetadata$timezone) }
```
zmoore-usgs, line 264: This line was what was causing the DV data to be rendered at noon instead of on the date lines because flexibleDateTimeParse was adding a time of noon to all of the time fields. Commenting out this line fixed that but it should be revisited after work is done on AQCU-816.

mhines-usgs:
pulling this local to test out

lindsayplatt:
....I'm confused - we went to great lengths awhile ago to have DVs render at noon. Is that not the desired behavior anymore?

thongsav-usgs:
This is for the "Step plots", the steps go noon to noon when they should go from midnight to midnight. For the actual DV points, we sitll want them at noon. For your upcoming task, @lindsaycarr, I'd reinstate the flexibleTimeParse and have the DV hydro code subtract 12. That is, the noon time will be default, and the step plot will be custom setting it back to 0000.

R/utils-shared.R
```diff
@@ -261,7 +261,8 @@ splitDataGaps <- function(data, ts, isDV){
...
+    #This is causing DV values to be rendered at noon instead of on the day marks. Re-enable after AQCU 816?
```
thongsav-usgs: I try not to reference JIRA tickets in the source. The source code is public and the JIRA is not, so it provides little historical value.

PR #405, opened 2016-09-16
Aqcu 818

kmschoep-usgs:
lines 267-270, should I also genericize the max_DV, mean_DV, etc. variables? Feels sloppy not to, but not sure of downstream implications and how much that would expand scope.


lindsayplatt:
@kmschoep-usgs use CTRL+SHIFT+F to search all scripts for where those occur - might only be one other place

R/uvhydrograph-data.R
```diff
@@ -23,23 +23,23 @@ parseUVData <- function(data, plotName, month) {
...
                                             subsetByMonth=TRUE, month=month, point_type=24, approvalsAtBottom=FALSE)
-    approvals_dv_mean <- getApprovals(data, chain_nm="derivedSeriesMean", legend_nm=paste("DV Mean", getTimeSeriesLabel(data, "derivedSeriesMean")),
+    approvals_first_stat <- getApprovals(data, chain_nm="firstDownChain", legend_nm=data[['reportMetadata']][["downChainDescriptions1"]],
                                             appr_var_all=c("appr_approved_dv", "appr_inreview_dv", "appr_working_dv"), 
                                             subsetByMonth=TRUE, month=month, point_type=21, approvalsAtBottom=FALSE)
-    approvals_dv_median <- getApprovals(data, chain_nm="derivedSeriesMedian", legend_nm=paste("DV Median", getTimeSeriesLabel(data, "derivedSeriesMedian")),
+    approvals_fourth_stat <- getApprovals(data, chain_nm="fourthDownChain", legend_nm=data[['reportMetadata']][["downChainDescriptions4"]],
                                             appr_var_all=c("appr_approved_dv", "appr_inreview_dv", "appr_working_dv"), 
-                                            subsetByMonth=TRUE, month=month, point_type=26, approvalsAtBottom=FALSE)
```
lindsayplatt, line 34: huh, I wonder how 26 did not cause a problem before - it's not actually a supported point type at this time

R/uvhydrograph-data.R
```diff
@@ -23,23 +23,23 @@ parseUVData <- function(data, plotName, month) {
...
-    approvals_dv_min <- getApprovals(data, chain_nm="derivedSeriesMin", legend_nm=paste("DV Min", getTimeSeriesLabel(data, "derivedSeriesMin")),
+                                            subsetByMonth=TRUE, month=month, point_type=22, approvalsAtBottom=FALSE)
+    approvals_third_stat <- getApprovals(data, chain_nm="thirdDownChain", legend_nm=data[['reportMetadata']][["downChainDescriptions3"]],
```
lindsayplatt, line 32: We should probably have the code ordered correctly: first, second, third, fourth. Seems weird to be out of order

lindsayplatt: and have that carried out through the rest of this section

R/uvhydrograph-data.R
kmschoep-usgs: I checked on QA and those points don't render.

R/uvhydrograph-data.R
kmschoep-usgs: I can do that.

R/uvhydrograph-data.R
```diff
@@ -264,10 +264,10 @@ getInverted <- function(data, renderName, plotName) {
...
                        comp_UV = "downsampledComparisonSeries",  
                        water_qual = "downsampledPrimarySeries",  #if primary is flipping, this will flip
-                       max_DV = "derivedSeriesMax",
-                       mean_DV = "derivedSeriesMean",
-                       median_DV = "derivedSeriesMedian",
-                       min_DV = "derivedSeriesMin")
+                       stat_2 = "secondDownChain",
+                       stat_1 = "firstDownChain",
+                       stat_4 = "fourthDownChain",
+                       stat_3 = "thirdDownChain")
```
lindsayplatt: testing the review feature

lindsayplatt (changes requested):
change the order and then we are good to merge!

R/uvhydrograph-data.R
lindsayplatt: ah, so we just did not notice before

lindsayplatt (approved):

thongsav-usgs:
Only thing I'd suggest is updating all the test data (I'd be fine doing a text search/replace rather than regenerating all of the data). As of now, the build passes because the those fields are optional and we aren't actually rendering them anymore in the tests.

PR #409, opened 2016-09-19
Fixed issue where tick marks labeled with just the day are adjusted to

thongsav-usgs:
be at midnight.


lindsayplatt (commented):
is this going to have repercussions on the work I just did to extend the line to the last day? L22 adds 23h 59m and then it is removed in L25...

thongsav-usgs:
Just to recap what we talked about in person. Looks like you used endDates for that functionality and not plotDates. plotDates should probably be renamed as it only seems to contain data for the tick marks.

PR #411, opened 2016-09-20
tiny changes to labels in AQCU-823

mhines-usgs:
changing to legend.name="Measured Water Level (GWSI)")  instead of Measured Water Level (NWIS)


kmschoep-usgs:
Dumb question, are there ever water levels from NWIS-RA and GWSI or are they all from GWSI?

thongsav-usgs:
There are from NWIS-RA but that is NATDB and doesn't GWSI feed NATDB? You should be able to find examples though, see the example data file for groundwater.

mhines-usgs:
I don't see anything specific in the example data/report that Kathy/Laura suggested that would indicate where they're from

thongsav-usgs:
https://github.com/USGS-R/repgen/blob/master/inst/extdata/uvhydrograph/uvhydro-groundwater.json#L1010

The example file (when it was pulled) had one gwLevel record.

mhines-usgs:
I think what Kathy is asking/ me too, is can we tell if they all should be from GWSI or will sometimes they be from NWIS-RA somehow?

The legend was hardcoded NWIS-RA, so I changed to GWSI, but is there a parameter we can key from in the metadata? I don't see one.

kmschoep-usgs:
This is a pretty trivial change we can probably just merge in, but confirm with Laura when she's back?

mhines-usgs:
Made a comment on the JIRA task to Laura about it..

ashalper-usgs:
I think NWIS's QWDATA (the legacy Water Quality system) does reference GWSI water levels, but I'd double-check with a Water Quality specialist to be sure.

PR #412, opened 2016-09-20
AQCU-827: Signal going off of the chart

zmoore-usgs:
Fixed an issue where UV Hydrographs were scaling the Y-Axis based off of only the first "chunk" of data when gaps were present. Also inadvertently fixed the issue where the comparison TS would not show up much of the time because the scaling was based off of only the primary TS. Including the comparison TS does make the main (left side) Y-Axis a bit messy however, and cause other scaling issues because it is plotted to the same axis as the primary TS.

I could not find any examples of DV Hydro or 5-Year reports that had scaling issues, including data sets with gaps, so I have only made changes to UV Hydro.

### DIFF
--- a/R/uvhydrograph-data.R
+++ b/R/uvhydrograph-data.R
@@ -19,6 +19,14 @@ parseUVData <- function(data, plotName, month) {
     
     series_corr <- subsetByMonth(getCorrections(data, "primarySeriesCorrections"), month)
     meas_Q <- subsetByMonth(getFieldVisitMeasurementsQPoints(data), month)  
+
+    #Add reference data to the plot if it is available and this is a Q plot type
+    if(any(grepl("Discharge", getReportMetadata(data,'primaryParameter'))))
+    {
+      #Reference Time Series Data
+      corr_UV_Qref <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeries"), month)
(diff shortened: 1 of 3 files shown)


R/uvhydrograph-data.R
```diff
@@ -19,6 +19,14 @@ parseUVData <- function(data, plotName, month) {
...
     series_corr <- subsetByMonth(getCorrections(data, "primarySeriesCorrections"), month)
     meas_Q <- subsetByMonth(getFieldVisitMeasurementsQPoints(data), month)  
+
+    #Add reference data to the plot if it is available and this is a Q plot type
+    if(any(grepl("Discharge", getReportMetadata(data,'primaryParameter'))))
+    {
+      #Reference Time Series Data
+      corr_UV_Qref <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeries"), month)
+      est_UV_Qref <- subsetByMonth(getTimeSeries(data, "downsampledReferenceSeries", estimatedOnly=TRUE), month)
+    }
```
zmoore-usgs, line 29: Note: Moving this block up was to fix an unrelated issue of the reference TS plotting on top of the approval points.

mhines-usgs:
pulling this locally to check it out!

mhines-usgs:
This looks good @zmoore-usgs -- feel free to merge!

PR #413, opened 2016-09-20
moving remove dupes below adding approvals so that any approval dupes…

mhines-usgs:
… are also removed


mhines-usgs:
@lindsaycarr / @ashalper-usgs check it out?

PR #417, opened 2016-09-28
AQCU-661: Axis issues when different parameter than primary

zmoore-usgs:


mhines-usgs:
checking this out now..

mhines-usgs:
will keep looking tomorrow, have to go for now and didn't get through it all.

inst/uvhydrograph/uvhydrograph.Rmd
```diff
@@ -38,13 +38,12 @@ for (i in seq_len(length(plots.tables.ALL))) {
...
   if(!is.null(x$plot1)){
     noData_vec <- c(noData_vec, FALSE)
-    par(mar=c(8, 3, 4, 3))
-    printReportFeature(x$plot1)
-    printReportFeature(x$table1, isTable=TRUE)
-    printReportFeature(x$status_msg1, m=months_nm[i])
-    printReportFeature(x$plot2)
-    printReportFeature(x$table2, isTable=TRUE)
-    printReportFeature(x$status_msg2, m=months_nm[i])
+    printReportFeature(x$plot1, mar_values=c(8, 3, 4, 3))
```
mhines-usgs: not sure it's possible maybe because you don't want them to be null coming into the function, but if the mar_values are all the same here, is there a reason we can't just default them to that somewhere in printReportFeature without passing them in?

PR #420, opened 2016-09-30
adding in another approval type

mhines-usgs:
This isn't a top priority, just came out of the task accidentally including a different approval type called admin read-only. This trumps all other approval types. But made me wonder how extensible adding an approval type is to our current scheme. Just fiddled around for a little bit with this and checking it in here as a thing to look at sometime, but it's not urgent to resolve. Maybe as we refactor it'll be useful to consider.

not entirely sure how to include properly in the index part of utils-json, nor the extendToWholeDays

I think these two spots assume Working is God, instead of Admin Read-Only. If that's true, I can make a better guess, but should confirm first.

This is just kind of a fun add-on, since Laura accidentally included this type in the ticket initially, but also gives us a good test to ask, how do we add another approval type, when the time comes?

### DIFF
--- a/R/utils-json.R
+++ b/R/utils-json.R
@@ -218,12 +218,14 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
     working_index <- getApprovalIndex(data, points, chain_nm, "Working");
     review_index <- getApprovalIndex(data, points, chain_nm, "In Review");
     approved_index <- getApprovalIndex(data, points, chain_nm, "Approved");
+    admin_index <- getApprovalIndex(data, points, chain_nm, "Admin Read-Only");
     
     review_index <- setdiff(review_index, working_index)
     approved_index <- setdiff(approved_index, working_index)
     approved_index <- setdiff(approved_index, review_index)
+    approved_index <- setdiff(approved_index, admin_index)
 
-    date_index_list <- list(list(type="Approved",approved_index), list(type="In Review",review_index), list(type="Working",working_index))
+    date_index_list <- list(list(type="Approved",approved_index), list(type="In Review",review_index), list(type="Working",working_index), list(type="Admin Read-Only", admin_index))
 
     for(sub_list in date_index_list){
       approval_info <- list()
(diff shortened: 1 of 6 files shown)


R/utils-json.R
```diff
@@ -218,12 +218,14 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
     working_index <- getApprovalIndex(data, points, chain_nm, "Working");
     review_index <- getApprovalIndex(data, points, chain_nm, "In Review");
     approved_index <- getApprovalIndex(data, points, chain_nm, "Approved");
+    admin_index <- getApprovalIndex(data, points, chain_nm, "Admin Read-Only");
     
     review_index <- setdiff(review_index, working_index)
     approved_index <- setdiff(approved_index, working_index)
     approved_index <- setdiff(approved_index, review_index)
+    approved_index <- setdiff(approved_index, admin_index)
```
mhines-usgs, line 226: fuzzy on this and if it's the admin read-only are the top dog of approvals, does it go above working instead ? need some clarity on what's going on here.

also, not entirely sure how the extendToWholeDays needs to be updated if admin read-only trump all others.

zmoore-usgs: The diffs work by removing any values from the first vector that also exist in the second vector. The logic here is to make the dv approval points show for the lowest approval status on any given day.

The three indices are initially filled with each day that this approval status is present on, and then in order to make the lowest approval status be the one that shows up for any given day we remove duplicates from the higher level status indices, leaving those days in the lower status indices. 

So in this case we:
1. Remove any points from "In Review" that are also in "Working" because "Working" is the lower status. 
2. Remove any points from "Approved" that are also in "In Review" because "In Review" is the lower status.
3. Remove any points from "Approved" that are also in "Working" because "Working" is the lower status.

Both "Working" and "In Reivew" have to be removed from "Approved" because there is not a guarantee that an "In Review" period exists between a "Working" period and an "Approved" period.

So in the case of adding this new status that is the top dog we would need all of the following because we want DV points to show the lowest approval status on any given day:

admin_index <- setdiff(admin_index, working_index)
admin_index <- setDiff(admin_index, review_index)
admin_index <- setDiff(admin_index, approved_index)

mhines-usgs:
Closing this without merging, we'll have it in case we ever need to actually add a new approval type/rating/whatevs.

PR #423, opened 2016-10-03
AQCU-827: Fixed a small typo that was casuing the Uncorrected TS to occasionally go off the chart.

zmoore-usgs:


mhines-usgs:
if all tests pass, i agree, go ahead and merge!

PR #425, opened 2016-10-03
AQCU-844: Temporary(?) fix for boxes rendering improperly with year 9999

zmoore-usgs:
Issue seems to be something within GSPLOT?

### DIFF
--- a/R/correctionsataglance-render.R
+++ b/R/correctionsataglance-render.R
@@ -102,7 +102,7 @@ plotLanes <- function(gsplotObject, laneData, whichCol,
 
     #Fix for Year 9999 not rendering properly
     if(laneData$endDates > addData$dateData$dateRange[2]){
-      laneData$endDates <- addData$dateData$dateRange[2] + days(1)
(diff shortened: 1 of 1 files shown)


mhines-usgs:
We came across this doing the corr report as well as in transitioning to the bars/rectangles for approvals. To "fix" we change the year if it was 9999 to 2100.

zmoore-usgs:
@mhines-usgs Ah gotcha. For consistency purposes do you think it would be best to change my fix from changing "9999-xx-xx" to the plot max date + 1 day to just changing it to be year 2100?

mhines-usgs:
I think your fix works just fine, but you might want to make a comment above it why it's there? so we can revisit it...

mhines-usgs:
duh, you already did that ;)

lindsayplatt:
We're trying to get away from the hard-coded date changes. I would suggest switching to the 2100 method. Use the function `toEndOfTime` for that https://github.com/USGS-R/repgen/blob/master/R/utils-time.R#L86

``` r
laneData$endDates <- toEndOfTime(addData$dateData$dateRange[2])
```

PR #426, opened 2016-10-03
AQCU-843: Made appearance of * and ** in report header dynamic based …

zmoore-usgs:
…on if it is included in the report or not.

### DIFF
--- a/R/sitevisitpeak-data.R
+++ b/R/sitevisitpeak-data.R
@@ -141,12 +141,12 @@ getIvDifference <- function(readingVal, ivVal) {
 }
 
 containsOutsideUncertainty <- function(data) {
-  diff <- list()
-  data <- data$readings
-  for(listRows in row.names(data)){
-    listElements <- data[listRows,]
-    diff <- append(diff, getIvDifference(listElements$value, listElements$associatedIvValue))
+  readings_diff <- list()
+  readings_data <- data$readings
+  for(listRows in row.names(readings_data)){
+    listElements <- readings_data[listRows,]
+    readings_diff <- append(readings_diff, getIvDifference(listElements$value, listElements$associatedIvValue))
   }
 
-  return(length(diff[grepl("\\*\\*", diff)]) > 0)
+  return(length(readings_diff[grepl("\\*\\*", readings_diff)]) > 0)
 }
--- a/R/sitevisitpeak-data.R
+++ b/R/sitevisitpeak-data.R
@@ -133,9 +133,20 @@ getIvDifference <- function(readingVal, ivVal) {
       result <- as.character(round(val, digits = nchar(ivVal)))
       
       if(abs(val) > 0.05) {
-        result <- paste(result, "*")
+        result <- paste(result, "**")
       }
     }
   }
   return(result)
+}
+
+containsOutsideUncertainty <- function(data) {
+  diff <- list()
+  data <- data$readings
+  for(listRows in row.names(data)){
+    listElements <- data[listRows,]
+    diff <- append(diff, getIvDifference(listElements$value, listElements$associatedIvValue))
+  }
+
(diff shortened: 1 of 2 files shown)


R/sitevisitpeak-data.R
```diff
@@ -133,9 +133,20 @@ getIvDifference <- function(readingVal, ivVal) {
...
+containsOutsideUncertainty <- function(data) {
+  diff <- list()
+  data <- data$readings
+  for(listRows in row.names(data)){
+    listElements <- data[listRows,]
+    diff <- append(diff, getIvDifference(listElements$value, listElements$associatedIvValue))
+  }
+
+  return(length(diff[grepl("\\*\\*", diff)]) > 0)
+}
```
zmoore-usgs, line 152: I'm really not sure if this is the most efficient way of figuring out if any of the values are outside of the specified range before actually rendering the table, so I'm certainly open to suggestions and feedback on this entire function.

R/sitevisitpeak-data.R
```diff
@@ -133,9 +133,20 @@ getIvDifference <- function(readingVal, ivVal) {
...
+        result <- paste(result, "**")
       }
     }
   }
   return(result)
 }
+
+containsOutsideUncertainty <- function(data) {
+  diff <- list()
+  data <- data$readings
```
lindsayplatt: probably name this `readings_data` or something to not just overwrite data -- or pass in `data$readings` to begin with

R/sitevisitpeak-data.R
zmoore-usgs: Code that is nearly identical to this already exists in one of the main rendering functions for this plot (formatSVPData) but I couldn't come up with a cleaner way of reusing it than this, so any suggestions on that would be welcome as well.

lindsayplatt: rather than a for loop -- we might be able to use apply? 

Possibly following something like this, where you take the body of the for loop and make a function. 

This:

``` r
mydf <- data.frame(firstVal = c(5, 12, 8), secondVal = c(3, 9, 1.5))
toRet <- data.frame(stringsAsFactors = FALSE)
for(listRows in row.names(mydf)){
  listElements <- mydf[listRows,]

  firstVal <- listElements$firstVal
  secondVal <- listElements$secondVal
  thirdVal <- sum(firstVal, secondVal)
  toAdd <- c(firstVal, secondVal, thirdVal)
  toRet <- rbind(toRet, data.frame(t(toAdd),stringsAsFactors = FALSE))
}
```

 becomes this:

``` r
myFunc <- function(df){
    firstVal <- df['firstVal']
    secondVal <- df['secondVal']
    thirdVal <- sum(firstVal, secondVal)
    toAdd <- c(firstVal, secondVal, thirdVal)
    toAddDf <- data.frame(t(toAdd),stringsAsFactors = FALSE)
    return(toAddDf)
}
mydf <- data.frame(firstVal = c(5, 12, 8), secondVal = c(3, 9, 1.5))
toRet_list <- apply(mydf, 1, myFunc)
toRet <- do.call(rbind, toRet_list)
```

R/sitevisitpeak-data.R
```diff
@@ -133,9 +133,20 @@ getIvDifference <- function(readingVal, ivVal) {
...
+  for(listRows in row.names(data)){
+    listElements <- data[listRows,]
+    diff <- append(diff, getIvDifference(listElements$value, listElements$associatedIvValue))
```
lindsayplatt: don't use `diff` as a variable name - it's an existing function

lindsayplatt (changes requested):
put some ideas about how to change this, but the larger suggested change using apply + function rather than a loop is probably out of scope.

lindsayplatt (approved):

PR #427, opened 2016-10-03
drop this because we do have timezone in all data$reportMetadata$time…

mhines-usgs:
…zone now


lindsayplatt:
HA - thanks for catching that! I completely forgot to remove that

mhines-usgs:
no prob, was looking at the timezone on dvhydro and searching for "timezone" to see all instances of its use in dvhydro, and this just happened to come up on top of the find in files list in R Studio :D

PR #430, opened 2016-10-06
AQCU-808: Added functionality for estimated series to plot with gaps.

zmoore-usgs:


mhines-usgs:
pulling this locally to test out.

mhines-usgs:
@zmoore-usgs what's a good example of testing here? Not seeing much of a difference in before/after with the one suggested by laura?  "the Location 01010000, Discharge, Jan to Aug 2015 for DV Hydro has estimates that should show, if that's useful."

zmoore-usgs:
@mhines-usgs Yeah, that additional one from Laura didn't have multiple estimated periods so there wasn't much difference. Take a look at the third example she posted in the description:
"DV Hydro, Location 01643700, Discharge, Dec 2013 to Feb 2014"

mhines-usgs:
thanks :)

PR #432, opened 2016-10-11
DO NOT MERGE -- Pure R implementation of library prerequisites installation.

ashalper-usgs:


inst/extdata/installRDependencies.sh
```diff
@@ -1,26 +0,0 @@
-#!/bin/bash
-Rscript -e 'remove.packages("curl")' || true
```
thongsav-usgs, line 2: We still need the sh script, it is hook into jenkins.

thongsav-usgs: I can figure out the conversion, but should not remove this file yet.

thongsav-usgs:
Do not merge until the sh file is restored.

inst/extdata/installRDependencies.sh
ashalper-usgs: Jenkins doesn't have a way to run R directly? Seems like if it can run `bash`, it should be able to run `Rscript`. Doesn't matter, just surprised.

ashalper-usgs:
OK, I restored the shell script.

mhines-usgs:
What else has to happen with this PR @thongsav-usgs ? Can we merge now that Andy restored the original sh file?

PR #433, opened 2016-10-17
Corrected bug in which certain approval bars were failing to plot.

ashalper-usgs:


lindsayplatt:
looks good to me :+1:

PR #434, opened 2016-10-20
Corrected top margin of DV hydrograph & five year GW summary plots.

ashalper-usgs:


R/utils-shared.R
```diff
@@ -534,25 +534,37 @@ RescaleYTop <- function(object) {
...
-      object$side.2$lim[1] <- (1 - m) * object$side.2$lim[1]
+      object$side.2$lim[1] <- e[2] - m * e.length
     }
     else {
-      object$side.2$lim[2] <- (1 + m) * object$side.2$lim[2]
+      # The 5.14 coefficient below is a hack that was reverse-engineered from 
+      # the SVG output (using Inkscape, and the "back-of-the-envelope"). At the 
+      # moment, we have no idea why it works, and is likely not robust enough 
+      # for production.
+      object$side.2$lim[2] <- e[2] + 5.14 * m * e.length
```
lindsayplatt, line 564: @ashalper-usgs is this the part that is necessary for DV hydro but not Five yr?

ashalper-usgs: I think I might've been incorrect with that earlier assessment. I thought it was at the time; however, today I think it was affecting them both, but only when the _y_-axis was not inverted (why, I still don't know). I think earlier I assumed it was a DV-hydrograph-only problem, because the test data for DV hydrographs had mostly not-inverted _y_ axes, and the five year GW summary had an inverted _y_-axis.

lindsayplatt: hmm OK. For now I think we are good to merge. I added it to the tech debt doc

PR #440, opened 2016-10-26
Increased scipen to 8 for all reports. Also had to update the seq

thongsav-usgs:
calculation for vdiagram so that spans less than 1 would auto calculate.

### DIFF
--- a/R/vdiagram-render.R
+++ b/R/vdiagram-render.R
@@ -47,6 +47,8 @@ renderVDiagram <- function(data){
 }
 
 createVdiagram <- function(data) {
+  options(scipen=8)
+  
   styles <- getVDiagramStyle()
   
   vdiagramData <- parseVDiagramData(data)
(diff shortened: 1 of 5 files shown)


R/vdiagram-render.R
```diff
@@ -81,19 +83,47 @@ createVdiagram <- function(data) {
...
+  vplot <- do.call(axis, list(object=vplot,side=c(1,2,4), at=c(x_seq, y_seq,y_seq)))
   
   print(vplot)
 }
 
+#This function produces a sequence if the axis does not span more than 1 whole integer
+getAxisSeq <-function(lims) {
+  lowerBound <- lims[1]
+  upperBound <- lims[2]
+  seq <- seq(lowerBound, upperBound)
```
lindsayplatt: use different name - this is the same as the function

R/vdiagram-render.R
```diff
@@ -81,19 +83,47 @@ createVdiagram <- function(data) {
...
+        newLowerBound <- newLowerBound - stepSize
+      }
+      seq <- seq(newLowerBound, newUpperBound, stepSize)
+    } else {
+      stepSize <- (abs(upperBound) - abs(lowerBound)) / 2 
+      seq <- seq(lowerBound, upperBound, stepSize)
+    }
+  }
+  
+  seq <- pretty(seq,n=length(seq),shrink.sml = 20)
```
lindsayplatt: I don't really understand why anything in the `if(length(seq) == 1)` needs to be there. What's wrong with just sending `seq(lowerBound, upperBound)` to `pretty`? It will work for small and negative values:

``` r
lims <- c(0.0001, 0.01)
pretty(lims)
[1] 0.000 0.002 0.004 0.006 0.008 0.010

lims <- c(-0.0001, 0.01)
pretty(lims)
[1] -0.002  0.000  0.002  0.004  0.006  0.008  0.010
```

lindsayplatt: not sure why it shows this as outdated... @thongsav-usgs this comment exists :)

thongsav-usgs: I don't think it matters for the original issue of scientific notation. That was actually fixed by doing the explicit axis calls and passing in the calculated sequence, however the issue being addressed with the if block is is with the original seq call, 

```
test <- seq(0.2, 0.5)
```

Test is only 1 point there and we do not get multiple tick marks on the x-axis.

thongsav-usgs: Doing more testing, you're right, it may not matter. I had wrote all the seq code before I figured out I had to do an explicit axis call. Give me a few minutes.

thongsav-usgs: It does matter somewhat if we want more tick marks on the x-axis. If the original seq only has 1 point you only get the 1 tick so I'm going to leave as is.

lindsayplatt: Right, but if you explicitly use pretty instead of seq, you should get the right ticks. Maybe I'm not following, but the example below illustrates what I mean.

E.g.

``` r
ylims <- c(0, 0.1)
y_seq <- seq(ylims[1], ylims[2]) #gives only 1 value - zero
gs <- gsplot() %>% 
    points(1:3, c(0.03, 0.05, 0.07), ylim=ylims) %>% 
    axis(side=2, at=y_seq)
gs

y_seq <- pretty(ylims) #gives correct sequence to use in axis()
gs <- gsplot() %>% 
    points(1:3, c(0.03, 0.05, 0.07), ylim=ylims) %>% 
    axis(side=2, at=y_seq)
gs
```

thongsav-usgs: You were right, I was goofing still using the "n" parameter. I cut everything out and we are relying on just the pretty function now.

PR #443, opened 2016-10-27
Five year GW summary plot style labeling of x-axis transcribed to DV hydrograph plot

ashalper-usgs:
Also made five year GW summary plot window same size as DV hydrograph plot window for consistency.


R/dvhydrograph-render.R
```diff
@@ -120,3 +127,61 @@ createRefPlot <- function(data, series) {
...
   }
 }
+
+XAxisLabelStyle <- function(object, start, end, timezone) {
+  # if chart interval is less than 1 year
+  i <- interval(start, end, tzone = attr(start, timezone))
+  if (as.period(i) < years(1)) {
+    # omit extended x-axis labeling
+    dates <- seq(start, end, by = 7 * 24 * 60 * 60)
+    dates <- toStartOfDay(dates)
```
lindsayplatt: you should be able to pass in `plotDates` from `dvhydrograph-render.R` as an argument instead of recreating it

R/dvhydrograph-render.R
```diff
@@ -120,3 +127,61 @@ createRefPlot <- function(data, series) {
...
+      1, at = dates,
+      labels = format(dates, "%b\n%d"),
+      padj = 0.5
+    )
+  }
+  else {
+    # if DV interval is not aligned on the 1st of a month
+    if (day(start) != 1) {
+      # we'll start labeling months at the following whole month
+      from <- floor_date(start %m+% months(1), "month")
```
lindsayplatt, line 159: never seen that notation before, cool :+1:

R/utils-shared.R
```diff
@@ -568,6 +570,41 @@ RescaleYTop <- function(object) {
...
+#' @return The passed-in gsplot object, with x-axis labeled.
+XAxisLabels <- function(object, month_label, month_label_location, date_seq_yr) {
+  return(
+    mtext(
+      object,
+      text = month_label, at = month_label_location,
+      cex = 0.5, side = 1
+    ) %>%
+      mtext(
+        text = year(date_seq_yr), at = date_seq_yr + (60 * 60 * 24 * 30 * 6),
```
lindsayplatt: this always places the year label around June, but that doesn't mean the year label will show up. For instance, the `dvhydro-example.json` example has 2014-11-20 to 2015-11-20, but you can only see the 2015 label since they are placed in June. Need a more dynamic way to place the labels - like the midpoint of the available data for each year.

E.g.

``` r
# do this inside of XAxisLabels; you'd need to pass in plotDates
yr.at <- sapply(year(date_seq_yr), function(y, plotDates){
  which.yr.dates <- which(year(plotDates) == y)
  med.date.each.year <- median(plotDates[which.yr.dates])
  return(med.date.each.year)
}, plotDates = plotDates)
```

lindsayplatt (changes requested):

R/utils-shared.R
lindsayplatt: @ashalper-usgs do you intend to change this or are you ready for a merge?

ashalper-usgs: Would you know how to return `med.date.each.year` as a date? Right now this seems to be returning a vector of `num`, so I think `mtext()` is failing to label the _x_-axis.

lindsayplatt: Whoops, yeah. When it "unlists" it does that conversion, so you need to use do.call and apply the function `c` to the output of `lapply`, not `sapply`:

``` r
yr.at <- do.call(c, lapply(year(date_seq_yr), function(y, plotDates){
    which.yr.dates <- which(year(plotDates) == y)
    med.date.each.year <- median(plotDates[which.yr.dates])
    return(med.date.each.year)
}, plotDates = plotDates))
```

If you want to know why, this stack overflow post explains it http://stackoverflow.com/questions/14449166/dates-with-lapply-and-sapply

ashalper-usgs: Missing cast arising from a cast?

```
 Error in do.call(c, sapply(year(years), function(y, plotDates) { : 
  second argument must be a list 
```

lindsayplatt: You've got to use `lapply` not `sapply` since the latter returns a vector and not a list.

lindsayplatt (approved):

ashalper-usgs:
This shouldn't be merged yet. Still some bugs that need to be fixed on the DV hydrographs.

PR #444, opened 2016-10-28
AQCU-858: Only show one max/min per day

zmoore-usgs:


R/extremes-data.R
```diff
@@ -135,6 +135,31 @@ createDataRows <- function(data, param, rowName, isUpchain, includeRelated=TRUE,
         dataRows <- dataRows[!duplicated(dataRows[c("date", "primary")]),]
       }
 
+      #Keep only the row with the min/max corresponding value for a given day
+      if(includeRelated && doMerge){
+        if(!isUpchain){
+          #Keep Maximum or Minimum based on current param
+          if(param == "max"){
+            duplicateRows <- dataRows[with(dataRows, order(dataRows$date, dataRows$related, decreasing = TRUE)),]
```
lindsayplatt: Using the function `with` means you don't need to explicitly say that the `date` column is from dataRows. So you can delete the `with` part or delete the `dataRows$` part. It's just redundant right now.

`duplicateRows <- dataRows[with(dataRows, order(date, related, decreasing = TRUE)),]`

or 

`duplicateRows <- dataRows[order(dataRows$date, dataRows$related, decreasing = TRUE),]`

R/extremes-data.R
```diff
@@ -135,6 +135,31 @@ createDataRows <- function(data, param, rowName, isUpchain, includeRelated=TRUE,
         dataRows <- dataRows[!duplicated(dataRows[c("date", "primary")]),]
       }
 
+      #Keep only the row with the min/max corresponding value for a given day
+      if(includeRelated && doMerge){
+        if(!isUpchain){
```
lindsayplatt: is there any way to combine the stuff in these two if statements with the ones on L132? Seems very similar, not sure if you've looked into that yet and it just didn't work.

R/extremes-data.R
```diff
@@ -135,6 +135,31 @@ createDataRows <- function(data, param, rowName, isUpchain, includeRelated=TRUE,
...
+          if(param == "max"){
+            duplicateRows <- dataRows[with(dataRows, order(dataRows$date, dataRows$related, decreasing = TRUE)),]
+          } else {
+            duplicateRows <- dataRows[with(dataRows, order(dataRows$date, dataRows$related, decreasing = FALSE)),]
+          }
+        } else {
+          #Keep Maximum or Minimum based on current param
+          if(param == "max"){
+            duplicateRows <- dataRows[with(dataRows, order(dataRows$date, dataRows$primary, decreasing = TRUE)),]
+          } else {
```
lindsayplatt: be explicit and say `else if(param == "min"){`

lindsayplatt:
You should add a test to make sure this is working as expected

lindsayplatt (approved):
possibly change the added test so that an entire json file isn't added, then merge!

PR #445, opened 2016-10-28
AQCU-880: UV Hydro sometimes fails to render

zmoore-usgs:


R/utils-json.R
```diff
@@ -264,7 +264,7 @@ getApprovals <- function(data, chain_nm, legend_nm, appr_var_all, month=NULL, po
     appr_dates <- NULL
     chain <- data[[chain_nm]]
     
-    if (!isEmpty(chain$approvals$startTime)) {
+    if (!isEmpty(chain$approvals$startTime) && !is.null(chain$startTime)) {
```
lindsayplatt: does the function `isEmptyOrBlank` work instead of the && statement? https://github.com/USGS-R/repgen/blob/master/R/utils-shared.R#L242

zmoore-usgs: I did give that a shot but chain$approvals$startTime can sometimes still exist even if chain$startTime does not, so both have to be checked. I think using isEmptyOrBlank to check both variables is probably the safest method though?

lindsayplatt: oh sorry  - did not notice that the reference inside of is.null was different than what was in isEmpty. But yes, isEmptyOrBlank is probably the safest

lindsayplatt (approved):

PR #446, opened 2016-10-31
AQCU-868: Slight change to label timezone formatting based on…

zmoore-usgs:
…feedback from Laura.

### DIFF
--- a/R/fiveyeargwsum-data.R
+++ b/R/fiveyeargwsum-data.R
@@ -11,8 +11,8 @@ parseFiveYrData <- function(data){
     max_iv <- getMaxMinIv(data, 'MAX')
     min_iv <- getMaxMinIv(data, 'MIN')
   } else if(!is.null(data[['reportMetadata']][['excludeMinMax']]) && data[['reportMetadata']][['excludeMinMax']] == TRUE){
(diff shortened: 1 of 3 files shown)


R/fiveyeargwsum-data.R
```diff
@@ -11,8 +11,8 @@ parseFiveYrData <- function(data){
     max_iv <- getMaxMinIv(data, 'MAX')
     min_iv <- getMaxMinIv(data, 'MIN')
   } else if(!is.null(data[['reportMetadata']][['excludeMinMax']]) && data[['reportMetadata']][['excludeMinMax']] == TRUE){
-    max_iv_label <- getMaxMinIv(data, 'MAX')
-    min_iv_label <- getMaxMinIv(data, 'MIN')
+    max_iv_label <- getMaxMinIv_fiveyr(data, 'MAX')
+    min_iv_label <- getMaxMinIv_fiveyr(data, 'MIN')
```
zmoore-usgs, line 15: These were originally using the getMaxMinIv_fiveyr function but when I was doing work on this ticket I accidentally switched them to using the DVHydro getMaxMinIv function. I don't believe the functions differ at all but I changed it back just to keep consistency with what was there before.

lindsayplatt (approved):

PR #447, opened 2016-10-31
improved logic for avoiding errors when data are missing for gage height in vdiagram

mhines-usgs:
more for AQCU-859


tests/testthat/test-vdiagram.R
```diff
@@ -19,4 +19,20 @@ test_that("vdiagram examples work", {
  expect_is(vdiagram(data, 'html'), 'character')
 })
 
+context("testing vdiagram when gage heights are blank")
+test_that("vdiagram examples work", {
```
lindsayplatt, line 23: I think we are trying to add less JSON files to pinpoint what part of the JSON is being tested. See `test-utils-json.R`. I don't think it's necessary to change this one, but just so you know in the future!

lindsayplatt (approved):

tests/testthat/test-vdiagram.R
mhines-usgs: ok, I guess that makes sense, but how should we catch the case where like this task, we're not handling missing data with a specific report and it fails to render?

lindsayplatt: hmmm good point :thinking:

mhines-usgs: I put it into the tech debt list, it's probably worth combing through all the reports and catching these issues before we run into them testing. Thanks!

PR #448, opened 2016-11-01
one more check for empty values before plotting on vdiagram AQCU-859

mhines-usgs:


lindsayplatt:
:+1: guess I didn't peer review as thoroughly as I should have!

mhines-usgs:
That was my fault, I knew this was causing issues but I didn't aggressively add the isEmptyOrNull checks.

PR #449, opened 2016-11-01
Explicitly set seq on axis for effective shifts.

thongsav-usgs:


lindsayplatt:
@thongsav-usgs I think the build failed because of a network issue, not your code. Can you try kicking it off again? I'll merge once checks pass.

PR #450, opened 2016-11-01
Fix 865

lindsayplatt:
Only implemented for uvhydro so far. 

@thongsav-usgs @mhines-usgs how's this looking?


R/utils-json.R
```diff
@@ -53,15 +53,32 @@ getMinStage <- function(ts, ...){
 }
 
 #finds if the plot data has any zero values
-zeroValues <- function(dataList, val_nm){    
-  logList <- lapply(dataList, function(x) {any(na.omit(x[[val_nm]]) == 0)})
-  logVector <- any(unlist(unname(logList)))
+zeroValues <- function(data, val_nm){
```
thongsav-usgs, line 61: can this just take in the timeseries instead of data and val_nm? Have the caller pass data[[val_nm]]?

R/utils-json.R
```diff
@@ -53,15 +53,32 @@ getMinStage <- function(ts, ...){
...
+    zeroData <- any(data[[val_nm]] == 0)
+  }
+  return(zeroData)
 }
 
 #finds if the plot data has any zero values
-negValues <- function(dataList, val_nm){    
-  logList <- lapply(dataList, function(x) {any(na.omit(x[[val_nm]]) < 0)})
-  logVector <- any(unlist(unname(logList)))
+negValues <- function(data, val_nm){
```
thongsav-usgs, line 73: Same question

R/uvhydrograph-data.R
```diff
@@ -164,8 +166,8 @@ parseUVSupplemental <- function(data, plotName, pts, useDownsampled=FALSE) {
     comp_UV_type <- data[['comparisonSeries']]$type
     comp_UV_TS_lbl <- getTimeSeriesLabel(data, "comparisonSeries");
     dates <- seq(lims_UV$xlim[1], lims_UV$xlim[2], by="days")
-    
-    logAxis <- isLogged(data, pts, "firstDownChain")
+    browser()
```
mhines-usgs: did you mean to check this in?

R/utils-json.R
lindsayplatt: unfortunately not because data could be a list, which is why there is an apply function call. Then for each list element it needs to use val_nm

R/utils-json.R
lindsayplatt: same answer :wink:

R/uvhydrograph-data.R
lindsayplatt: whoops absolutely not!

lindsayplatt: I actually did not mean to check in any of those changes.

PR #451, opened 2016-11-03
loose ends from AQCU-865 (exclude <= 0 flows)

lindsayplatt:
logic for determining if there will be a logging error or not was not working properly. It would log even when there were zeros or negatives but no excludeZeroNegative flag was indicated (empty rather than T or F).


R/uvhydrograph-render.R
```diff
@@ -109,11 +109,24 @@ createPrimaryPlot <- function(data, month, useDownsampled=FALSE){
...
+    # could be added using logic in if statement above
+    # if(isLogged(data, ylimPrimaryData, primaryInfo$primarySeriesName)){
+    #   plot_object <- view(plot_object, side=primarySide, log='y')
+    # }
+    # if(comparisonExist && isLogged(data, ylimComparisonData, primaryInfo$comparisonSeriesName)){
+    #   plot_object <- view(plot_object, side=comparisonSide, log='y')
+    # }
+    # if(referenceExist && isLogged(data, ylimReferenceData, primaryInfo$referenceSeriesName)){
+    #   plot_object <- view(plot_object, side=referenceSide, log='y')
+    # }
```
lindsayplatt, line 127: this is commented out because if one vertical side is logged, it logs all of them. This will hopefully be fixed in [gsplot #414](https://github.com/USGS-R/gsplot/issues/414)

PR #452, opened 2016-11-03
catching the case where there are no measurements, so histFlag is emp…

mhines-usgs:
…ty and no length. Setting histFlag to false so it can continue through the measurement function without erroring.

AKA AQCU-883

### DIFF
--- a/R/vdiagram-data.R
+++ b/R/vdiagram-data.R
@@ -42,7 +42,9 @@ defaultHistFlags <- function(histFlag){
   if (length(histFlag)==1 && histFlag == " "){
     histFlag <- rep(TRUE, length(x))
   }
-  
+  if (isEmptyOrBlank(histFlag)){
+    histFlag <- FALSE
+  }
   return(histFlag);
 }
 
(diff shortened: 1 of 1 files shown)


R/vdiagram-data.R
```diff
@@ -42,7 +42,9 @@ defaultHistFlags <- function(histFlag){
   if (length(histFlag)==1 && histFlag == " "){
     histFlag <- rep(TRUE, length(x))
   }
-  
+  if (histFlag==""){
```
lindsayplatt: I think isEmptyOrBlank will do this

mhines-usgs: k, i'll try that

mhines-usgs: yup :D :+1:

lindsayplatt: just trying to take advantage of all those functions we made :)

mhines-usgs: agreed, thanks for being vigilant :sheriffs-badge:

PR #453, opened 2016-11-05
Corrected y-axis interval on V diagrams.

ashalper-usgs:


lindsayplatt (approved):
Easy fix- nice! Could you add a comment that says those to if statements need to come after all data is added? I want to make sure we do that in the future if we add more data.

lindsayplatt:
Also, make sure you build and reload the package (go to the Build tab in RStudio and click Build & Reload). This compiles changes to function documentation and the NAMESPACE.

ashalper-usgs:
Is it true that the two conditionals need to execute after all plot objects are added? I was wildly speculating when I wrote that. Don't know gsplot or R graphics well enough to know.

ashalper-usgs:
I have been running `Build & Reload` after every edit.

mhines-usgs:
@ashalper-usgs in order to have the build and reload rebuild documentation, there is a setting in the RStudio options to check. Otherwise you might have to force it with roxygen2 & roxygenize()

lindsayplatt:
I thought the order change was how you actually fixed the issue. You first add all of the plot data so the limits are set based on those points. Then, you do a final check to make sure the plot limits include the max and min stage lines. If not, you change them. If they do include those points, then you leave it. 

Although, I guess the fix might have worked because you deleted the empty axis call.

ashalper-usgs:
@mhines-usgs OK, so you're actually checking in derived files (such as roxygen outputs) as source, instead of only building them at install? This is fine, and I'm sure you have your reasons; just not accustomed to working this way.

lindsayplatt:
@ashalper-usgs 
1. RStudio build tab
2. More dropdown
3. Configure Build Tools
4. Check "Generate documentation with Roxygen"
5. Now build and reload

ashalper-usgs:
@lindsaycarr I really have little to no idea how I corrected the bug. Mostly coding by instinct here.

PR #454, opened 2016-11-09
Modified order that time series points are layered in UV hydrographs

ashalper-usgs:
Corrected points now (always)  appear on top of uncorrected points.

PR #457, opened 2016-11-15
Corrected bug in plotting V diagrams' y-axis interval

ashalper-usgs:


R/vdiagram-render.R
```diff
@@ -67,7 +67,10 @@ createVdiagram <- function(data) {
 
   vplot <- testCallouts(vplot, xlimits = xlim(vplot)$side.1)
   
-  ylims <- ylim(vplot)$side.2
+  ylims <- c(
+      min(c(ylim(vplot)$side.2, vdiagramData$minStage)),
+      max(c(ylim(vplot)$side.2, vdiagramData$maxStage))
+      )
```
lindsayplatt, line 73: I think you've got a combo fix, where there should be one or another. It doesn't make a difference functionally, but we don't need to duplicate setting the ylims based on max or min stage. 

Just move all of the grid line + axis stuff below changing the limits for maxStage/minStage. So:

``` r
  vplot <- testCallouts(vplot, xlimits = xlim(vplot)$side.1)

  if (vdiagramData$minStage < ylim(vplot)$side.2[1]) {
    vplot$side.2$lim[1] <- vdiagramData$minStage
  }

  if (ylim(vplot)$side.2[2] < vdiagramData$maxStage) {
    vplot$side.2$lim[2] <- vdiagramData$maxStage
  }

  ylims <- ylim(vplot)$side.2
  xlims <- xlim(vplot)$side.1
  y_seq <- pretty(ylims, shrink.sml = 20)
  x_seq <- pretty(xlims, shrink.sml = 20)

  vplot <-
    do.call(abline, append(list(object = vplot, h = y_seq), styles$ablines))
  vplot <-
    do.call(abline, append(list(object = vplot, v = x_seq), styles$ablines))
  vplot <-
    do.call(axis, list(
      object = vplot, side = c(1, 2, 4), at = c(x_seq, y_seq, y_seq)
    ))

  print(vplot)
```

OR do what you've done to change ylims, and then remove the logic for changing limits if max and min stage are not included (code chunk below). I personally like this option - it's cleaner to remove all that logic, and keep your ylim solution using min() and max() of current limits + min/max stage.

``` r
  if (vdiagramData$minStage < ylim(vplot)$side.2[1]) {
    vplot$side.2$lim[1] <- vdiagramData$minStage
  }

  if (ylim(vplot)$side.2[2] < vdiagramData$maxStage) {
    vplot$side.2$lim[2] <- vdiagramData$maxStage
  }
```

lindsayplatt (commented):
I noticed that this PR has those updated .Rd files and NAMESPACE! Just make sure you pull from upstream and commit those changes if #454 gets merged first.

R/vdiagram-render.R
ashalper-usgs: The latter solution certainly make sense "on paper", but doesn't produce the expected behavior when implemented, and I don't know what at the moment.

lindsayplatt: whoops - that would be since we still aren't changing the limits. You should be able to just add `vplot <- view(vplot, side=2, ylim=ylims)`

lindsayplatt: also, while you're changing this stuff, can you change the axis call to be just an axis call rather than inside a do.call. `do.call` is only needed when you have the list of styles to add or in a loop.

This should work just fine:

``` r
vplot <- axis(vplot, side = c(1, 2, 4), at = c(x_seq, y_seq, y_seq)) %>% 
    view(side=2, ylim=ylims)
```

PR #458, opened 2016-11-16
Corrected position of approval bars under certain plotting circumstances

ashalper-usgs:
Corrected position and height of approval bars when and only when, all plotted time series are a horizontal line, parallel to the _x_-axis.


R/utils-shared.R
```diff
@@ -448,13 +448,31 @@ getWaterDataUrl <- function(data) {
...
 #' @return gsplot object with approval bar rectangle styles applied.
 ApplyApprovalBarStyles <- function(object, data) {
+  ylim <- ylim(object)$side.2
+  ylog <- object$global$par$ylog
+  
+  if (is.null(ylog)) {
+    # presume the semantics of NULL as FALSE, which may or not be correct, but 
+    # prevents the code from terminating here
+    ylog <- FALSE
+  }
```
lindsayplatt: remove this if structure and include a call to isEmptyOrBlank in L466 (see comment below)

R/utils-shared.R
```diff
@@ -448,13 +448,31 @@ getWaterDataUrl <- function(data) {
...
+
+  if (ylim[1] == ylim[2]) {
+    # Cope with the rare case of the time series plot being a horizontal line,
+    # in which case we have to preemptively compensate for some y-axis interval
+    # defaulting code inside R graphics. The 40% factor here comes from the R
+    # source code, last seen at 
+    # http://docs.rexamine.com/R-devel/Rgraphics_8h.html#a5233f80c52d4fd86d030297ffda1445e
+    if (ylog) {
```
lindsayplatt: replace with `if(!isEmptyOrBlank(ylog) && ylog) {`

lindsayplatt (changes requested):

R/utils-shared.R
ashalper-usgs: OK, I've made these changes, but it looks like tests are failing on (unrelated code on) `master` now?

lindsayplatt (commented):
I am kind of questioning whether any of these changes are necessary - I have the master branch and the dataset mentioned in the ticket comes back with "the dataset requested is empty" as it should. I am not able to reproduce the red horizontal line.

lindsayplatt (approved):

ashalper-usgs:
@lindsaycarr, try [this report](https://cida-eros-aqcudev.er.usgs.gov:8443/aqcu-webservice/service/reports/uvhydrograph/?primaryTimeseriesIdentifier=6978204d68b34e479f33e05273090013&upchainTimeseriesIdentifier=3ec1f3f59fe944079fc0ea5a4eb03e83&referenceTimeseriesIdentifier=&firstDownChainIdentifier=5f1efbddf00848a493ff78b0bbd89f09&secondDownChainIdentifier=&thirdDownChainIdentifier=&fourthDownChainIdentifier=&ratingModelIdentifier=Gage+height-Discharge.STGQ%4001047200&secondaryRatingModelIdentifier=&station=01047200&startDate=2016-05-01&endDate=2016-08-31).

R/utils-shared.R
```diff
@@ -448,16 +448,28 @@ getWaterDataUrl <- function(data) {
...
+    # defaulting code inside R graphics. The 40% factor here comes from the R
+    # source code, last seen at 
+    # http://docs.rexamine.com/R-devel/Rgraphics_8h.html#a5233f80c52d4fd86d030297ffda1445e
+    if (!isEmptyOrBlank(ylog) && ylog) {
+      ylim <- c(10^(0.6 * log10(ylim[1])), 10^(1.4 * log10(ylim[2])))
+    }
+    else {
+      ylim <- c(0.6 * ylim[1], 1.4 * ylim[2])
+    }
+  }
```
lindsayplatt, line 466: It seems like we should have a calculation of y top and y bottom with buffer & reverse be a standalone function since it appears in `RescaleYTop`,  `ApprovalBarY`, and now here. I'm going to add that to tech debt.

R/utils-shared.R
```diff
@@ -500,12 +512,6 @@ ApprovalBarYBottom <- function(lim, ylog, reverse) {
 #' @param ratio A scaling ratio to adjust top or bottom of approval bar rectangle.
 #' @return Approval bar, top or bottom y-axis point, in world coordinates.
 ApprovalBarY <- function(lim, ylog = NULL, reverse, ratio) {
-  if (is.null(ylog)) {
-    # presume the semantics of NULL as FALSE, which may or not be correct, but 
-    # prevents the code from terminating here
-    ylog <- FALSE
-  }
-
```
lindsayplatt, line 508: since you deleted this - you need to add `isEmptyOrBlank(ylog) && ylog` to the if statement below. That's likely why tests are failing. While you're at it, you might as well add `isEmptyOrBlank(reverse) && reverse` instead of just `reverse` to both logic statements.

lindsayplatt (changes requested):
I think you just need to add those empty checks to `ApprovalBarY` and then you'll be good to go

R/utils-shared.R
ashalper-usgs: OK, I couldn't get the Boolean expression to work (looks like R doesn't do fast-fail conditional evaluation?), but I added some functionally equivalent syntax.

lindsayplatt: It does - I believe that's the difference between a single `&` and double `&&`

lindsayplatt (approved):

lindsayplatt:
build error was because of the timeout thing `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

merged anyways

PR #460, opened 2016-11-17
Added global report title heading and style

ashalper-usgs:


lindsayplatt:
is this different from #462 ??

lindsayplatt:
@ashalper-usgs I think all you need to do here is add `title(main = "UV Hydrograph")` to `createPrimaryPlot` in `uvhydrograph-render.R`.

lindsayplatt:
This added a title to each UV Hydro plot, but @zmoore-usgs added the plot titles in the report header in #460 so now this is redundant. @zmoore-usgs will be removing the titles added to each plot object in a new PR.

PR #462, opened 2016-11-18
AQCU-400: Final header format

zmoore-usgs:
Note: The new shared header file applies to every report except for V-Diagram because for some reason the custom rendering that V-Diagram uses causes it to be unable to access the shared-header.rmd file. The logic for V-Diagrams is present in the header, however, so that it can be easily transitioned over in the future.


lindsayplatt:
added v-diagram rendering pattern as technical debt

PR #467, opened 2016-11-22
fixing check problems

mhines-usgs:


lindsayplatt:
wait...for reals....

mhines-usgs:
@lindsaycarr i don't think these are showstoppers because they're not causing dysfunction, just found a bunch of problems in the check that are errors that probably should be addressed and could be why travis is sad

dpattermann-usgs:
Didn't @zmoore-usgs remove the Pandoc/"html" options recently? Sorry if I'm out of the loop.

zmoore-usgs:
Yeah, I removed those here: https://github.com/USGS-R/repgen/pull/485

zmoore-usgs:
Whoops, meant to just click "comment" not "comment and close". Sorry! :slightly_frowning_face:

PR #468, opened 2016-11-28
AQCU-904: Treat optionally user-excluded zero and negative values as gaps.

zmoore-usgs:


R/utils-json.R
```diff
@@ -81,6 +81,43 @@ negValues <- function(data, val_nm){
 }
 
 #' @export
+# adds periods of zero or negative data to the gaps field of the specified ts
+addZeroNegativeGaps <- function(field, data, isDV){
+  #Ensure we are supposed to remove zeros and negatives before doing so
```
lindsayplatt, line 86: good idea!

R/utils-json.R
```diff
@@ -81,6 +81,43 @@ negValues <- function(data, val_nm){
...
+  if(!loggedData || isEmptyOrBlank(flagZeroNeg) || !flagZeroNeg){
+    return(NULL)
+  }
+  
+  y <- data[[field]]$points[['value']]
+  x <- data[[field]]$points[['time']]
+  
+  if(!is.null(y) & !is.null(x)){
+    time <- flexibleTimeParse(x, data$reportMetadata$timezone, isDV)
+    uv_series <- data.frame(time=time, rawTime=x, value=y, stringsAsFactors = FALSE)
```
lindsayplatt: L93:98 could be cleaner w/ dplyr rather than creating a new df:

```r
uv_series <- data[[field]]$points
if(!is.null(uv_series) & nrow(uv_series) != 0){
   uv_series <- uv_series %>% 
    rename(rawTime = time) %>% 
    mutate(time = flexibleTimeParse(rawTime, data$reportMetadata$timezone, isDV)) %>% 
    select(time, rawTime, value)
```

R/utils-json.R
```diff
@@ -81,6 +81,43 @@ negValues <- function(data, val_nm){
...
+
+    #Select times from each point that will be excluded
+    newGaps <- uv_series %>% filter(value > 0) %>% select(time, rawTime)
```
lindsayplatt: maybe rename `potentialNewGaps`

R/utils-json.R
```diff
@@ -81,6 +81,43 @@ negValues <- function(data, val_nm){
...
+
+    #Determine start / end times for gaps created by these points
+    if(isDV){
+      newGaps <- newGaps %>% mutate(diff = c(difftime(tail(strptime(time, "%Y-%m-%d %H:%M:%S"), -1), head(strptime(time, "%Y-%m-%d %H:%M:%S"), -1), units="days"),0), prev = lag(diff))
+      startGaps <- newGaps %>% filter(diff > 1) %>% select(rawTime)
+      endGaps <- newGaps %>% filter(prev > 1) %>% mutate(rawTime = as.character(as.Date(rawTime, "%Y-%m-%d") - as.difftime(1, unit="days"))) %>% select(rawTime)
```
lindsayplatt: I don't follow why this line is necessary

lindsayplatt: If it's not, I think we would want to consolidate these lines -- there's a lot of duplication.

```r
gapTolerance <- ifelse(isDV, 1, 15)
gapUnits <- ifelse(isDV, "days", "mins")
newGaps <- newGaps %>% mutate(diff = c(difftime(tail(strptime(time, "%Y-%m-%d %H:%M:%S"), -1),
                                                head(strptime(time, "%Y-%m-%d %H:%M:%S"), -1), 
                                                units=gapUnits),0), 
                              prev = lag(diff))
startGaps <- newGaps %>% filter(diff > gapTolerance) %>% select(rawTime)
endGaps <- newGaps %>% filter(prev > gapTolerance) %>% select(rawTime)
```

R/utils-json.R
```diff
@@ -81,6 +81,43 @@ negValues <- function(data, val_nm){
...
+    } else {
+      newGaps <- newGaps %>% mutate(diff = c(difftime(tail(strptime(time, "%Y-%m-%d %H:%M:%S"), -1), head(strptime(time, "%Y-%m-%d %H:%M:%S"), -1), units="mins"),0), prev = lag(diff))
+      startGaps <- newGaps %>% filter(diff > 15) %>% select(rawTime)
+      endGaps <- newGaps %>% filter(prev > 15) %>% select(rawTime)
+    }
+    
+    appGaps <- do.call(rbind, Map(data.frame, startTime=startGaps, endTime=endGaps))
```
lindsayplatt: I think this can just be: `appGaps <- data.frame(startTime = startGaps, endTime = endGaps)`

lindsayplatt: To add to the actual gaps and then return all of them (add `gaps` as an arg in this function):

```r
if(!isEmptyOrBlank(gaps) && nrow(gaps) != 0){
    gaps <- rbind(gaps, appGaps)
} else {
    gaps <- appGaps
}

return(gaps)
```

R/utils-shared.R
```diff
@@ -277,6 +277,14 @@ isNullOrFalse <- function(variable) {
 splitDataGaps <- function(data, ts, isDV){
   
   data_list <- data[[ts$field[1]]]
+
+  #Add zero/negative gaps
+  if("gaps"  %in% names(data_list)){
```
lindsayplatt, line 283: seems like you would want to add the zero negative gaps to the `gaps` field even if it does not exist yet, and that should happen inside of `addZeroNegativeGaps`. Otherwise it should be called `findZeroNegativeGaps`, and then be added as it is doing here. See comment on L114 of `addZeroNegativeGaps`.

If you keep it as `addZeroNegativeGaps`:
```r
 #Add zero/negative gaps
 data_list$gaps <- addZeroNegativeGaps(data_list$gaps, ts$field[1], data, isDV)
```

Otherwise:
```r
#Add zero/negative gaps
zeroNegativeGaps <- findZeroNegativeGaps(ts$field[1], data, isDV)
if("gaps"  %in% names(data_list)){
    data_list$gaps <- rbind(data_list$gaps, zeroNegativeGaps)
} else {
    data_list$gaps <- zeroNegativeGaps
}
```

lindsayplatt (changes requested):

lindsayplatt:
Also, don't forget to remove the disclaimers: https://github.com/USGS-R/repgen/pull/451/commits/e7b94bafd77445939a03024fdeb6568abce55f45

R/utils-json.R
zmoore-usgs: That line was necessary prior to my fix on lines 365/366 as the last day in the gap was still getting cutoff. I forgot to change that after I fixed the other issue so the shorter version you suggested should work now.

R/utils-json.R
```diff
@@ -81,6 +81,42 @@ negValues <- function(data, val_nm){
...
+    gapTolerance <- ifelse(isDV, 1, 15)
+    gapUnits <- ifelse(isDV, "days", "mins")
+    potentialNewGaps <- potentialNewGaps %>% mutate(diff = c(difftime(tail(strptime(time, "%Y-%m-%d %H:%M:%S"), -1),
+                                                    head(strptime(time, "%Y-%m-%d %H:%M:%S"), -1), 
+                                                    units=gapUnits),0), 
+                                                    prev = lag(diff))
+    startGaps <- potentialNewGaps %>% filter(diff > gapTolerance) %>% select(rawTime)
+    endGaps <- potentialNewGaps %>% filter(prev > gapTolerance) %>% select(rawTime)
+    
+    appGaps <- data.frame(startTime = unname(startGaps), endTime = unname(endGaps))
```
lindsayplatt: this is just a bit awkward since `startGaps` and `endGaps` are data frames. The cleaner way might be `data.frame(startTime = startGaps$rawTime, endTime = endGaps$rawTime)`. That way you are building a dataframe from two vectors rather than two, single column dataframes

R/utils-json.R
```diff
@@ -81,6 +81,42 @@ negValues <- function(data, val_nm){
...
+    #Select times from each point that will be excluded
+    potentialNewGaps <- uv_series %>% filter(value > 0) %>% select(time, rawTime)
+
+    #Determine start / end times for gaps created by these points
+    gapTolerance <- ifelse(isDV, 1, 15)
+    gapUnits <- ifelse(isDV, "days", "mins")
+    potentialNewGaps <- potentialNewGaps %>% mutate(diff = c(difftime(tail(strptime(time, "%Y-%m-%d %H:%M:%S"), -1),
+                                                    head(strptime(time, "%Y-%m-%d %H:%M:%S"), -1), 
+                                                    units=gapUnits),0), 
+                                                    prev = lag(diff))
```
lindsayplatt, line 109: getting realllllyyy nitpicky here, but why not?? highlight this section and click CTRL+I. Some of the lines should be indented differently

lindsayplatt (approved):

PR #469, opened 2016-11-28
AQCU-866: Fixed an issue where qualifiers containing HTML special…

zmoore-usgs:
…characters would not render properly. Also fixed an issue where Site Visit Peak report qualifiers tables would not always contain all qualifiers when main table rows had multiple qualifiers.

### DIFF
--- a/R/sitevisitpeak-data.R
+++ b/R/sitevisitpeak-data.R
@@ -196,7 +196,7 @@ containsOutsideUncertainty <- function(data) {
 #'@export
 svpQualifiersTable <- function(data, table){
   #Construct List of all qualifiers
-  qualifiersList <- data.frame(unlist(data$readings$associatedIvQualifiers, recursive=FALSE))
+  qualifiersList <- do.call("rbind", data$readings$associatedIvQualifiers)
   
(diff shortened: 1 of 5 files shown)


R/sitevisitpeak-data.R
```diff
@@ -196,7 +196,7 @@ containsOutsideUncertainty <- function(data) {
 #'@export
 svpQualifiersTable <- function(data, table){
   #Construct List of all qualifiers
-  qualifiersList <- data.frame(unlist(data$readings$associatedIvQualifiers, recursive=FALSE))
+  qualifiersList <- do.call("rbind", data$readings$associatedIvQualifiers)
```
lindsayplatt: one of the tests is failing because `data$readings$associatedIvQualifiers` is NULL. You'll need to safeguard against that. I suggest:

```r
if(!isEmptyOrBlank(data$readings$associatedIvQualifiers)){
    qualifiersList <- do.call("rbind", data$readings$associatedIvQualifiers)
} else {
    qualifiersList <- data.frame()
}
```

lindsayplatt (changes requested):

lindsayplatt (approved):

PR #470, opened 2016-11-29
AQCU-866: Fixed HTML special characters not showing up in the main report tables...

zmoore-usgs:
... for SVP and SRS reports. Had to use a hacky character replace fix due to pre-existing HTML hacks that require the use of the HTMLTable function for table rendering and won't allow the use of kable() for  table rendering which would fix the problems on its own.


lindsayplatt:
@zmoore-usgs what are the pre-existing HTML hacks? Should we capture those in tech debt?

zmoore-usgs:
@lindsaycarr https://github.com/USGS-R/repgen/blob/master/R/sensorreading-data.R#L125

SVP doesn't seem to have a similar hack but I stuck with using the same HTMLTable function for consistency and in order to keep the groups and mixed alignment.

I do think this is something that should be captured by in tech debt.

lindsayplatt:
Ok, I've added it.

lindsayplatt (approved):

PR #471, opened 2016-11-29
Fixed an issue where empty qualifiers lists in SVP and SRS reports…

zmoore-usgs:
…would cause rendering errors.


mhines-usgs:
Pulled this locally and the svp and srs reports test just fine with a bunch of example files.

PR #474, opened 2016-12-01
Omit min./max. shifts when min., max., or both are not present

ashalper-usgs:
Modified to omit plotting of min./max. shifts on UV hydrographs when min., max., or both are not present in data.

### DIFF
--- a/R/utils-json.R
+++ b/R/utils-json.R
@@ -175,14 +175,45 @@ getFieldVisitMeasurementsShifts <- function(ts){
     df <- na.omit(df)
     return(df)
   }
-  y <- ts$fieldVisitMeasurements[['shiftInFeet']]
-  x <- ts$fieldVisitMeasurements[['measurementStartDate']]
-  minShift <- ts$fieldVisitMeasurements[['errorMinShiftInFeet']]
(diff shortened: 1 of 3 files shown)


R/utils-json.R
```diff
@@ -175,14 +175,45 @@ getFieldVisitMeasurementsShifts <- function(ts){
...
+      minShift <- c(minShift, errorMinShiftInFeet[i])
+      maxShift <- c(maxShift, errorMaxShiftInFeet[i])
+    }
+  }
+  
   time = as.POSIXct(strptime(x, "%FT%T"))
-  month <- format(time, format = "%y%m") #for subsetting later by month
-  return(data.frame(time=time, value=y, minShift=minShift, maxShift=maxShift, month=month, 
-                    field=rep("fieldVisitMeasurements", length(time)), stringsAsFactors = FALSE))
+  month <-
```
mhines-usgs: if none of this from here down changed, i think you wouldn't want to check it in?

ashalper-usgs: Auto-formatting by RStudio so I could read the code better. I'll undo this part.

PR #475, opened 2016-12-05
Corrected approval bar positioning for linear case in ApprovalBarY()

ashalper-usgs:


R/utils-shared.R
```diff
@@ -537,12 +537,7 @@ ApprovalBarY <- function(lim, ylog = NULL, reverse, ratio) {
     }
   }
   else {
-    if (reverse) {
-      y <- e.1 + ratio * (e.1 - e.0)
```
lindsayplatt, line 541: The change you made corrects the issue for linear, but I think there would also be an issue with logged here. So, I believe we would want to change everything to:

```r
if (ylog) {
  buff <- ratio * (log10(e.1) - log10(e.0))
  y <- 10^(log10(e.0) - buff)
} else {
  buff <- ratio*(e.1 - e.0)
  y <- e.0 - buff
}
```

lindsayplatt (changes requested):

R/utils-shared.R
ashalper-usgs: OK, done.

lindsayplatt (approved):

PR #479, opened 2016-12-07
AQCU-916: Footnote labels in chart overlap, making them indistinguishable

zmoore-usgs:


R/uvhydrograph-data.R
```diff
@@ -225,17 +225,63 @@ parseLabelSpacing <- function(data, info) {
...
-      }
-      spacingInfo <- list(y=y_positions, label=seq(length(data[[1]]$time)))
-    } else {
-      spacingInfo <- list(y=y_positions, label=seq(length(data[[1]]$time)))
-    }
+
+    #The percentage of the y-range to subtract each time we add a new label to a column
+    subtractor <- (limits$ylim[[2]] - limits$ylim[[1]]) * 0.065
+
+    #Generate necessary data for corrections label spacing
```
zmoore-usgs: Yeah, the stuff below is a really huge and ugly piped dplyr call.. I spent quite awhile working on the logic to try to make it shorter but couldn't come up with anything shorter than this. If you have any suggestions let me know! I tried to comment it well enough that it makes sense but I'd be happy to walk through the logic, it's certainly possible that I am doing more than necessary or that I missed something.

R/uvhydrograph-styles.R
```diff
@@ -23,14 +33,19 @@ getUvStyle <- function(data, info, correctionLabels, plotName, dataSides, dataLi
...
                 corr_UV = list(lines = list(x=x, y=y, ylim=dataLimits$primary, ylab=primary_lbl, ann=TRUE, col="black", lty=1, legend.name=paste("Corrected UV", primary_lbl))),
                 est_UV = list(lines = list(x=x, y=y, col="orange", lty=4, lwd=2, legend.name=paste("Estimated UV", primary_lbl))),
                 uncorr_UV = list(lines = list(x=x, y=y, col="darkturquoise", lty=4, legend.name=paste("Uncorrected UV", primary_lbl))),
                 comp_UV = list(lines = list(x=x, y=y, ylim=dataLimits$comparison, side=dataSides$comparison, axes=compAxes, ylab=compLabel, ann=compAnnotations, col="green", lty=1, legend.name=comp_lbl)), 
-                series_corr = list(abline=list(v=x, untf=FALSE, col="blue", legend.name=paste("Data correction entry", primary_lbl)),
-                                   text=list(x=x, y=correctionLabels$y, label=correctionLabels$label, pos=4, col="blue")), 
+
+                series_corr = list(lines=list(side=7, x=0, y=0, xlim = c(plotStartDate, plotEndDate), axes=FALSE),
+                                    abline=list(v=x, untf=FALSE, col="blue", legend.name=paste("Data correction entry", side=7, axes=FALSE, primary_lbl)),
```
lindsayplatt: `side=7` & `axes=FALSE` are inside the paste() function right now! I assume you want them outside of it :)

lindsayplatt: also, can you explain your reason for using side 7? I removed the instances of `side=7` & `axes=FALSE`, and it seemed to look the same in my example

R/uvhydrograph-styles.R
```diff
@@ -23,14 +33,19 @@ getUvStyle <- function(data, info, correctionLabels, plotName, dataSides, dataLi
...
+                                    arrows=list(x0=corrArrowPositions$xorigin, x1=corrArrowPositions$x, y0=corrArrowPositions$y, side=7, axes=FALSE, y1=corrArrowPositions$y, col="blue", code=1, length = 0),
+                                    points=list(x=correctionLabels$x, y=correctionLabels$y, pch=0, col=rgb(0,0,255,180,max=255), side=7, axes=FALSE, cex=correctionLabels$r),
+                                    points=list(x=correctionLabels$x, y=correctionLabels$y, pch=15, col=rgb(255,255,255,125,max=255), side=7, axes=FALSE, cex=correctionLabels$r-0.1),
```
lindsayplatt: Boxes pointing to their respective correction lines = :heart_eyes:  

I **_think_** you can get rid of this points call by changing the other point type and making it's background color transparent white:

```r
points=list(x=correctionLabels$x, y=correctionLabels$y, pch=22, col=rgb(0,0,255,180,max=255), , bg=rgb(255,255,255,125,max=255), cex=correctionLabels$r)
```
It makes the boxes just a little bit smaller, so you might want to increase `correctionLabels$r` a wee bit.

R/uvhydrograph-styles.R
```diff
@@ -51,12 +66,17 @@ getUvStyle <- function(data, info, correctionLabels, plotName, dataSides, dataLi
...
                 effect_shift = list(lines=list(x=x,y=y, type='l', col = 'green3', lty = 1, lwd=2, side=4, legend.name=paste(secondary_lbl, info$tertiary_lbl)),
                                     text=list(x=x[1], y=y[1], labels="", side=4)),
+                
+                series_corr2 = list(lines=list(side=7, x=0, y=0, xlim = c(plotStartDate, plotEndDate), axes=FALSE),
+                                    abline=list(v=x, untf=FALSE, col="blue", legend.name=paste("Data correction entry", side=7, axes=FALSE, secondary_lbl)),
+                                    arrows=list(x0=corrArrowPositions$xorigin, x1=corrArrowPositions$x, y0=corrArrowPositions$y, side=7, axes=FALSE, y1=corrArrowPositions$y, col="blue", code=1, length = 0),
+                                    points=list(x=correctionLabels$x, y=correctionLabels$y, pch=0, col=rgb(0,0,255,180,max=255), side=7, axes=FALSE, cex=correctionLabels$r),
+                                    points=list(x=correctionLabels$x, y=correctionLabels$y, pch=15, col=rgb(255,255,255,125,max=255), side=7, axes=FALSE, cex=correctionLabels$r-0.1),
```
lindsayplatt: remove this line if it works out above!

R/uvhydrograph-styles.R
```diff
@@ -6,6 +6,16 @@ getUvStyle <- function(data, info, correctionLabels, plotName, dataSides, dataLi
   comp_lbl <- info$comp_UV_lbl
   comp_type <- info$comp_UV_type
   legend.name <- data[[1]]$legend.name
+  corrArrowPositions <- list()
+  plotEndDate <- tail(info$plotDates,1) + hours(23) + minutes(45)
+  plotStartDate <- info$plotDates[1]
+
+  #Make the correction label lines connect to the outside of the bounding box and not to the center of the label
+  if(!isEmptyOrBlank(correctionLabels)){
+    corrArrowPositions <- correctionLabels %>% as.data.frame() %>% select(x, xorigin, r, y) %>%
```
lindsayplatt, line 16: `x` in `correctionLabels` is coming through as a number, not as POSIXct. Seems strange since `xorigin` is POSIXct. Works fine right now, but I'm always skeptical about differing data formats when it comes to timezones, etc. 

BTW, I'm using JSON from the example mentioned in the ticket.

R/uvhydrograph-styles.R
```diff
@@ -6,6 +6,16 @@ getUvStyle <- function(data, info, correctionLabels, plotName, dataSides, dataLi
   comp_lbl <- info$comp_UV_lbl
   comp_type <- info$comp_UV_type
   legend.name <- data[[1]]$legend.name
+  corrArrowPositions <- list()
+  plotEndDate <- tail(info$plotDates,1) + hours(23) + minutes(45)
```
lindsayplatt: Should we use `toEndOfMonth` here?

R/uvhydrograph-data.R
```diff
@@ -215,8 +215,8 @@ parseUVSupplemental <- function(data, plotName, pts, useDownsampled=FALSE) {
 correctionsTable <- function(data) {
   if (any(names(data) %in% c("series_corr", "series_corr_ref", "series_corr_up", "series_corr2"))) {
     corrections <- data[[grep("series_corr", names(data))]]
-    corrections_table <- as.data.frame(cbind(seq(nrow(corrections)), corrections$comment))
-    colnames(corrections_table) <- c("", "Comments")
+    corrections_table <- as.data.frame(cbind(seq(nrow(corrections)), as.character(corrections$time), corrections$comment))
+    colnames(corrections_table) <- c("", "Time", "Comments")
```
lindsayplatt, line 219: :+1:

R/uvhydrograph-data.R
```diff
@@ -225,17 +225,56 @@ parseLabelSpacing <- function(data, info) {
...
-    if(length(differences) > 0) {
-      for (i in seq_len(length(differences))) {
-        if(abs(differences[i]) < 86400) {y_positions[i+1] <- y_positions[i]-(0.04*info$lims_UV$ylim[2])}
-        i <- i + 1
-      }
-      spacingInfo <- list(y=y_positions, label=seq(length(data[[1]]$time)))
-    } else {
-      spacingInfo <- list(y=y_positions, label=seq(length(data[[1]]$time)))
-    }
+    #Number of hours to offset labels by for display
```
lindsayplatt: Aren't all of these numbers in seconds? 60 sec/min * 60 min/hr = # seconds per hour? Then 4 is just an arbitrary number? If you do `as.numeric(as.POSIXct("2015-10-01 00:00:00"))` you will get a number of seconds since some date in 1970 (I believe 1972-01-01). So all numeric dates are stored as a number of seconds - that's why I ask.

R/uvhydrograph-data.R
```diff
@@ -225,17 +225,56 @@ parseLabelSpacing <- function(data, info) {
...
+    digitHours <- 4 * 60 * 60
+    #Total width of both bounding box left and right margins 
+    baseBoxSize <- 4 * 60 * 60
+    #Minimum space between the right side of a label box and the next correction line to not merge columns
+    minSpacerSize <- 2 * 60 * 60
+    #The percentage of the y-range to subtract each time we add a new label to a column
+    subtractor <- (limits$ylim[[2]] - limits$ylim[[1]]) * 0.065
+
+    #Save original order as label and re-order by time and then by label (descending)
+    corrs <- data[[1]] %>% select(time) %>% mutate(label = row_number()) %>% arrange(time, desc(label))
```
lindsayplatt, line 302: Jordan W suggests doing this in a more recursive fashion. Rather than making a column that determines which row fit some criteria (e.g. need a new column), and then using cumsum, maybe we should be doing something that extracts groups by time and applies some function, then continues? The current way makes us nervous about making any changes to this -- it would be very confusing. Let's chat about some possible pair coding on Monday.

lindsayplatt: All in all though, this functions well - just would be nervous about future changes/updates made by people who were not the creators.

R/uvhydrograph-styles.R
lindsayplatt: The `cumsum` functions using `xpos` is what changes the value from POSIXct to numeric. I think we can just convert back, but you'll need the timezone. So, just add `timezone <- getReportMetadata(data,'timezone')` right below the `isInverted` line in `parseUVSupplemental` and then you can use the timezone inside of `parseLabelSpacing`  by saying `info$timezone`.

lindsayplatt (changes requested):
overall this works really well - I think we should consider some pair coding on the major calculation section to make it more clear.

R/uvhydrograph-data.R
zmoore-usgs: Yeah, so this is the part that took me the longest to figure out and it is pretty hacky as is. 

These numbers are the "width" of various things in hours once they are rendered onto the plot. So with the current font and size settings in uvhydrograph-styles for the labels one digit ends up being 4 hours "wide" when it is rendered onto the plot. I need to know the final rendered size of the boxes in order to check for overlap and this was the only way I could figure out to do that without rendering the plot before adding the text (in order to be able to use the width function to get the width that the text would actually be). These values all end up being in in hours and they are used to space the labels and check for overlap.

If there is a cleaner way to get the final rendered size of things that would certainly be an improvement here. The 0.065 on line 237 and the 0.475 on line 275 are also arbitrary numbers that I got after changing the sizes and then rendering the plot until I found sizes that fit well. If there was a way to convert between times and the actual plot scale then these could be significantly cleaned up as well.

R/uvhydrograph-data.R
zmoore-usgs: I agree that this is pretty confusing and hard to follow as-is. I think pair programming sounds like a good idea. If I am able to make it into the office this afternoon I'll let you know. In the meantime I'll do some thinking about it and some tinkering with it and see if I can come up with some ideas to make it more recursive and less dplyr-dependent.

R/uvhydrograph-styles.R
zmoore-usgs: Yeah probably, I just copied this line from uvhydrograph-render.R

R/uvhydrograph-styles.R
zmoore-usgs: Whoops, yep thank you!

So I used side 7 because when the corrections are on a side < 7 they are covered by any data that is on sides 4 or 6. I wasn't sure if they should be rendered on top of or below data on those other axes, but the only way I could get them on top of the data on axes 4 and 6 was to put them on a side with a number greater than both, so I went with 7. If you take the Gage Height JSON example from the ticket and use that as the Upchain TS rather than the Primary TS then the effective shifts will plot on top of the corrections without side=7.

R/uvhydrograph-styles.R
zmoore-usgs: I'll give that a shot, thanks!

R/uvhydrograph-data.R
lindsayplatt: I think the method here is fine, but I think you mean that the number is actually the number of _seconds_ in 4 hours. Just was wondering about your comments which say "Number of hours to offset" when the value is really in seconds. It was just a minor detail :) Was just doing some dimensional analysis to check that:

4 hrs x (60 s/ 1 min) x (60 min / 1 hr) = 14400 seconds (which is the value sitting in `hourOffset`)

zmoore-usgs: Ah yeah that's correct, I'll update the documentation! Thanks for catching that!

R/uvhydrograph-styles.R
lindsayplatt: actually, I think this is different. `toEndOfMonth` will put it  on the last day of that month, but that might not be the last day of the data. It might be worth adding `plotStartDate` and `plotEndDate` to `primaryInfo` so that you don't have to do this calc in more than one location. Then you can just change anything that uses it to `primaryInfo$plotEndDate`

R/uvhydrograph-data.R
lindsayplatt: Just watch the variable names, too!

R/uvhydrograph-data.R
```diff
@@ -215,27 +215,120 @@ parseUVSupplemental <- function(data, plotName, pts, useDownsampled=FALSE) {
...
+    corrections_table <- as.data.frame(cbind(seq(nrow(corrections)), as.character(corrections$time), corrections$comment))
+    colnames(corrections_table) <- c("", "Time", "Comments")
     return(corrections_table)
   } else (return(corrections_table <- NULL))
 }
 
+addGroupCol <- function(data, newColumnName, isNewCol, newGroupValue=NULL, groupChildValue=NULL, vars=NULL){
+  build_vec <- c()
+  prev <- NULL
+  for(r in 1:nrow(data)){
```
lindsayplatt, line 227: I think this gets at what we discussed yesterday when talking about loops vs apply statements for efficiency. We are looping over the data, but this data is only comments so it should have less than 100 (presumably). So in this case, I think using the loop is fine. Just something to make note of for the future.

lindsayplatt (approved):

lindsayplatt:
Passed checks locally

PR #480, opened 2016-12-12
Moved readings to primary plot

thongsav-usgs:


lindsayplatt (approved):
as long as Laura :+1: the uncertainty implementation, I'd say we are good to go on this

PR #482, opened 2016-12-16
AQCU-928: DV Hydro- Step plots are not connected going from estimated to not estimated, and not estimated to estimated

zmoore-usgs:


R/dvhydrograph-data.R
```diff
@@ -163,7 +180,7 @@ extendStep <- function(toPlot){
   isStep <- 'type' %in% names(toPlot) && toPlot[['type']] == "s"
   
   if(isStep){
-    toPlot$x <- c(toPlot$x,  tail(toPlot$x, 1) + 90000) #this is 2400, if changing to POSTLT, need to use lubridate
```
zmoore-usgs: This change was necessary because the step of each period was extending past its day marker and into the next period, causing the new vertical steps to not line up properly.

R/dvhydrograph-data.R
```diff
@@ -163,7 +180,7 @@ extendStep <- function(toPlot){
...
+    toPlot$x <- c(toPlot$x,  tail(toPlot$x, 1) + 87000) #this is 2400, if changing to POSTLT, need to use lubridate
```
zmoore-usgs: This change was necessary because the step of each period was extending past its day marker and into the next period, causing the new vertical steps to not line up properly.

dpattermann-usgs: Talked to you about this but consider changing to 60*60*24 (86,400) which is the number of seconds in a day or making a variable containing the number of seconds in a day.

R/dvhydrograph-data.R
```diff
@@ -163,7 +180,7 @@ extendStep <- function(toPlot){
   isStep <- 'type' %in% names(toPlot) && toPlot[['type']] == "s"
   
   if(isStep){
-    toPlot$x <- c(toPlot$x,  tail(toPlot$x, 1) + 90000) #this is 2400, if changing to POSTLT, need to use lubridate
+    toPlot$x <- c(toPlot$x,  tail(toPlot$x, 1) + 60*60*24) #this is 2400, if changing to POSTLT, need to use lubridate
```
lindsayplatt: can you change the comment associated with this to make it a little clearer? like what does this even do? "this is 2400" doesn't really mean much - I realize that's what was there before, but would be nice to update!

R/dvhydrograph-data.R
```diff
@@ -103,6 +105,21 @@ parseRefData <- function(data, series) {
   return(plotData)
 }
 
+parseEstimatedEdges <- function(data, allVars){
```
lindsayplatt: add description & argument definitions to this (trying to be better about documentation whenever we add new functions)

lindsayplatt: I don't think you need to pass in `data`. You don't use it and then overwrite it on line 115 anyways :)

R/dvhydrograph-data.R
```diff
@@ -103,6 +105,21 @@ parseRefData <- function(data, series) {
...
+  datasets <- unlist(allVars, recursive = FALSE)[which(grepl("stat\\d", names(unlist(allVars, recursive = FALSE))))]
+  times <- datasets[which(grepl("time", names(datasets)))]
+  values <- datasets[which(grepl("value", names(datasets)))]
+  estData <- data.frame(time=unname(unlist(times[which(grepl("est", names(times)))])), value=unname(unlist(values[which(grepl("est", names(values)))])), set="est")
+  statData <- data.frame(time=unname(unlist(times[which(!grepl("est", names(times)))])), value=unname(unlist(values[which(!grepl("est", names(values)))])), set="stat")
```
lindsayplatt: not a huge fan of what's happening here at the beginning - you go through and grab all of the times and values into single lists and then separate back out into est and stat dataframes. I think the more logical way would be to pull the est and stat data from allVars separately:

```r
est_df_list <- lapply(allVars[grepl('est_stat\\d', names(allVars))], as.data.frame, stringsAsFactors=F)
estData <- bind_rows(est_df_list) %>% select(-legend.name) %>% mutate(set=rep('est', nrow(.)))

stat_df_list <- lapply(allVars[grepl('^stat\\d', names(allVars))], as.data.frame, stringsAsFactors=F)
statData <- bind_rows(stat_df_list) %>% select(-legend.name) %>% mutate(set=rep('stat', nrow(.)))
```

then you can continue on to the code below.

R/dvhydrograph-styles.R
```diff
@@ -13,6 +13,7 @@ getDvStyle <- function(data, info = NULL, ...){
                    est_stat2 = list(lines = list(x=x, y=y, type="s", col="red2", lty=3, pch=20, cex=0.5, legend.name=legend.name)),
                    est_stat3 = list(lines = list(x=x, y=y, type="s", col="red3", lty=6, pch=20, cex=0.5, legend.name=legend.name)),
                    est_comp = list(lines = list(x=x, y=y, type="s", col="red4", lty=6, pch=20, cex=0.5, legend.name=legend.name)),
+                   estEdges = list(arrows = list(x0=data[[1]]$x, x1=data[[1]]$x, y0=data[[1]]$y0, y1=data[[1]]$y1, lwd=1, lty=ifelse(data[[1]]$newSet == "est", 1, 2), col=ifelse(data[[1]]$newSet == "est", "blue", "red"), code=1, length = 0)),
```
lindsayplatt: I like this implementation - in my head I was picturing adding a new point somewhere to make the lines connect, but this makes more sense.

lindsayplatt (changes requested):
functional, but suggested some small improvements - mostly unnecessary code

PR #485, opened 2016-12-19
AQCU-937: Delete PDF-related bits from Java and Repgen

zmoore-usgs:


DESCRIPTION
```diff
@@ -21,8 +21,7 @@ Imports:
     knitr,
     dplyr,
     lubridate,
-    htmlTable,
```
zmoore-usgs, line 24: I'm not entirely sure if we were intending to go this far (removing related libraries completely) when removing stuff for PDFs but if we were I'm also not entirely sure if I got all of the libraries that were only needed for PDFs, so if any of the other ones can be removed I can do that as well.

I'm also not completely sure Pander wasn't being used for something, but as far as I could find it was only used in one place which wasn't necessary and everything still seems to function fine without pander installed.

lindsayplatt: ^ pander was for PDF :) I'll look for some others

lindsayplatt (approved):
Looks great - I didn't find any other PDF stuff. Checks passed locally, so ignoring broken build.

PR #486, opened 2016-12-20
AQCU 931 Fix

dpattermann-usgs:
Done by adding another case to flexibleTimeParsing.

Time was coming in as "Ymd HMS" but we weren't checking for it (instead only checking for "Ymd" for DV).


zmoore-usgs (approved):
Code changes look good to me. Might still want to have @lindsaycarr look at this if she hasn't already because she knows this function better than me, but as far as I can tell this should work and should fix the issue.

lindsayplatt:
I took a look at this earlier today and :+1: it!

lindsayplatt (approved):
:100:

PR #487, opened 2016-12-21
removing tests to see if that affects AQCU-906; the travis failure/ti…

mhines-usgs:
…meout


lindsayplatt:
Every time I get an email saying you've committed to this PR, I hold my breath.

mhines-usgs:
@lindsaycarr sorry, i should probably figure out how to add travis to my fork instead so you're all not spammed.

This is really obnoxious, it seems like it's ignoring things like NO EXAMPLES, because the job continually gets hung up on checking examples... 

It also doesn't dump the logs, no matter what command I give it. Getting old.

lindsayplatt:
@mhines-usgs when do we pull the plug and switch to Jenkins?

mhines-usgs:
@lindsaycarr probably by tomorrow in sprint planning if i don't figure it out?

PR #489, opened 2016-12-27
AQCU-916: Added a missing documentation file.

zmoore-usgs:


mhines-usgs:
Nice, saw this mucking with the 'fix travis task'. thanks! :shipit:

PR #491, opened 2016-12-28
Cleaned-up R to obviate warnings from devtools::check()

ashalper-usgs:


thongsav-usgs:
Changes now included in 500

PR #493, opened 2016-12-29
AQCU-963 Add discharge measurements to explanation for UV Hydro and DV Hydro

mhines-usgs:
-added Measurement ID as legend label to dvhydrograph (data were there but no legend entry was present)
-move legend label from error bar to point call for uvhydrograph discharge measurements


mhines-usgs:
Because these are pretty minor text changes/legend edits, going to merge.

PR #494, opened 2016-12-29
AQCU-959 / AQCU-960: V-Diagram header information is mislabelled / ...

zmoore-usgs:
... DV Hydrograph label inconsistencies

### DIFF
--- a/inst/dvhydrograph/dvhydrograph.Rmd
+++ b/inst/dvhydrograph/dvhydrograph.Rmd
@@ -20,14 +20,20 @@ cat(getLogo())
 
 `r if(!isEmptyOrBlank(data[['reportMetadata']][['primaryDescriptions']])) {paste("**Primary Time Series:**", data[['reportMetadata']][['primaryDescriptions']])}`
 
-`r if(!isEmptyOrBlank(data[['reportMetadata']][['downChainDescriptions1']])) {paste("**First Down Chain Time Series:**", data[['reportMetadata']][['downChainDescriptions1']])}`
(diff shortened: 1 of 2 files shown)


inst/dvhydrograph/dvhydrograph.Rmd
```diff
@@ -20,15 +20,21 @@ cat(getLogo())
...
 </div>
 <div class="header-col-right">
 **Period:**  `r as.Date(data[['reportMetadata']][['startDate']])` to `r as.Date(data[['reportMetadata']][['endDate']])`
 
+`r if(!isEmptyOrBlank(data[['reportMetadata']][['inputDataDescriptions2']])) {paste("**Secondary Reference Time Series:**", data[['reportMetadata']][['inputDataDescriptions2']])}`
+
+`r if(!isEmptyOrBlank(data[['reportMetadata']][['inputDataDescriptions3']])) {paste("**Tertiary Reference Time Series:**", data[['reportMetadata']][['inputDataDescriptions3']])}`
+
+`r if(!isEmptyOrBlank(data[['reportMetadata']][['inputDataDescriptions4']])) {paste("**Quaternary Reference Time Series:**", data[['reportMetadata']][['inputDataDescriptions4']])}`
+
```
dpattermann-usgs, line 37: Should we consider changing the names of these to Reference Time Series 2 (, 3, 4) to keep the convention of Stat Derived Time Series? @thongsav-usgs

zmoore-usgs: That's a good point. Changes would need to be made on the UI too in the advanced settings cog. Might be worth asking Laura?

thongsav-usgs: I would vote yes for anything making things more consistent, even the internal parameter names, but that would need to be update through the whole stack and tests.

PR #495, opened 2016-12-30
AQCU-967: set time zone in sig files so reports render as they do on the server, locally

zmoore-usgs:


mhines-usgs:
I can pull this locally to test out. I'll try changing my computer settings and see if it changes for me.

mhines-usgs:
:+1:

PR #496, opened 2017-01-03
Changed point vs. interval comparisons in extremesTable() from closed-open to closed-closed intervals

ashalper-usgs:


lindsayplatt (commented):
Looks good. Only thing I would suggest is adding a test. Here's one I came up with:

```r
qualifiers <- data.frame(startDate = "2015-11-01", endDate = "2016-11-16",
                         identifier = "ESTIMATED", code = "E", displayName = "Estimated",
                         stringsAsFactors = FALSE)

points1 <- data.frame(time = c("2016-11-15"), value = c(4.05),
                     stringsAsFactors = FALSE)
points2 <- data.frame(time = c("2016-11-16"), value = c(5.7),
                     stringsAsFactors = FALSE)

q1 <- repgen:::applyQualifiersToValues(points1, qualifiers)
expect_true(grepl("E", q1$value))

q2 <- repgen:::applyQualifiersToValues(points2, qualifiers)
expect_false(grepl("E", q2$value))
```

lindsayplatt (approved):

ashalper-usgs:
@lindsaycarr, do you know what it means when `applyQualifiersToValues()` compares:

`(2016-11-16)` to `[2016-11-15T00:00:00.0000001-05:00,2016-11-16T00:00:00.0000001)`? Is the implied local offset of `(2016-11-16)`, `-05:00`?

lindsayplatt:
When you run locally, it is whatever your computer is set at. But we should all be running the `Sys.setenv(TZ = "UTC")` command so it is in UTC. If you want, you can make those times specific.

ashalper-usgs:
So it's (implicitly) comparing the intersection of `[2016-11-16T00:00:00.0000000Z,2016-11-17T00:00:00.0000000Z)` to `[2016-11-15T00:00:00.0000001-05:00,2016-11-16T00:00:00.0000001)`? This doesn't seem correct. Shouldn't all the operand values be in the context of `-05:00`?

PR #498, opened 2017-01-05
Removing anything related to utils-auth AQCU-976

mhines-usgs:
removing examples with URLs where auth was required in sig files
removing methods where sig files were using httr/ getJSON from the URLs using auth
deleting utils-auth and utils-web because they are no longer used


mhines-usgs:
Woops, I guess we needed getLogo() in utils-web.R.. fixing.

PR #499, opened 2017-01-05
Removing utils-auth.R AQCU-976

mhines-usgs:
deleting any sig file parts where examples of fetching via authenticated URL
deleting the URL methods from sig files
removing the getJSON function and httr dependency
moved getLogo to utils-shared out of utils-web (?)


lindsayplatt:
:clap:

mhines-usgs:
@lindsaycarr or @thongsav-usgs i deleted utils-web.R because getJSON and getLogo were the only functions in there, and actually didn't see getLogo until I merged was still used (oops)..  but recovered getLogo() and i put it into utils-shared.R

At least it's working again and we can decide Monday where it should go from there?

PR #500, opened 2017-01-09
Cleaned up warnings, added roxygen

ashalper-usgs:


thongsav-usgs:
Going to redo after refactor

PR #501, opened 2017-01-09
Refactor for utils-shared

mhines-usgs:
major revision for utils-shared breakdown


mhines-usgs:
https://travis-ci.org/USGS-R/repgen/jobs/190332535

This build should be fine per this travis link ^^

PR #509, opened 2017-01-10
tests for subsetByMonth

mhines-usgs:


tests/testthat/test-utils-filter.R
```diff
@@ -104,4 +104,23 @@ test_that('do negative or zero values get removed from the data frame?', {
   expect_false(df[2,1]==0)
 })
 
+test_that('does it subset by the month I ask it to', {
+  library(jsonlite)
+  library(lubridate)
+  pts <- fromJSON(system.file('extdata','uvhydrograph','uvhydro-example.json', package = 'repgen'))
```
thongsav-usgs: Instead of using the example file, I'd copy and paste the just enough of the data here and wittle it down to just a few points (eg: 1 point for each month). That way the test explicitly shows what it is testing without an external dependency.

EG:  https://github.com/USGS-R/repgen/blob/master/tests/testthat/test-utils-fetch.R#L6

PR #521, opened 2017-01-10
Pull out timeFormatting function from siteVisitPeak.R to utils-html,

kmschoep-usgs:
make more generic


R/utils-html.R
```diff
@@ -204,4 +204,30 @@ nullMask <- function(val) {
...
+#' @param dateFormatMask String with preferred output date format
+#' @return list with date in first position, time in second position.
+timeFormatting <- function(timeVals, dateFormatMask){
+  if(!isEmpty(timeVals)) {
+    dateTime <- (strsplit(timeVals, split="[T]"))
+    dateFormat <- strftime(dateTime[[1]][1], dateFormatMask)
+    
+    #Break apart, format dates/times, put back together.
+    timeFormatting <- sapply(dateTime[[1]][2], function(s) {
+      m <- regexec("([^-+]+)([+-].*)", s)
```
dpattermann-usgs, line 222: Could you put a comment saying what this does?

kmschoep-usgs: You mean the regexec?  Not sure exactly, but the end result is that it breaks the date/time string into a date string and a time string.

PR #523, opened 2017-01-11
Reawesome: Added full test coverage and slight functionality updates to readTimeSeries and readEstimatedTimeSeries.

zmoore-usgs:


R/utils-legend.R
```diff
@@ -16,4 +16,15 @@ rmDuplicateLegendItems <- function(object){
   }
   
   return(object)
+}
+
+getTimeSeriesLabel<- function(ts, field){
```
dpattermann-usgs, line 21: Should we have kept this in the same naming convention of read/fetch?

zmoore-usgs: That function hasn't been refactored yet, just moved to the proper file from the wrong one. I'm not sure if it would fit in the read/fetch convention because it should probably instead take the parts of the name as arguments and then combine them, rather than looking them up from the JSON as it currently is doing. Maybe something like createTimeSeriesLabel?

PR #524, opened 2017-01-11
revision to utils-html to use [[]] instead of $

mhines-usgs:
and a new test for format comments still in the works. will add two or three more small tests for some of these functions but not go crazy since the long term goal is to move most of the html functions to java.


tests/testthat/test-utils-html.R
```diff
@@ -0,0 +1,49 @@
+context("utils-html tests")
+
+wd <- getwd()
+setwd(dir = tempdir())
+
+test_that('does it replace the escaped characters with real html breaks?', {
+  library(jsonlite)
+  readings <- fromJSON('
```
thongsav-usgs: FYI, as you have it written, the json is actually a "reportObject" and not a "readings" object because you have the top level nesting in the json. If you want just the json for a readings object you can drop the `{ "readings":` from the start and parse just the array. Then it would be a readings (list) object being parsed.

tests/testthat/test-utils-html.R
```diff
@@ -0,0 +1,49 @@
...
+                       "party": "ARC",
+                       "nearestcorrectedValue": "5.24",
+                       "qualifiers": [],
+                       "nearestcorrectedTime": "2016-02-29T11:00:00-05:00",
+                       "nearestrawTime": "2016-02-29T11:00:00-05:00",
+                       "nearestrawValue": "5.24"
+                       }
+                        ] }
+                       ')
+  refComm <- repgen:::getComments(readings[["readings"]][["recorderComments"]])
```
thongsav-usgs: Related to comment above, this should actually be written as `reportObject[["readings"]][["recorderComments"]]`.

thongsav-usgs (changes requested):

tests/testthat/test-utils-html.R
mhines-usgs: Yeah, I figured. I'm still working on finishing this test up and will push up the corrections. Does it matter which way I go, with the reportObject or with readings?

mhines-usgs: Going to go with reportObject as it's more 'real world'

tests/testthat/test-utils-html.R
mhines-usgs: changes to reportObject

R/utils-html.R
```diff
@@ -92,8 +92,8 @@ printReportFeature <- function(feature, isTable=FALSE, m=NULL, mar_values=c(8, 3
 formatComments <- function(comments){
   split_comments <- unlist(comments)
   if(isEmptyOrBlank(split_comments)){return(split_comments)}
-  htmlbreaks_inside <- lapply(split_comments, gsub, pattern="\r\n", replacement="</br>")
-  htmlbreaks_end <- lapply(htmlbreaks_inside, paste0, "</br>", collapse="")
+  htmlbreaks_inside <- lapply(split_comments, gsub, pattern="\r\n", replacement="<br/>")
```
mhines-usgs, line 95: according to this http://www.w3schools.com/code/tryit.asp?filename=FBNJO5J8MUTJ

this ```</br>``` works, but it could only be working because the browser is allowing it to.

PR #525, opened 2017-01-11
roxygen edits and minor correction to isEmptyVar()

ashalper-usgs:


mhines-usgs:
@ashalper-usgs -- in case you didn't get the notice in email or slack, there is a conflict here with utils-read. We're not exporting all the functions now. Can you resolve that and re-commit so this passes tests?

ashalper-usgs:
I think it's going to be easier for me to close this pull request, clean up some stuff, then open a new one.

PR #526, opened 2017-01-11
Reawesome/utils fetch: Converted getGroundWaterLevels and getWaterQualityMeasurements...

zmoore-usgs:


R/uvhydrograph-data.R
```diff
@@ -22,7 +22,11 @@ parsePrimaryUVData <- function(data, month, useDownsampled=FALSE) {
   est_UV <- subsetByMonth(getTimeSeries(data, timeSeriesNames$primarySeriesName, estimatedOnly=TRUE), month)
   uncorr_UV <- subsetByMonth(getTimeSeries(data, timeSeriesNames$primarySeriesRawName ), month)
   comp_UV <- subsetByMonth(getTimeSeries(data, timeSeriesNames$comparisonSeriesName ), month)
-  water_qual <- subsetByMonth(getWaterQualityMeasurements(data), month)
+  water_qual <- tryCatch({
+    subsetByMonth(readWaterQualityMeasurements(data), month)
```
lindsayplatt, line 26: would you want `subsetByMonth` to occur outside of the tryCatch for `readWaterQualityMeasurements`? I guess it depends on if you want the code to actually stop if there's an error with subsetting or just return an NA data frame.

lindsayplatt:
tryCatch implementation looks correct to me :+1:

R/uvhydrograph-data.R
zmoore-usgs: That's a good point, probably best to split this up into 2 separate things.

thongsav-usgs: The one liner is ok if errors from either function would be handled the same way. They only really need to be split if we have different responses to different error conditions.

PR #529, opened 2017-01-12
Reawesome/utils fetch: Shared empty data function

zmoore-usgs:
Added a shared function for checking required fields. Could use some improvement perhaps? I'm not totally sure if this is the best way to implement it.

The first part of it checks to see if the fields supplied as requiredFields all exist in the names of the data provided.

The second part is used only if the supplied data is an array read in from the JSON (which is converted into a dataframe). This ensures that all entries in the array have all of the fields supplied as requiredFields.


R/utils-read.R
```diff
@@ -9,6 +9,31 @@ sizeOf <- function(df){
   return(nrow(df))
 }
 
+#' Returns whether or not all required columns are present
+#' 
+#' 
+emptyData <- function(data, name, requiredFields){
```
thongsav-usgs: This seems like it belongs in utils-validation. The function name also doesn't  really describe what's happening.

R/utils-read.R
```diff
@@ -9,6 +9,31 @@ sizeOf <- function(df){
...
+      partialFields <- partialFields[which(names(partialFields) %in% requiredFields)]
+      
+      
+      naCols <- c(missingFields, partialFields)
+      stop(paste("Report JSON for requested field: ", name, " is missing required fields: {",  paste(naCols, collapse=', '), "}"))
+    } else {
+      return(FALSE)
+    }
+  } else if(is.null(data)){
+    stop(paste("Requested field: '", name, "' not found in report JSON."))
```
thongsav-usgs: Rather than pass the name function in for the purpose of labeling the errors, we should leave it up to the caller to make the error more specific.

I think we have the same pattern going on in validParam, I will update that as well.

thongsav-usgs (changes requested):

R/utils-read.R
zmoore-usgs: So return the list of missing values and leave it up to the caller to throw the actual stop? That makes sense to me

R/utils-fetch.R
```diff
@@ -1,6 +1,16 @@
 # This R file's purpose is for extracting values from json 
 # The functions shouldn't modify the data, and should handle missing json or empty json parameters
 
+#' Fetch Field
+#'
+#' @description Generalize fetch function that fetches the provided field
+#' @param reportObect The full report data loaded from the report JSON
+#' @param field The field to fetch from the data
+fetchReportField <- function(reportObect, field){
```
dpattermann-usgs: Fix the typo?

R/utils-fetch.R
```diff
@@ -265,4 +275,12 @@ fetchGroundWaterLevels <- function(reportObject){
 fetchWaterQualityMeasurements <- function(reportObect){
   val <- reportObect$waterQuality
   return(val)
+}
+
+#' Fetch field visit measurements
+#' @description Given a report object, will pull the field visit measurements
+#' @param reportObect the full report data
+fetchFieldVisitMeasurements <- function(reportObect){
+  val <- reportObect$fieldVisitMeasurements
```
dpattermann-usgs: Typo?

R/utils-fetch.R
zmoore-usgs: This is a function meant to fetch any supplied JSON field from the report, not a typo of fieldVists. I'll change the name to something else to make that more clear though since it does look like a typo as it is.

dpattermann-usgs: Oh, I meant reportObect vs. reportObject

zmoore-usgs: Actually went ahead and removed fetchReportField because it isn't really necessary now that we have fetch functions for every part of the reportObject.

zmoore-usgs: Oh! I didn't notice that haha, thanks!

PR #531, opened 2017-01-12
roxygen additions and edits; some normalization of function names; minor corrections to isEmptyVar() and fetchWaterQualityMeasurements()

ashalper-usgs:


NAMESPACE
```diff
@@ -14,6 +14,7 @@ export(getFieldVisitMeasurementsQPoints)
 export(getFieldVisitMeasurementsShifts)
 export(getStatDerived)
 export(getTimeSeries)
+export(isEmptyVar)
```
mhines-usgs: we probably don't need to export this function, and there are other functions we may not need to export here too. not sure what you changed versus what might be from other people's code.

inst/extdata/installDependencies.R
```diff
@@ -79,7 +79,16 @@ InstallPackages(pkgs, "http://mran.microsoft.com/snapshot/2016-03-31")
 
 # packages that must be held back at older versions
 InstallPackages("httr", "http://mran.microsoft.com/snapshot/2016-01-27")
-InstallPackages("devtools", "http://mran.microsoft.com/snapshot/2015-03-13")
+
```
mhines-usgs, line 82: i'm not sure this is something we want in the code repository? @thongsav-usgs (all the cruft about local things todo)

mhines-usgs (commented):
in addition to these small things i commented, it seems like we're inconsistently using @return in the documentation and in functions. not sure if there is a direction we should try to converge upon there.

inst/extdata/installDependencies.R
ashalper-usgs: This is an AQCU-949/AQCU-936 patch that accidentally crept into my `re-awesome` branch. It's obsolete now, and irrelevant anyway, as `installDependencies.R` is not being invoked by Jenkins now. Stay tuned for new-and-improved `installDependencies.R`, as well as invocation by the Jenkins deployer.

NAMESPACE
ashalper-usgs: OK, I removed the @export tag.

inst/extdata/installDependencies.R
mhines-usgs: ahh, good point, i didn't realize this was your .R file, not the .sh file! derp.

ashalper-usgs: The R script was initially a line-by-line translation of the shell script, so honest mistake! :)

thongsav-usgs:
@ashalper-usgs I'm going to close this as the majority of these files have changed drastically with our big cleanup sprint.

PR #532, opened 2017-01-13
adding utils-html tests and modifying utils-html

mhines-usgs:
this might break some reports but I can update those as well so they don't break


R/utils-html.R
```diff
@@ -107,14 +107,13 @@ formatComments <- function(comments){
 #' 
 #' @return the HTML link for SIMS url
 #' 
-getSimsUrl<- function(reportObject){
-  url <- reportObject[["simsUrl"]]
-  if(isEmptyOrBlank(url)) {
-    url <- "SIMS URL: NA"
+getSimsUrl<- function(simsUrl){
```
mhines-usgs, line 110: changed this from the entire reportObject to just use the URL from the object

R/utils-html.R
```diff
@@ -126,14 +125,13 @@ getSimsUrl<- function(reportObject){
 #'
 #'@return The HTML link for waterdata url
 #'
-getWaterDataUrl <- function(reportObject) {
-  url <- reportObject[["waterdataUrl"]]
-  if (isEmptyOrBlank(url)) {
-    url <- "waterdata.usgs.gov URL: NA"
+getWaterDataUrl <- function(waterdataUrl) {
```
mhines-usgs, line 128: ditto as above

mhines-usgs:
Now with report updates bonus

PR #533, opened 2017-01-13
Utils-limits tests

dpattermann-usgs:
Created tests for utils-limits and changed testCalloutsByView to not use xtfrm


R/utils-limits.R
```diff
@@ -47,14 +46,12 @@ testCallouts <- function(plot_obj, xlimits){
...
+  i_view <- which(names(plot_obj[[view_num]]) == "callouts")
   
   for(i in i_view){
     callout_args <- plot_obj[[view_num]][[i]]
-    if (!is.na(xtfrm(callout_args$x[i])) |  
-        !is.na(xtfrm(callout_args$y[i])) |
-        is.null(xtfrm(callout_args$x[i])) |  
-        is.null(xtfrm(callout_args$y[i]))) {  
+    
+    if (!isEmptyOrBlank(callout_args$x)) {
```
zmoore-usgs, line 54: Should this still be checking `callout_args$x[i]` since it's within the for loop?

R/utils-limits.R
```diff
@@ -47,14 +46,12 @@ testCallouts <- function(plot_obj, xlimits){
 #' 
 testCalloutsByView <- function(plot_obj, view_num, xlimits_real, width_char, xrange){
   
-  i_view <- which(names(plot_obj$view_num) == "callouts")
+  i_view <- which(names(plot_obj[[view_num]]) == "callouts")
```
zmoore-usgs, line 49: I think `view_num` needs to be wrapped in quotes here when not using the `$` operator, otherwise it treats it as a variable. I think the `$` adds implicit quotes. I'm not totally sure about that though. @lindsaycarr?

PR #536, opened 2017-01-13
Get approvals refactor

thongsav-usgs:


R/utils-fetch.R
```diff
@@ -82,6 +82,15 @@ fetchTimeSeries <- function(reportObject, seriesName){
   return(val)
 }
 
+#' Fetch time series approvals
+#' @description Given a report object, will pull time series with given name
+#' @param reportObject the full report data
+#' @param seriesName the time series name to fetch
+fetchTimeSeriesApprovals <- function(ts){
+  val <- ts$approvals
```
zmoore-usgs: The documentation on this function should probably be updated.

Also for consistency with other fetch functions it might be best to use the bracket (`[['approvals']]`) syntax.
Something like:

```
fetchTimeSeriesApprovals <- function(reportObject, seriesName){
    val <- reportObject[[seriesName]][['approvals']]
    return(val)
}
```

zmoore-usgs (changes requested):

zmoore-usgs (approved):

PR #537, opened 2017-01-17
utils-legend documentation and tests

mhines-usgs:


R/utils-legend.R
```diff
@@ -18,6 +17,16 @@ rmDuplicateLegendItems <- function(object){
...
+#' 
+#' @description Returns the desired time series label for the time series
+#' requested formatted for inclusion in the legend
+#' 
+#' @param ts The time series object which contains the time series information
+#' @param field The type of time series data requested
+#' 
+#' @return The type of time series requested as well as the units
+#' 
 getTimeSeriesLabel<- function(ts, field){
```
thongsav-usgs, line 30: I think TS is supposed to be reportObject? I don't think you need to update it now since the only report using it is UV hydrograph and I'm refactoring that now. Merging.

PR #539, opened 2017-01-17
Extremes clean up

thongsav-usgs:


R/extremes-data.R
```diff
@@ -1,13 +1,29 @@
...
+          ))
+}
+
 #' Create a Flat Text "extremes table" Type Output Table
 #' 
-#' @param rawData An extremes report JSON string.
+#' @param reportObject An extremes report JSON string.
 #' @importFrom dplyr mutate
 #' @return string table
 #' @export
```
zmoore-usgs, line 15: Does this still need to be exported?

zmoore-usgs (approved):
Changes look good to me, just the one small comment about whether or not extremesTable needs to be exported.

R/extremes-data.R
thongsav-usgs: Nope, Rmd/rmarkdown runs inside of the scope of the caller (inside repgen)

zmoore-usgs: Gotcha, so should the `@export` tag be removed then?

thongsav-usgs: Bah, sorry, I thought you were asking if it should be included. Will take out on next pr.

PR #541, opened 2017-01-19
UV/5yr separated dynamic gsplot logic from static styles

thongsav-usgs:


NAMESPACE
```diff
@@ -4,7 +4,6 @@ export(correctionsataglance)
 export(createDataRows)
```
mhines-usgs: prob don't need this exported either? (from extremes-data)

R/fiveyeargwsum-render.R
```diff
@@ -74,3 +74,32 @@ createfiveyeargwsumPlot <- function(data){
...
+  args <- list(...)
+  
+  styles <- switch(names(reportObject), 
+      stat = list(
+          lines = append(list(x=x, y=y, legend.name=legend.name), styles$stat_lines)
+          ),
+      est_stat = list(
+          lines = append(list(x=x, y=y, legend.name=legend.name), styles$est_stat_lines)
+          ),
+      max_iv = list(
```
lindsayplatt, line 93: maybe add a comment here as to why legend.name can't be included in the reportObject? Just so our future selves know

R/fiveyeargwsum-styles.R
```diff
@@ -1,17 +1,11 @@
...
-    styles <- switch(names(data), 
-                     stat = list(lines = list(x=x, y=y, type="o", col="black", pch=20, cex=0.5, lwd=0.8, legend.name=legend.name)),
-                     est_stat = list(lines = list(x=x, y=y, col="red", lty=2, legend.name=legend.name)),
-                     max_iv = list(points = list(x=x, y=y, col="red", pch=8, cex=2, legend.name=paste(args$maxLabel, info$type, ":", y))),
-                     min_iv = list(points = list(x=x, y=y, col="blue", pch=8, cex=2, legend.name=paste(args$minLabel, info$type, ":", y))),
-                     gw_level = list(points = list(x=x,y=y, pch = 8, bg = 'orange', col = 'orange', cex = 1.2, lwd=1, legend.name="Measured Water Level (GWSI)"))
-                     )
-    
-    return(styles)
+getFiveyearStyle <- function() {
```
lindsayplatt, line 1: I really like this idea of separating them out :+1:

R/fiveyeargwsum-styles.R
```diff
@@ -1,17 +1,11 @@
...
+  styles <- list(
+      stat_lines = list(type="o", col="black", pch=20, cex=0.5, lwd=0.8),
+      est_stat_lines = list(col="red", lty=2),
+      max_iv_points = list(col="red", pch=8, cex=2),
+      min_iv_points = list(col="blue", pch=8, cex=2),
+      gw_level_points = list(pch = 8, bg = 'orange', col = 'orange', cex = 1.2, lwd=1, legend.name="Measured Water Level (GWSI)")
```
lindsayplatt, line 7: should `legend.name` go in the switch statement instead? I know it's hardcoded, but it is more closely tied to the data than plotting styles.

R/uvhydrograph-render.R
```diff
@@ -290,26 +288,17 @@ createSecondaryPlot <- function(data, month, useDownsampled=FALSE){
 
 #' Add UV hydrograph objects to a plot object.
 #' @param object A gsplot, plot object.
-#' @param data Time series data to add to hydrograph.
-#' @param info Meta-data about time series data in "data" parameter.
-#' @param plotName Plot name string. Presently an element of
-#'                 domain {primary,secondary}.
-#' @param dataSides Sides data frame. See getUvStyle().
-#' @param dataLimits Limits data frame. See getUvStyle().
+#' @param list of gsplot calls to make
```
lindsayplatt: this needs to describe the arguments in the form `@param [arg name] [arg description]` . I think you are missing the argument name here.

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
   
   return(y.endpoint)
 }
+
+
+#' TODO
```
lindsayplatt: I think TODO is referring to this, but you need all of the function documentation

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
...
+  corrArrowPositions <- getCorrectionArrowPositions(x, y, correctionLabels)
+  corrAblinePositions <- getCorrectionAbLines(reportObject)
+  
+  primary_lbl <- info$primary_lbl
+  reference_lbl <- info$reference_lbl
+  comp_lbl <- paste("Comparison", info$comp_UV_TS_lbl, "@", info$comp_UV_lbl)
+  
+  compAxes <- TRUE
+  compAnnotations <- TRUE
+  compLabel <- primary_lbl
```
lindsayplatt, line 370: what is the difference between comp_lbl and compLabel? Can we have some comments specifying?

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
...
+  
+  if(dataSides$comparison == 6){
```
lindsayplatt, line 372: can you add some comments about what's going on here?

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
...
+          ),
+      csg_readings = list(
+          points=append(list(x=x, y=y), styles$csg_readings_points), 
+          error_bar=append(list(x=x, y=y, y.low=reportObject$csg_readings$uncertainty, y.high=reportObject$csg_readings$uncertainty), styles$csg_readings_error_bars)
+          ),
+      hwm_readings = list(
+          points=append(list(x=x, y=y), styles$hwm_readings_points), 
+          error_bar=append(list(x=x, y=y, y.low=reportObject$hwm_readings$uncertainty, y.high=reportObject$hwm_readings$uncertainty), styles$hwm_readings_error_bars)
+          ),
+      appr_approved_dv = list(
```
lindsayplatt, line 426: why is pch variable and not static?

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
...
+          ),
+      appr_working_dv = list(
+          points = append(list(x=x, y=y, pch=reportObject[[1]]$point_type, legend.name=legend.name), styles$appr_working_dv_points)
+          )
+  )
+  
+  return(plotConfig)
+}
+
+#' TODO
```
lindsayplatt: again, assuming TODO = add function documentation :)

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
...
+getCorrectionArrowPositions <- function(time, value, correctionLabels) {
+  x <- time
+  y <- value
+  
+  corrArrowPositions <- list()
+  
+  #Make the correction label lines connect to the outside of the bounding box and not to the center of the label
+  if(!isEmptyOrBlank(correctionLabels)){
+    # work around irrelevant warnings from devtools::check()
```
lindsayplatt, line 508: you can also use `select_()` instead of `select()` to get around those

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
...
+  
+  return(plotConfig)
+}
+
+#' TODO
+getCorrectionArrowPositions <- function(time, value, correctionLabels) {
+  x <- time
+  y <- value
+  
+  corrArrowPositions <- list()
```
lindsayplatt, line 504: I would remove this and put it in an `else` statement after the L513 stuff. Putting it here makes it look like you need to initialize the variable before being able to do anything else when I think it just needs to return an empty list if `correctionLabels` are empty.

R/uvhydrograph-render.R
```diff
@@ -363,3 +352,180 @@ YEndpoint <- function (corr.value.sequence, uncorr.value.sequence) {
...
+        mutate(x = ifelse(x > xorigin, x - 60 * 60 * 2.85 * correctionLabels$r, x + 60 * 60 * 2.85 * correctionLabels$r)) %>% 
+        as.list()
+  }
+  
+  return(corrArrowPositions)
+}
+
+#' TODO
+getCorrectionAbLines <- function(reportObject) {
+  corrAblinePositions <- list()
```
lindsayplatt, line 525: same here - put this in an `else` statement below L527

R/uvhydrograph-styles.R
```diff
@@ -1,102 +1,55 @@
...
-                                    text=list(x=correctionLabels$x, y=correctionLabels$y, labels=correctionLabels$label, srt=0, cex=0.6, side=7, axes=FALSE, pos=1, offset = -0.12, col=rgb(0,0,255,240,maxColorValue=255))),
-
-                gage_height = list(points=list(x=x, y=y, pch=21, bg='black', col='black', cex=.8, lwd=1, legend.name="Measurement gage height"),
-                                   callouts=list(x=x, y=y, labels=data$gage_height$n)),
-                gw_level = list(points = list(x=x,y=y, pch = 8, bg = 'orange', col = 'orange', cex = 1.2, lwd=1, legend.name="Measured Water Level (GWSI)")), 
-                meas_shift = list(points=list(x=x, y=y, pch=21, bg='green', col='green', cex=1, lwd=1, side=4, legend.name="Effective shift and error limits"),
-                                  error_bar=list(x=x, y=y, y.low=(y-data$meas_shift$minShift), y.high=(data$meas_shift$maxShift-y), col='green', lwd=.7, side=4)))
-  } 
+getUvStyles <- function() {
```
lindsayplatt, line 4: I really like this - it's much cleaner!

lindsayplatt (dismissed):
A couple of high-level thoughts:
1. Can you finish documentation before this is merged?
2. Don't forget that we are trying to move away from using $ to using [[ ]] where possible.

R/uvhydrograph-render.R
zmoore-usgs: I wrote this part back when adding the third Y-Axis. This is the logic for specifying how to label each side. If any TS are plotted on the same side they should have the same label and not plot their own axis marks, but if they're on different sides they should use different labels and show their axis marks. So this is checking what side each TS is being plotted on and applying the proper label. I can add some comments to clarify the logic in a separate PR after this is merged.

R/uvhydrograph-render.R
zmoore-usgs: See my comment below.

zmoore-usgs (commented):
I don't have any additional feedback outside of what Lindsay posted. I think this pattern makes sense, so I'll implement for DV Hydro as well. Once this is merged I'll also add comments to my axis labeling logic.

thongsav-usgs:
@lindsaycarr will start moving from [[]] notation, which is now possible since dropping the downsampling stuff. I am intentionally saving documentation for last as the methods are still changing quite a bit. Will incorporate the low hanging fruit and get this merged sooner rather than later because some there is some utils-read changes pending that we should get merged for SRS/SVP work being done by @kmschoep-usgs.

R/uvhydrograph-render.R
thongsav-usgs: It's actually considered good practice to always initialize variables to their desired default values. Reasoning being if the logic which comes after does not hit all edge cases you still get the default (eg: if there are nested if/else statements where someone leaves out an edge case, we still get the default value).

R/fiveyeargwsum-styles.R
thongsav-usgs: I was thinking of it as more styling than data. This value has changed a couple of times and it's been based on user preference.

PR #542, opened 2017-01-19
changing nullMask to also handle list

mhines-usgs:
and this will allow us to get rid of getComments in utils-html


mhines-usgs:
Kathy gives :+1:

PR #545, opened 2017-01-19
AQCU-991: DV Hydro and Five Year GW Sum Refactor - WIP

zmoore-usgs:
# Changes in shared code:

**utils-fetch:**

Added: _fetchMinMaxIVs()_

**utils-read:**

Added: _readMinMaxIVs()_

**utils-dataFlags:**

Modified: _isLogged()_ - Modified it to work with time series returned from readTimeSeries and readEstimatedTimeSeries


R/dvhydrograph-render.R
```diff
@@ -8,10 +8,11 @@ createDvhydrographPlot <- function(data) {
   options(scipen=8)
   
   dvData <- parseDVData(data)
```
thongsav-usgs, line 10: FYI, I'm on a similar path as you and maybe going a step further to break out the uber "dvData" and "dvInfo" objects. Rather than calling one uber parse function, going to try to break the parse function up into smaller parse functions and collect those objects separately to pass to the rendering.  I haven't started yet so I don't know how small it's going to be, but going to look for some natural logical divisions. You might want to continue down that path as well.

thongsav-usgs: I'm going to hold off on this for the day but would like to discuss with @zmoore-usgs and @lindsaycarr to decide how far we want to go with it.

PR #548, opened 2017-01-23
Vdiagram refactor

dpattermann-usgs:
Goes along with Java side v diagram refactor, otherwise will break.

Java refactor: https://github.com/USGS-CIDA/AQCU/pull/644


R/vdiagram-data.R
```diff
@@ -1,33 +1,45 @@
...
+#' @param reportObject An R object with the raw data required for a V Diagram
+#' @return A list containing the relevant data for a V Diagram
+#'
 parseVDiagramData <- function(reportObject){
   ratingShifts <- fetchRatingShifts(reportObject)
   
-  shiftPoints <- ratingShifts$shiftPoints
+  ### Which of these are required? None of them stop if they're not valid currently.
+  ### Alternatively, do we want to assign validParam(blah, blah) to each (i.e. shiftPoints <- validParam(etc,etc))
+  ### I assume we do as otherwise the "validParam" statements do nothing.
```
dpattermann-usgs, line 10: @lindsaycarr @thongsav-usgs  Could you guys look at this comment and the following validParam statements below? Should we be checking validParams each time or should we be assuming they're required?

lindsayplatt: We should ask Laura to double check this list (using JSON list names). I only did the last three.

```
ratingsShifts >> shiftPoints
ratingsShifts >> stagePoints
ratingsShifts >> shiftNumber
measurements >> errorMaxShiftInFeet 
measurements >> errorMinShiftInFeet
measurements >> shiftInFeet
measurements >> shiftNumber
measurements >> meanGageHeight
measurements >> measurementNumber
measurements >> historic (NOT REQUIRED)
maximumStageHeight (REQUIRED)
minimumStageHeight (REQUIRED)
```

R/vdiagram-data.R
```diff
@@ -55,6 +67,10 @@ parseVDiagramData <- function(reportObject){
     minStage=minStage))
 }
 
+#' History Measurements Label
+#' @description A function that checks to see if there are any historic measurements and, if so, creates a label
```
lindsayplatt, line 76: when looking in vdiagram-styles, it seems like the historic points are black, but the lines are blue.

R/vdiagram-data.R
```diff
@@ -63,21 +79,15 @@ historyMeasurementsLabel <- function(reportObject) {
   }
 }
 
+#' Set default historic flags
```
lindsayplatt, line 87: what's the difference between this and the result of `historyMeasurementsLabel`?

R/vdiagram-data.R
```diff
@@ -63,21 +79,15 @@ historyMeasurementsLabel <- function(reportObject) {
...
-  x <- NULL
-  if (length(histFlag)==1 && histFlag == " "){
-    histFlag <- rep(TRUE, length(x))
-  }
-
-  if (isEmptyOrBlank(histFlag)){
+  
+  if (isEmptyOrBlank(histFlag) || (length(histFlag)==1 && histFlag == " ")){
     histFlag <- FALSE
   }
```
lindsayplatt, line 95: why does a length of 1 warrant FALSE?

seems weird for the `histFlag` to be anything but a logical. Why would it have a value?

R/vdiagram-data.R
lindsayplatt: Maybe this instead of 2 lines?
`shiftPoints <- validParam(ratingShifts[["shiftPoints"]], "shiftPoints")`

R/vdiagram-render.R
```diff
@@ -181,20 +135,20 @@ addRatingShifts <- function(vplot, vdiagramData, styles) {
 vdiagramTable <- function(reportObject){
   ratingShifts <- fetchRatingShifts(reportObject)
   
-  shiftPoints <- ratingShifts$shiftPoints
+  shiftPoints <- ratingShifts[["shiftPoints"]]
```
lindsayplatt: Some of this is repeated in `parseVDiagramData`, but I don't see an easy way to share functionality. I'd suggest this be two different functions:

1. `fetchVDiagramTable <- function(ratingShifts)` which returns a list of data, where `names(data)` is `shiftPoints, stagePoints, shiftID, startTime, rating, nShift`
2. `renderVDiagramTable <- function(shiftDataList)`

R/vdiagram-data.R
dpattermann-usgs: Agree on asking Laura to say which are necessary for the plot. I personally like the two line option because it makes it more evident that it's a two step thing, but don't have strong opinion about it.

R/vdiagram-data.R
dpattermann-usgs: Can you give an example of this tomorrow to me?

R/vdiagram-data.R
dpattermann-usgs: I think this is to set the flag for if it's historic or not, while the other piece just creates the label. They're two different pieces of the metadata it looks like?

R/vdiagram-data.R
dpattermann-usgs: Yeah, that was a weird piece of logic. It did that, then changed it to false. I assume it's to make " " become FALSE because empty strings would look bad? The length==1 seems to be to check that it isn't null?

R/vdiagram-render.R
dpattermann-usgs: Yeah, I thought the same thing, but didn't think of that solution. Good idea, will implement the fetcher and the actual rendering bit.

R/vdiagram-data.R
lindsayplatt: I haven't visually noticed, just looked in the code.

R/vdiagram-data.R
lindsayplatt: Should we add `" "` to isEmptyOrBlank?

dpattermann-usgs: I'm not against it. We should ask other team members, but sure.

dpattermann-usgs:
@lindsaycarr Do you think this is okay to merge?

R/vdiagram-data.R
```diff
@@ -19,6 +19,12 @@ parseVDiagramData <- function(reportObject){
   shiftId <- ratingShifts[["shiftNumber"]]
```
lindsayplatt: wouldn't you want to use the fetch functions you just created here, too?

R/utils-fetch.R
```diff
@@ -141,4 +141,58 @@ fetchMinMaxIVs <- function(reportObject, stat){
...
   val <- reportObject$readings
   return(val)
+}
+
+#' Fetch the shift points
+#'
+#' @description Given a report object, will pull the shift points
+#' @param reportObject the full report data
+fetchShiftPoints <- function(reportObject){
+  val <- reportObject[["shiftPoints"]]
```
lindsayplatt: shouldn't the body of these functions be:

```r
ratingShifts <- fetchRatingShifts(reportObject)
val <- ratingShifts[["shiftPoints"]]
```

I think it's misleading to have this named `reportObject` since that refers to the un-parsed list that was converted from JSON and still has everything

lindsayplatt:
@dpattermann-usgs I added a few more comments, but after those are addressed this should be good to go!

R/utils-fetch.R
dpattermann-usgs: @zmoore-usgs

lindsayplatt:
@zmoore-usgs can you approve + merge based on the last commit changes? You're obviously more familiar w/ the fetch-read pattern since you're currently implementing it.

zmoore-usgs (approved):

PR #550, opened 2017-01-23
Begin breaking down primary/secondary data lists, starting with corrections.

thongsav-usgs:
Primary benefit is we no longer rely on series_corrX naming convention in several ambiguous places. We are, however, moving away from the logical data grouping of primary and secondary graphs at the data level. This might be OK sense that was why we started relying on implicit naming conventions to begin with.


R/utils-plot.R
```diff
@@ -133,4 +133,25 @@ formatTimeSeriesForPlotting <- function(series, removeZeroNegativeFlag=NULL){
...
+}
+
+
+#' Calculate Lims
+#' For a data frame of points, will calculate a lims object. X and Y field names can be configured for the points.
+#' @param pts data frame of points
+#' @param xMinField name of the field which will contain the min x value (default "time")
+#' @param xMaxField name of the field which will contain the max x value (default "value")
+#' @param yMinField name of the field which will contain the min y value (default "time")
+#' @param yMaxField name of the field which will contain the max y value (default "value")
```
lindsayplatt, line 145: your documentation is a little off here - L142-143 should have "time", then the following two should have "value"

R/uvhydrograph-data.R
```diff
@@ -1,12 +1,24 @@
+#' Get months
+#' @description For a report, return a list of all months with data (sorted) in format 'YYMM'.
+#' @param reportObject a UV Hydrograph report
```
lindsayplatt, line 3: it's not the report itself, it's a list of the raw JSON data, right?

R/uvhydrograph-data.R
```diff
@@ -1,12 +1,24 @@
...
+#' @return list of months, sorted, as 'YYMM'
```
lindsayplatt, line 4: returns a vector of months, not a list

R/uvhydrograph-data.R
```diff
@@ -1,12 +1,24 @@
...
 #'@importFrom lubridate parse_date_time
-
 getMonths <- function(reportObject){
   corr <- getTimeSeries(reportObject, "primarySeries")
   uncorr <- getTimeSeries(reportObject, "primarySeriesRaw")
   months <- unique(c(corr$month, uncorr$month))
   return(sort(months))
 }
 
+readCorrectionsByMonth <- function(reportObject, fieldName, month) {
```
lindsayplatt, line 13: :+1: 

needs documentation

R/uvhydrograph-data.R
```diff
@@ -267,13 +264,14 @@ parseSecondarySupplementalInfo <- function(reportObject, pts) {
   return(supplemental)
 }
 
-correctionsTable <- function(data) {
-  if (any(names(data) %in% c("series_corr", "series_corr_ref", "series_corr_up", "series_corr2"))) {
-    corrections <- data[[grep("series_corr", names(data))]]
+correctionsAsTable <- function(corrections) {
+  if(!is.null(corrections) && nrow(corrections) > 0) {
```
lindsayplatt, line 300: I believe `!isEmptyVar(corrections)` will work here

R/utils-plot.R
```diff
@@ -133,4 +133,25 @@ formatTimeSeriesForPlotting <- function(series, removeZeroNegativeFlag=NULL){
   }
   
   return(series)
+}
+
+
+#' Calculate Lims
```
mhines-usgs, line 139: Does this make more sense to go in utils-limits?

R/uvhydrograph-data.R
```diff
@@ -267,13 +264,14 @@ parseSecondarySupplementalInfo <- function(reportObject, pts) {
...
     corrections_table <- as.data.frame(cbind(seq(nrow(corrections)), as.character(corrections$time), corrections$comment))
     colnames(corrections_table) <- c("", "Time", "Comments")
     return(corrections_table)
-  } else (return(corrections_table <- NULL))
+  } else {
+    return(corrections_table <- NULL)
```
lindsayplatt, line 305: I would change this so that there is only one `return` statement. Delete `return(corrections_table)` and replace `return(corrections_table <- NULL)` with just `corrections_table <- NULL`. Then have one return statement, `return(corrections_table)`. This is more common R syntax (I know that's how it was before - but it seems weird to have `<-` inside of the return function, so we should change while we are updating now anyways).

R/uvhydrograph-data.R
```diff
@@ -267,13 +264,14 @@ parseSecondarySupplementalInfo <- function(reportObject, pts) {
   return(supplemental)
 }
 
-correctionsTable <- function(data) {
-  if (any(names(data) %in% c("series_corr", "series_corr_ref", "series_corr_up", "series_corr2"))) {
-    corrections <- data[[grep("series_corr", names(data))]]
+correctionsAsTable <- function(corrections) {
```
lindsayplatt, line 299: much better without that vector of data names!

R/uvhydrograph-data.R
```diff
@@ -337,83 +336,64 @@ yposGroupValue <- function(data, prev, r, build_vec, vars) {
...
+#' TODO
 #' @importFrom dplyr row_number
 #' @importFrom dplyr desc
-parseLabelSpacing <- function(data, limits) {
-  
-  if (names(data) %in% c("series_corr", "series_corr_ref", "series_corr_up", "series_corr2")){
-    #Number of seconds to offset labels by for display
-    secondOffset <- 4 * 60 * 60
+parseCorrectionsLabelSpacing <- function(corrections, limits) {
+  #Number of seconds to offset labels by for display
```
lindsayplatt, line 375: should we have `if(!isEmptyVar(corrections))` in this function?

R/uvhydrograph-render.R
```diff
@@ -168,20 +174,22 @@ createPrimaryPlot <- function(primaryData, primaryInfo){
...
     plot_object <-
         AddToGsplot(plot_object, getPrimaryPlotConfig(primaryData[i], primaryInfo$plotStartDate, primaryInfo$plotEndDate, 
                 primaryInfo$primary_lbl, primaryInfo$reference_lbl, paste("Comparison", primaryInfo$comp_UV_TS_lbl, "@", primaryInfo$comp_UV_lbl),
-                sides, ylims, primaryInfo[[grep("lims_UV", names(primaryInfo))]]))
+                sides, ylims, lims))
   }
   
+  # corrections have been pulled out of primaryData and are their own top level object. Need to get it's own style info
+  # and add to plot. NOTE: this is out of original order for now.
+  plot_object <- AddToGsplot(plot_object, getCorrectionsPlotConfig(corrections, primaryInfo$plotStartDate, primaryInfo$plotEndDate,
```
lindsayplatt, line 196: I like this function for the `do.call` stuff. But agree that explicit calls (such as what CORR does) would be better.

R/uvhydrograph-render.R
```diff
@@ -437,6 +436,40 @@ getPrimaryPlotConfig <- function(primaryPlotData, plotStartDate, plotEndDate, pr
...
+#' @param corrections list of data objects relavant to plotting corrections data
+#' @param plotStartDate start date of this plot 
+#' @param plotEndDate end date of this plot
+#' @param label label of that these corrections are associated with
+#' @param limits list of lims for all of the data which will be on here
+#' @importFrom grDevices rgb
+getCorrectionsPlotConfig <- function(corrections, plotStartDate, plotEndDate, label, 
+    limits) {
+  
+  if(nrow(corrections) <= 0) {
```
lindsayplatt, line 482: Just a single return statement would be preferred for consistency - this would have `plotConfig <- list()`. I also think you can use `if(isEmptyVar(corrections))`

lindsayplatt:
@thongsav-usgs I like the direction this is going - put some comments about individual functions or lines. Overall I think this is a good pattern to follow.

PR #551, opened 2017-01-23
Aqcu 990 - SVP refactor

kmschoep-usgs:


mhines-usgs:
@kmschoep-usgs  a reminder that you can/should delete sitevisitpeak-render if you haven't already!

tests/testthat/test-sitevisitpeak.R
```diff
@@ -4,17 +4,70 @@ wd <- getwd()
 setwd(dir = tempdir())
 
 context("testing sitevisitpeak")
+
 test_that("sitevisitpeak examples work",{
   library(jsonlite)
   
   data <- fromJSON(system.file('extdata','sitevisitpeak','sitevisitpeak-example.json', package = 'repgen'))
-  expect_is(sitevisitpeak(data,'html', 'Author Name'), 'character')
+  expect_is(sitevisitpeak(data, 'html', 'Author Name'), 'character')
```
mhines-usgs: I think you can remove 'html' now from these

thongsav-usgs: Yes, you can. This is actually passing "html" as the author name and the author name is ignored.

thongsav-usgs (approved):

tests/testthat/test-sitevisitpeak.R
kmschoep-usgs: OK, I'll do this, then someone with write access can merge.

PR #552, opened 2017-01-23
srs improvements

mhines-usgs:
moving two general functions from formatSRS into their own functions


mhines-usgs:
This isn't ready to merge yet, but getting close. Need to address problems in the summary qualifiers table showing up after using the generic function and also need to redo existing test to use generic functions

inst/sensorreading/sensorreading.Rmd
```diff
@@ -63,10 +63,14 @@ if(!identical(tbl, "The dataset requested is empty.")){
...
 if(!identical(tbl, "The dataset requested is empty.")){
-  tbl2 <- srsQualifiersTable(data, tbl)
-  if(!is.null(tbl2) && nrow(tbl2) > 0){
+  quals <- as.data.frame(data$readings$qualifiers)
+  if(!is.null(quals) && length(quals) > 0){ 
+    quals <- quals[, c("code", "identifier", "displayName")]
+    tbl2 <- formatQualifiersTable(quals)
     if(!is.null(tbl2) && nrow(tbl2) > 0){
-      kable(tbl2, align='l', row.names=FALSE, caption = "Qualifiers")
+      if(!is.null(tbl2) && nrow(tbl2) > 0){
```
zmoore-usgs: Is this a duplicate of the line above?

R/sensorreading-data.R
```diff
@@ -264,17 +235,17 @@ getAppliedCorrection <- function(raw, corrected) {
...
+#' @param nearestcorrectedValue The nearest corrected value to the reading value
 #'
 #' @param uncertainty The uncertainty specified for the reading value
 #' 
 #' @return Yes or No if the corrections applied are within the reference
 #' values for their specified uncertainty
 #' 
-getCorrectedRef <- function (value, nearestCorrectedValue, uncertainty) {
-  if ((!isEmpty(value)) && (!isEmpty(uncertainty)) && (!isEmpty(nearestCorrectedValue))) {
+getCorrectedRef <- function (value, nearestcorrectedValue, uncertainty) {
```
zmoore-usgs, line 245: Just a little styling question: Should nearestcorrectedValue be camel cased to nearestCorrectedValue?

R/utils-html.R
```diff
@@ -193,7 +193,7 @@ getComments <- function(comments) {
 #' @param inQualifiers data frame of filtered (for SVP) or all (for SRS) qualifiers.
 #' @return list of deduplicated qualifiers with column names.
 formatQualifiersTable <- function(inQualifiers) {
-    if (isEmptyOrBlank(inQualifiers) || nrow(inQualifiers) == 0) return()
+    if(isEmptyOrBlank(inQualifiers) || (isEmptyVar(inQualifiers))) return()
```
zmoore-usgs: One thing that Lindsay suggested on my stuff was to limit functions to only having a single return. I guess maybe that should be the case here? Maybe wrap the rest of the function in a negated version of this if statement?

R/utils-html.R
```diff
@@ -210,10 +210,10 @@ formatQualifiersTable <- function(inQualifiers) {
 #' @param inQualifiers data frame of filtered (for SVP) or all (for SRS) qualifiers.
 #' @return comma-delimited string of qualifier codes
 formatQualifiersStringList <- function(inQualifiers) {
-  if(length(inQualifiers[[1]]) < 1) return("");
+  if(isEmptyOrBlank(inQualifiers) || (isEmptyVar(inQualifiers))) return("");
```
zmoore-usgs: Same as above.

zmoore-usgs (commented):
Overall changes look good to me! Just a couple of small comments

R/sensorreading-data.R
mhines-usgs: I actually had it that way earlier, and then I realized in the JSON its nearestcorrectedValue

but I could be swayed? :)

inst/sensorreading/sensorreading.Rmd
mhines-usgs: I guess so, I left it here from before, but I can try removing it and see if it matters or not. I find it hard to work with the Rmd files because I haven't figured out how really to debug within them

R/sensorreading-data.R
zmoore-usgs: Personally I'd say that the JSON should be corrected to be properly camel cased and that the variable should be the same way. But it doesn't really matter all that much. @thongsav-usgs ?

mhines-usgs: I'll make a future task for this, I don't think it's urgent and it would be good to do them together when we don't have a hotfix floating around? @thongsav-usgs

inst/sensorreading/sensorreading.Rmd
mhines-usgs: Actually this might have been introduced by me from a merge conflict, oops. Fixed.

R/utils-html.R
mhines-usgs: yes, I see that this is pretty close to being what we need how you like so I'll get it to work with one return. Thanks!

mhines-usgs:
re: dataframes not returning true for is.null when they have no rows: http://stackoverflow.com/questions/28556658/why-does-an-empty-dataframe-fail-an-is-null-test

R/utils-html.R
```diff
@@ -193,15 +193,16 @@ getComments <- function(comments) {
 #' @param inQualifiers data frame of filtered (for SVP) or all (for SRS) qualifiers.
 #' @return list of deduplicated qualifiers with column names.
 formatQualifiersTable <- function(inQualifiers) {
-    if(isEmptyOrBlank(inQualifiers) || (isEmptyVar(inQualifiers))) return()
   
-        columnNames <- c("Code",
-                         "Identifier",
-                         "Description")
+  toRet <- data.frame()
```
mhines-usgs, line 197: create an empty data frame so we can return empty if the if block is skipped

inst/sitevisitpeak/sitevisitpeak.Rmd
```diff
@@ -71,10 +71,8 @@ if(!identical(tbl, "The dataset requested is empty.")){
 ```{r,echo=FALSE,result='asis',comment=NA,warning=FALSE}
 if(!identical(tbl, "The dataset requested is empty.")){
   tbl2 <- formatQualifiersTable(readAllFieldVisitQualifiers(readFieldVisitReadings(data)))
-  if(!is.null(tbl2) && nrow(tbl2) > 0){
-    if(!is.null(tbl2) && nrow(tbl2) > 0){
-      kable(tbl2, align='l', row.names=FALSE, caption = "Qualifiers")
-    }
+  if(!isEmptyOrBlank(tbl2)){
```
mhines-usgs, line 74: suffering from the same redundant if block here so modified it

lindsayplatt:
@mhines-usgs that's why isEmptyOrBlank also checks nrow

tests/testthat/test-utils-html.R
```diff
@@ -135,7 +135,7 @@ test_that('formatQualifiersTable handles no qualifiers', {
 }')
   qualsDf <- repgen:::readQualifiers(inQualifiers,NULL)
   qualList <- repgen:::formatQualifiersTable(qualsDf)
-  expect_true(is.null(qualList))
+  expect_true(nrow(qualList)==0)
```
mhines-usgs, line 138: had to change this test to check for a dataframe with 0 rows instead of null since dataframes can't be tested for null http://stackoverflow.com/questions/28556658/why-does-an-empty-dataframe-fail-an-is-null-test

R/sensorreading-data.R
mhines-usgs: https://internal.cida.usgs.gov/jira/browse/AQCU-1015 for the JSON change and changes to sensorreading-data.R file

PR #553, opened 2017-01-23
AQCU-991: DV Hydrograph refactor continued.

zmoore-usgs:


R/utils-fetch.R
```diff
@@ -126,12 +126,9 @@ fetchCorrections <- function(reportObject, seriesCorrName){
 #' @param reportObject the full report data
 #' @param stat 'MIN' or 'MAX' data to select
 fetchMinMaxIVs <- function(reportObject, stat){
-  if(!is.null(stat)){
```
dpattermann-usgs, line 129: Do we not need this null check because we check if it's null somewhere else? I worry that we'll pass in NULL and try and use a function on it.

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
+#' @description Given the full report JSON object, reads the field
+#' visit measurements and handles read errors.
+#' @param reportObject the full report JSON object
+parseDVFieldVisitMeasurements <- function(reportObject){
   meas_Q <- tryCatch({
     readFieldVisitMeasurementsQPoints(reportObject)
   }, error = function(e) {
     warning(paste("Returning empty data frame as DV Hydro field visit measurements. Error:", e))
-    na.omit(data.frame(time=as.POSIXct(NA), value=as.numeric(NA), minQ=as.numeric(NA), maxQ=as.numeric(NA), n=as.numeric(NA), month=as.character(NA), stringsAsFactors=FALSE))
+    NULL
```
dpattermann-usgs: Is this just formatted weird on Github? NULL is just chilling on it's own line.

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
-  returnSeries <- tryCatch({
-    if(estimated){
-      readEstimatedTimeSeries(reportObject, series, timezone, descriptionField=description, isDV=TRUE)
-    } else {
-      readNonEstimatedTimeSeries(reportObject, series, timezone, descriptionField=description, isDV=TRUE)
-    }
-  }, error=function(e) {
-    warning(paste("Returning empty list as DV Hydro Time Series. Error:", e))
-    list()
+    NULL
```
dpattermann-usgs: Same for this NULL.

R/dvhydrograph-data.R
zmoore-usgs: That is an implied return because it its the last line in the function. I can make it more explicit so that it's more clear though.

R/utils-fetch.R
zmoore-usgs: I removed it because the other fetch functions don't do NULL checking either. The NULL checking is done from the read function. Might be worth discussing with everyone if fetch functions should do NULL checking or not though.

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
+#'
+#' @description Given a full report object and series name,
+#' reads and formats the estimated and non-estaimted time
+#' series for plotting.
+#' @param reportObject the full report JSON object
+#' @param seriesField the JSON field name for the TS data 
+#' @param descriptionField the JSON field name for the TS legend name
+#' @param timezone The timezone to parse the TS points into
+#' @param excludeZeroNegativeFlag whether or not to exclude zero and negative values
+parseDVTimeSeries <- function(reportObject, seriesField, descriptionField, timezone, excludeZeroNegativeFlag, estimated=FALSE){
```
thongsav-usgs, line 13: Can you add a fully fleshed out @return field? Kind of self explanatory but we've included it elsewhere.

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
-  refTimeSeries <- getDVHydroTimeSeries(reportObject, series, description, timezone)
-  refTimeSeriesEst <- list()
+#' @description Given the full report JSON object, reads the
+#' min and max IVs and formats them properly for plotting
+#' @param reportObject the full report JSON object
+#' @param timezone the time zone to parse points into
+#' @param type the type of TS that these points belong to
+#' @param invertedFlag whether or not the axis for the TS is inverted
+#' @param excludeMinMaxFlag wheter or not min / max IVs should be plotted or labeled
+#' @param excludeZeroNegativeFlag whether or not zero/negative values are included
```
thongsav-usgs, line 89: @return label would be useful here, for example.

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
-      dvConfig <- getDVHydrographPlotConfig(dvData[i], dvInfo)
-      for (j in names(dvConfig)) {
-        dvConfig[[j]] <- extendStep(dvConfig[[j]])
-        plot_object <- do.call(names(dvConfig[j]), append(list(object = plot_object), dvConfig[[j]]))
-      }
+  plot_object <-
+    XAxisLabelStyle(plot_object, startDate, endDate, timezone, plotDates)
+
+  # for non-approval-bar objects
+  for (i in grep("^appr_", names(dvData), invert = TRUE)) {
```
thongsav-usgs, line 91: I haven't gotten this far in my UV Hydro refactor but anticipated coming up to the same thing. Do you think it's worth while not to collect all of the data in a dvData list? This loop, for example, would be replaced by individual calls for each approval data element.

It's more code, but I get nervous doing string matches against env items to find things. There is a lot of potential for unseen side effects (eg, someone adds a env var latter that matches unintentionally).

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
+    if(grepl("TimeSeries", names(dvData[i]))){
```
thongsav-usgs, line 92: Same concern here, worthwhile to make a rule to always remove grep/grepl on env names? @lindsaycarr, thoughts?

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
   }
-  else {
-    plot_object <- NULL
-  }
-}
 
-createRefPlot <- function(data, series, descriptions) {
+  # approval bar styles are applied last, because it makes it easier to align
+  # them with the top of the x-axis line
+  plot_object <- ApplyApprovalBarStyles(plot_object, dvData)
```
thongsav-usgs, line 105: May want to refactor ApplyApprovalBarStyles to take more explicit params than the uber dvData object.

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
-      abline(v=seq(from=startDate, to=endDate, by="days"), lty=3, col="gray", where='first') %>%
-      abline(v=seq(from=startDate, to=endDate, by="weeks"), col="darkgray", lwd=1, where='first')
-    
-    # patch up top extent of y-axis
-    plot_object <- RescaleYTop(plot_object)
-    
-    return(plot_object)
   }
+  
+  plot_object <- ApplyApprovalBarStyles(plot_object, refData)
```
thongsav-usgs, line 215: And here...

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
-  legend.name <- reportObject[[1]]$legend.name
+  x <- plotItem[[1]]$time
+  y <- plotItem[[1]]$value
+  legend.name <- plotItem[[1]]$legend.name
   args <- list(...)
   
-  styles <- switch(names(reportObject), 
-    stat1Timeseries = list(
-      lines = append(list(x=x, y=y, legend.name=legend.name), styles$stat1_lines)
+  styles <- switch(names(plotItem),
```
thongsav-usgs, line 243: I think we should add an exception after this switch statement if nothing is found. Would cover the case where someone is asking for a plotItem that's not supported/expected.

I will do that for UV Hydrographs.

R/dvhydrograph-render.R
```diff
@@ -211,36 +303,67 @@ getDVHydrographPlotConfig <- function(reportObject, info = NULL, ...){
...
-  legend.name <- reportObject[[1]]$legend.name
+  x <- plotItem[[1]]$time
+  y <- plotItem[[1]]$value
+  legend.name <- plotItem[[1]]$legend.name
   args <- list(...)
 
-  styles <- switch(names(reportObject), 
-    secondaryRefTimeseries = list(
-      lines = append(list(x=x, y=y, legend.name=legend.name), sref_lines)
+  styles <- switch(names(plotItem),
```
thongsav-usgs, line 320: Same here, let's throw an exception for not found.

thongsav-usgs (commented):
I have some feedback but would be good merging now and doing feedback after the fact. Some of the shared functions will be helpful for the UV Hydrograph and 5 year refactoring.

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
+#' @param excludeZeroNegativeFlag whether or not to exclude zero and negative values
+parseDVTimeSeries <- function(reportObject, seriesField, descriptionField, timezone, excludeZeroNegativeFlag, estimated=FALSE){
+  timeSeries <- tryCatch({
+    if(estimated){
+      readEstimatedTimeSeries(reportObject, seriesField, timezone=timezone, descriptionField=descriptionField, isDV=TRUE)
+    } else {
+      readNonEstimatedTimeSeries(reportObject, seriesField, timezone=timezone, descriptionField=descriptionField, isDV=TRUE)
+    }
+  }, error=function(e) {
+    warning(paste("Returning NULL for DV Hydro Time Series: {", seriesField, "}. Error:", e))
```
lindsayplatt, line 21: looks like a good tryCatch!

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
+#' Parse DV Time Series
+#'
+#' @description Given a full report object and series name,
+#' reads and formats the estimated and non-estaimted time
+#' series for plotting.
+#' @param reportObject the full report JSON object
+#' @param seriesField the JSON field name for the TS data 
+#' @param descriptionField the JSON field name for the TS legend name
+#' @param timezone The timezone to parse the TS points into
+#' @param excludeZeroNegativeFlag whether or not to exclude zero and negative values
```
lindsayplatt, line 12: add `estimated` as a param here

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
-  #Approvals
-  approvals <- readApprovalBar(approvalSeries, timezone, legend_nm=approvalSeries[['legend.name']], snapToDayBoundaries=TRUE)
+#' Parse DV Approvals
+#'
+#' @description Given a time series, read and format the approvals
+#' data from it into apprpoval bars to put on the plot.
+#' @param timeSeries the time series to get approvals for
+#' @param timezone the timezone to parse the approval times into
+parseDVApprovals <- function(timeSeries, timezone){
+  return(readApprovalBar(timeSeries, timezone, legend_nm=timeSeries[['legend.name']], snapToDayBoundaries=TRUE))
```
lindsayplatt, line 39: separate `return` from function called, i.e. move readApprovalBar to a separate line and save to an object, then return that object. It's easier for debugging later.

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
     readFieldVisitMeasurementsQPoints(reportObject)
   }, error = function(e) {
     warning(paste("Returning empty data frame as DV Hydro field visit measurements. Error:", e))
-    na.omit(data.frame(time=as.POSIXct(NA), value=as.numeric(NA), minQ=as.numeric(NA), maxQ=as.numeric(NA), n=as.numeric(NA), month=as.character(NA), stringsAsFactors=FALSE))
+    return(NULL)
   })
-  
-  #Ground Water Levels
+
+  if(!anyDataExist(meas_Q)){
```
lindsayplatt, line 55: in parseDVTimeSeries, you had to look specifically at the point fields in anyDataExist -  `!anyDataExist(timeSeries[['points']])`. Is that not applicable here? Just want to be sure

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
-      readNonEstimatedTimeSeries(reportObject, series, timezone, descriptionField=description, isDV=TRUE)
-    }
-  }, error=function(e) {
-    warning(paste("Returning empty list as DV Hydro Time Series. Error:", e))
-    list()
+    return(NULL)
   })
 
-  return(returnSeries)
+  if(!anyDataExist(gw_level)){
```
lindsayplatt, line 74: same `[['points']]` question

R/dvhydrograph-data.R
```diff
@@ -1,251 +1,138 @@
...
-  timezone <- fetchReportMetadataField(reportObject, 'timezone')
-
-  #Time Series Data
-  refTimeSeries <- getDVHydroTimeSeries(reportObject, series, description, timezone)
-  refTimeSeriesEst <- list()
+#' @description Given the full report JSON object, reads the
+#' min and max IVs and formats them properly for plotting
+#' @param reportObject the full report JSON object
+#' @param timezone the time zone to parse points into
+#' @param type the type of TS that these points belong to
```
lindsayplatt, line 86: list what options are available for `type`

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
+  dvData[['stat1TimeSeriesEst']] <- parseDVTimeSeries(reportObject, 'firstDownChain', 'downChainDescriptions1', timezone, excludeZeroNegativeFlag, estimated=TRUE)
+  dvData[['stat2TimeSeries']] <- parseDVTimeSeries(reportObject, 'secondDownChain', 'downChainDescriptions2', timezone, excludeZeroNegativeFlag)
+  dvData[['stat2TimeSeriesEst']] <- parseDVTimeSeries(reportObject, 'secondDownChain', 'downChainDescriptions2', timezone, excludeZeroNegativeFlag, estimated=TRUE)
+  dvData[['stat3TimeSeries']] <- parseDVTimeSeries(reportObject, 'thirdDownChain', 'downChainDescriptions3', timezone, excludeZeroNegativeFlag)
+  dvData[['stat3TimeSeriesEst']] <- parseDVTimeSeries(reportObject, 'thirdDownChain', 'downChainDescriptions3', timezone, excludeZeroNegativeFlag, estimated=TRUE)
+  dvData[['comparisonTimeSeries']] <- parseDVTimeSeries(reportObject, 'comparisonSeries', 'comparisonSeriesDescriptions', timezone, excludeZeroNegativeFlag)
+  dvData[['comparisonTimeSeriesEst']] <- parseDVTimeSeries(reportObject, 'comparisonSeries', 'comparisonSeriesDescriptions', timezone, excludeZeroNegativeFlag, estimated=TRUE)
+  primaryPresentTS <- names(dvData)[1]
```
lindsayplatt, line 48: add comment here - not sure I like the name using "Present". What does that mean?

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
+
+  if(!is.null(dvData[['comparisonTimeSeries']]) && !is.null(dvData[['comparisonTimeSeriesEst']])){
+    dvData[['comparisonEdges']] <- getEstimatedEdges(dvData[['comparisonTimeSeries']][['points']], dvData[['comparisonTimeSeriesEst']][['points']])
+  }
+
+  #Get Additional Plot Data
+  dvData[['gw_level']] <- parseDVGroundWaterLevels(reportObject)
+  dvData[['meas_Q']] <- parseDVFieldVisitMeasurements(reportObject)
+  dvData <- append(dvData, parseDVMinMaxIVs(reportObject, timezone, dvData[[primaryPresentTS]][['type']], invertedFlag, excludeMinMaxFlag, excludeZeroNegativeFlag))
+  dvData <- append(dvData, parseDVApprovals(dvData[[primaryPresentTS]], timezone))
```
lindsayplatt, line 76: why do we switch to using `append` rather than just saying `dvData[['data.name']] <- ....`?

R/dvhydrograph-render.R
lindsayplatt: I believe we discussed this somewhere else. I think it's better if we get rid of the string matching pieces and move to explicit calls.

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
-      tzf <- sub("([[:digit:]]{2,2})$", ":\\1", tzf)
-      formatted_label <- paste0(dvData[[ml]]$legend.name, data$firstDownChain$units, 
-                                format(as.POSIXct(dvData[[ml]]$time), " %b %d, %Y %H:%M:%S"), " (UTC ", tzf, ")")
-      
-      plot_object <- mtext(plot_object, formatted_label, side = 3, axes=FALSE, cex=0.85, line = line, adj = 0)
-      
-      line <- line + 1
+    dvConfig <- getDVHydrographPlotConfig(dvData[i], yLabel=yLabel)
+    for (j in names(dvConfig)) {
+      dvConfig[[j]] <- extendStep(dvConfig[[j]])
```
lindsayplatt, line 98: add comment - what is this doing?

R/dvhydrograph-render.R
lindsayplatt: ^ agree, but maybe that's a diff PR and task

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
-    startDate <- flexibleTimeParse(data$reportMetadata$startDate, timezone=data$reportMetadata$timezone)
-    endDate <- toEndOfDay(flexibleTimeParse(data$reportMetadata$endDate, timezone=data$reportMetadata$timezone))
+  plot_object <- rmDuplicateLegendItems(plot_object)
+  
+  # custom gridlines below approval bar
+  plot_object <- plot_object %>% 
+    abline(v=seq(from=toStartOfDay(startDate), to=toStartOfDay(endDate), by="days"), lty=3, col="gray", where='first') %>%
+    abline(v=seq(from=toStartOfDay(startDate), to=toStartOfDay(endDate), by="weeks"), col="darkgray", lwd=1, where='first')
+  
+  # patch up top extent of y-axis
```
lindsayplatt, line 114: more explicit comment  - what does "patch up" mean? If I looked at this two months from now, I don't think I'd know the point of this.

R/dvhydrograph-render.R
```diff
@@ -3,199 +3,291 @@ dvhydrographPlot <- function(data) {
...
+  options(scipen=8)
+
+  #Get Necessary Report Metadata
+  ref_name_capital <- switch(series,
+    'secondaryReferenceTimeSeries' = 'Secondary',
+    'tertiaryReferenceTimeSeries' = 'Tertiary',
+    'quaternaryReferenceTimeSeries' = 'Quaternary'
+  )
+  seriesEst <- paste0(series, 'Est', setp="")
+  seriesEstEdges <- paste0(series, 'EstEdges', setp="")
```
lindsayplatt, line 162: not sure what `setp=""` is. For paste(), you can add `sep=""` to say that you don't want any spaces separating the strings *OR* you can use paste0(), whose default for `sep` is no space.

R/dvhydrograph-render.R
lindsayplatt: yeah, that's just having something at the end without a name:
```r
nm <- "third"
switch(nm,
       first="stuff",
       second="more stuff",
       "my default")
[1] "my default"
```

lindsayplatt: Or you could throw an error or warning:

```r
nm <- "third"
switch(nm,
       first="stuff",
       second="more stuff",
       stop("not found"))
Error: not found

switch(nm,
       first="stuff",
       second="more stuff",
       warning("not found"))
Warning message:
not found 
```

R/utils-read.R
```diff
@@ -681,6 +620,19 @@ readEstimatedTimeSeries <- function(reportObject, seriesName, timezone, descript
...
+#'
+#' @description Reads and formats a time series from the provided full report object
+#' @param reportObject the full JSON report object
+#' @param timezone the timezone to parse times to
+#' @param seriesName the name of the time series to extract
+#' @param shiftTimeToNoon [DEFAULT: FALSE] whether or not to shift DV times to noon
+#' @param isDV whether or not the specified time series is a daily value time series
+#' @param requiredFields optional overriding of required fields for a time series
+readNonEstimatedTimeSeries <- function(reportObject, seriesName, timezone, descriptionField=NULL, shiftTimeToNoon=FALSE, isDV=FALSE, requiredFields=NULL) {
+  return(readEstimatedTimeSeries(reportObject, seriesName, timezone, descriptionField, shiftTimeToNoon, isDV, requiredFields, inverted=TRUE))
```
lindsayplatt, line 633: readEstimatedTimeSeries and return on different lines

lindsayplatt (commented):
I'd love to see the fetch vs read vs parse functions defined somewhere (like in a Google doc or possibly the README). When we need to add more, it would be nice to have this to reference so we understand the philosophy behind each.

R/dvhydrograph-render.R
zmoore-usgs: I don't use env to create the dvData list anymore, it's hand created up above. I do agree about removing the string matching though, it's on my todo list for additional refactoring steps after getting it running.

R/dvhydrograph-render.R
zmoore-usgs: Yeah, probably a different task. I'm not sure if anyone has taken a look at refactoring the approvals stuff yet?

R/dvhydrograph-data.R
zmoore-usgs: I don't think it's as applicable here because I believe the field visit measurements and ground water level arrays will never be filled if there are no points, while a time series field could be. Probably better to be safe and include it though, so I will. 👍

R/dvhydrograph-data.R
zmoore-usgs: See above

R/dvhydrograph-render.R
zmoore-usgs: I switched to using append because those functions are returning lists of several things instead of a single data frame. Might be cleaner to switch them to returning individual data frames, though that would require changes the the readApprovalBar function in utils-approvals.

zmoore-usgs: I could also store those as sub-lists of the dvData list using the same method as with the other pieces of information. I.E: it becomes dvData[['approvals']] <- parseDVApprovals() and then accessed by dvData[['approvals']][['appr_uv_working']] etc.

I'm not sure which of those options would be the best or if it matters?

R/dvhydrograph-render.R
zmoore-usgs: Whoops, those are both typos of `sep` and leftover before I learned about paste0 haha.

R/dvhydrograph-render.R
lindsayplatt: Hmmm sounds like the approvals stuff needs to be reworked anyways, but I think consistency here might be better. So I like your second suggestion. @thongsav-usgs what do you think?

PR #554, opened 2017-01-23
New & improved R package installer for repgen.

ashalper-usgs:


inst/extdata/dev-packages.R
```diff
@@ -0,0 +1,6 @@
+# Install repgen development packages. These should only be necessary if you 
+# intend to do repgen development. They are not needed on a production server.
+
+source(paste0(getwd(), "/inst/extdata/installPackages.R"))
+
+installPackages(c("devtools", "roxygen2", "testthat"))
```
lindsayplatt: :+1:

inst/extdata/packages.R
```diff
@@ -0,0 +1,94 @@
...
+  "lazyeval",
+  "knitr",
+  "lubridate",
+  "rmarkdown",
+  "yaml"
+)
+
+# Known production tiers, listed at
+# https://docs.google.com/document/d/1vBOTUPtIdeCTGe7Or2cQGMMOFS_hD545yZ1E6YBGizs/edit
+tiers <-
```
mhines-usgs: I don't know that we shouldn't have the info here in the script but why couldn't this come from Jenkins?

inst/extdata/installRDependencies.sh
```diff
@@ -1,26 +0,0 @@
...
-Rscript -e 'remove.packages("lubridate")' || true
-Rscript -e 'remove.packages("htmlTable")' || true
-
-echo '***** installing upward dependencies'
-
-Rscript -e 'install.packages(c("bitops","Rcpp","digest","magrittr","stringi","stringr","yaml","evaluate","formatr","highr","R6","assertthat","lazyeval","DBI","jsonlite","brew","mime","curl","caTools","openssl","memoise","whisker","rstudioapi","git2r","withr","markdown","knitr","htmltools","rmarkdown","RCurl","htmlTable","roxygen2","dplyr","lubridate"), repos="http://mran.microsoft.com/snapshot/2016-03-31", type="source")'
-
-Rscript -e 'install.packages("httr", repos = "http://mran.microsoft.com/snapshot/2016-01-27",type="source")'
-Rscript -e 'install.packages("devtools", repos = "http://mran.microsoft.com/snapshot/2015-01-19",type="source")'
-Rscript -e 'Sys.setenv(TZ="UTC")'
```
mhines-usgs, line 26: Is this no longer needed?

inst/extdata/packages.R
```diff
@@ -0,0 +1,94 @@
...
+if (nodename %in% tiers) {
+  source("installPackages.R")
+} else {
+  # presume it's a development machine with the source checked out
+  source(paste0(getwd(), "/inst/extdata/installPackages.R"))
+}
+
+# all packages except devtools and its prerequisites are held back to older
+# versions, keyed by their repo URLs below
+installPackages(pkgs, "http://mran.microsoft.com/snapshot/2016-03-31")
```
mhines-usgs: So this also handles all the required packages upstream dependencies?

inst/extdata/packages.R
thongsav-usgs: Agreed, we should not be checking in infrastructure information into project code. The application should not know or react to infrastructure it's on.

thongsav-usgs: If the desire is to have this parameterized, it could be a simple "includeDevPackages" boolean, and the caller would decide what to pass in. Versus the script code deciding based on where it's sitting.

inst/extdata/installRDependencies.sh
ashalper-usgs: Correct. `packages.R`/`installPackages.R` supersedes it.

inst/extdata/packages.R
ashalper-usgs: Yes. `install.packages()` does that by default.

inst/extdata/packages.R
ashalper-usgs: @thongsav-usgs how would you have the code remove `BH`, `devtools`, `httr` from a production system? These are development packages that are only required to install `gsplot`. After `gsplot` is installed, they become cruft.

thongsav-usgs: Have the caller of the script pass in the boolean saying to remove dev packages (either pass in a param or set an env variable), the application itself should not have tier specific logic.

inst/extdata/installRDependencies.sh
mhines-usgs: I'm specifically asking about the timezone line here

-Rscript -e 'Sys.setenv(TZ="UTC")'

inst/extdata/packages.R
mhines-usgs: For some reason I remember explicitly calling the requirements in order but can't remember why that mattered, hopefully it no longer does.

inst/extdata/installRDependencies.sh
ashalper-usgs: owe, good point. I wasn't sure about that. Should that go in `packages.R`?

inst/extdata/packages.R
ashalper-usgs: OK, will do.

inst/extdata/installRDependencies.sh
mhines-usgs: sure, i don't think it much matters as long as it happens as a part of the install/set up.  OR if you're doing any other environment-specific bits, maybe it makes sense to do this there? like where you set the R_LIBS? @thongsav-usgs is there a specific place this needs to be called at? I know we have done it 'last' but not sure it needs to be that way.

inst/extdata/packages.R
ashalper-usgs: I've stared at the diffs between DEV and my workstation quite a lot now. I suspect R's `install.packages()` is smarter now than when the original package installer was written.

Also it's crossed my mind that since we're in the process of doing this rather major re-factor, it might make sense to bring the package snapshot set up to the present date. I've noticed that the package dependency tree for the development packages at least, seems to have changed since then. I would expect the same for the production packages too.

mhines-usgs: I think it's fine, as long as we are able to verify moving the mran date to a later date works as expected. But I think it makes sense to wait until we're done with the refactor and then change that date? I'd make a future task for it, I think.

ashalper-usgs: If it were me (and it's mostly not), and I was being all Agile, I'd want the upstream prerequisites established earlier and oftener; so that I knew if there were any API changes in the newer package versions, I could cope with those earlier in repgen.

mhines-usgs: I think we want to update but not in the middle of the refactor, is my point. I think others agreed in slack when we had this convo.

ashalper-usgs: As you like. Mostly not me.

mhines-usgs: I made this skeleton task that we should pull in when we're done with the refactor, feel free to add details or suggestions!

https://internal.cida.usgs.gov/jira/browse/AQCU-1009

ashalper-usgs:
Please don't merge this yet. Found a problem.

mhines-usgs:
@ashalper-usgs you asked me to review this again but there isn't anything new.

ashalper-usgs:
Kind of don't know what is going on. I closed this PR, and opened a new one: https://github.com/USGS-R/repgen/pull/555

PR #555, opened 2017-01-25
New new & improved R package installer.

ashalper-usgs:


mhines-usgs:
Thanks, I see this now.

ashalper-usgs:
phew! :)

inst/extdata/installPackages.R
```diff
@@ -0,0 +1,129 @@
...
+    file = cert_bundle,
+    sep = "\n",
+    append = TRUE
+  )
+  file.append(cert_bundle, doiRootCA)
+  file.remove(doiRootCA)
+}
+
+# https://github.com/USGS-R/gsplot#installation
+devtools::install_github("USGS-R/gsplot", quiet = TRUE)
```
mhines-usgs: I think we need to parameter-ize this like the current jenkins job does this

inst/extdata/installPackages.R
```diff
@@ -0,0 +1,129 @@
...
+installPackages(pkgs, lib, "http://mran.microsoft.com/snapshot/2016-03-31")
+installPackages("httr", lib, "http://mran.microsoft.com/snapshot/2016-01-27")
+
+# To be able to install gsplot from GitHub using devtools (below), download
+# DOI root certificate and append to openssl package's certificate bundle
+# (ruthlessly, without asking). See also
+# https://usgs-cida.slack.com/archives/r-activities/p1469472383000332
+
+# locate & read the openssl certificate bundle
+cert_bundle <- httr:::find_cert_bundle()
```
mhines-usgs: this might be unnecessary on the server but I am not entirely sure. the doi cert _should_ already be there, and I doubt the R user or Tomcat has the power to edit it if it found it this way?

inst/extdata/installPackages.R
```diff
@@ -0,0 +1,129 @@
...
+      print(w)
+    },
+    error = function(e) {
+      return()
+    })
+  }
+} else {
+  installPackages(c("roxygen2", "testthat"), lib)
+  
+  cat(
```
mhines-usgs: this isn't necessary on the servers either (the part about windows)

mhines-usgs (dismissed):

inst/extdata/installPackages.R
ashalper-usgs: This was only necessary for devtools to be able to use curl to download gsplot from GitHub. At the time I wasn't thinking that gsplot was being installed directly by Jenkins. I'll move it to the developer configuration.

inst/extdata/installPackages.R
```diff
@@ -0,0 +1,62 @@
...
+}
+
+lib <- Sys.getenv("R_LIBS")
+if (nchar(lib) == 0) {
+  stop("Could not get a value for R_LIBS environment variable")
+}
+
+# all packages except devtools and its prerequisites are held back to older
+# versions
+installPackages(pkgs, lib, "http://mran.microsoft.com/snapshot/2016-03-31")
```
mhines-usgs, line 44: Edit: got it now. the function sources then sets this as the repo when it's called.

We probably should use https here? any reason not to? or just avoiding the cert issues?

mhines-usgs:
i say, wipe out the RLIBS dir on dev and test this out with a new jenkins job?

we can always re-run the current job if this bombs

inst/extdata/installPackages.R
ashalper-usgs: No reason not to use HTTPS. I just copy-and-pasted the URL from the older installer code.

ashalper-usgs:
I made a tarball of the current DEV `R_libs` so it can be restored if necessary.

PR #556, opened 2017-01-26
AQCU-991: Continued DV Hydro refactor

zmoore-usgs:


zmoore-usgs:
The failing tests are caused by UV Hydrograph since I have the new gaps stuff now but UV Hydrograph is not using them yet. UV Hydrograph tests are the only ones failing for me locally. These should be fixed once @thongsav-usgs finishes his work on UV Hydro.

R/dvhydrograph-data.R
```diff
@@ -48,12 +50,13 @@ parseDVFieldVisitMeasurements <- function(reportObject){
   meas_Q <- tryCatch({
     readFieldVisitMeasurementsQPoints(reportObject)
   }, error = function(e) {
-    warning(paste("Returning empty data frame as DV Hydro field visit measurements. Error:", e))
+    warning(paste("Returning NULL as DV Hydro field visit measurements. Error:", e))
     return(NULL)
   })
 
-  if(!anyDataExist(meas_Q)){
+  if(!anyDataExist(meas_Q) || nrow(meas_Q) == 0){
```
lindsayplatt, line 57: I think `anyDataExist` checks `nrow(data) == 0` so this wouldn't be necessary.

R/dvhydrograph-data.R
```diff
@@ -67,12 +70,13 @@ parseDVGroundWaterLevels <- function(reportObject){
   gw_level <- tryCatch({
     readGroundWaterLevels(reportObject)
   }, error = function(e) {
-    warning(paste("Returning empty data frame as DV Hydro ground water levels. Error:", e))
+    warning(paste("Returning NULL as DV Hydro ground water levels. Error:", e))
     return(NULL)
   })
 
-  if(!anyDataExist(gw_level)){
+  if(!anyDataExist(gw_level) || nrow(gw_level) == 0){
```
lindsayplatt, line 77: Check what `anyDataExist` is doing. See `utils-validation.R`

R/dvhydrograph-data.R
```diff
@@ -104,20 +109,27 @@ parseDVMinMaxIVs <- function(reportObject, timezone, type, invertedFlag, exclude
 
   #Max Checking
   if( (!isEmptyOrBlank(excludeMinMaxFlag) && excludeMinMaxFlag) || 
-      (!isEmptyOrBlank(excludeZeroNegativeFlag) && excludeZeroNegativeFlag && !isEmptyOrBlank(max_iv[['value']]) && max_iv[['value']] <= 0)){
+      (!isEmptyOrBlank(excludeZeroNegativeFlag) && excludeZeroNegativeFlag && !isEmptyOrBlank(max_iv[['value']]) && as.numeric(max_iv[['value']]) <= 0)){
```
lindsayplatt, line 112: if we are coercing `max_iv[['value']]` into numeric within this statement, I imagine we would want to return it as numeric in `returnList`. So, maybe coercion to numeric should happen within `getMinMaxIV` ?

R/dvhydrograph-data.R
```diff
@@ -104,20 +109,27 @@ parseDVMinMaxIVs <- function(reportObject, timezone, type, invertedFlag, exclude
...
     returnList <- append(returnList, list(min_iv_label=min_iv))
   } else if(anyDataExist(min_iv[['value']])) {
     returnList <- append(returnList, list(min_iv=min_iv))
   }
 
+  #Check if the IVs allow for a log axis or not
+  returnList[['canLog']] <- TRUE
+  
+  if((!isEmptyOrBlank(returnList[['max_iv']][['value']]) && returnList[['max_iv']][['value']] <= 0) || (!isEmptyOrBlank(returnList[['min_iv']][['value']]) && returnList[['min_iv']][['value']] <= 0)){
+    returnList[['canLog']] <- FALSE
```
lindsayplatt, line 130: and the intent is that this does not consider the `excludeZeroNegative` flag, correct?

R/dvhydrograph-data.R
```diff
@@ -126,17 +138,28 @@ parseDVMinMaxIVs <- function(reportObject, timezone, type, invertedFlag, exclude
 #' vertical lines connecting the steps between those TS.
 #' @param stat the parsed non-estimated time series
 #' @param est the parsed estimated time series
+#' @param excludeZeroNegativeFlag whether or not zero / negative values
+#' will be removed
 #' @return a list of vertical lines connecting steps between stat and est
 #' @importFrom dplyr arrange
-getEstimatedEdges <- function(stat, est){
+#' @return a list of the edges with the folowing columns time, y0, y1,
```
lindsayplatt, line 145: it looks like you have two `@return` statements - I think you'll need to delete one.

lindsayplatt: describe "edges" here

R/dvhydrograph-data.R
```diff
@@ -126,17 +138,28 @@ parseDVMinMaxIVs <- function(reportObject, timezone, type, invertedFlag, exclude
...
+#' and newSet. Time is the x position of the edge, y0 is the first y
+#' value and y1 is the second. newSet is what the next time series is
+getEstimatedEdges <- function(stat, est, excludeZeroNegativeFlag=NULL){
   estEdges <- list()
 
-  if(isEmptyOrBlank(est[['value']]) || isEmptyOrBlank(stat[['value']])){
+  if(is.null(stat) || is.null(est) || isEmptyOrBlank(est[['points']][['value']]) || isEmptyOrBlank(stat[['points']][['value']])){
```
lindsayplatt, line 151: I don't think you need the first two `is.null` checks for est and stat. If they are NULL, `isEmptyOrBlank` will return TRUE as you'd expect.

```r
est <- NULL
repgen:::isEmptyOrBlank(est)
[1] TRUE
```

R/dvhydrograph-data.R
```diff
@@ -126,17 +138,28 @@ parseDVMinMaxIVs <- function(reportObject, timezone, type, invertedFlag, exclude
...
     return(NULL)
   }
-  
-  est <- est[c('time', 'value')]
-  stat <- stat[c('time', 'value')]
+
+  #Don't render edges for values that will be removed if we are exculding zero and negative values
+  if(!is.null(excludeZeroNegativeFlag) && excludeZeroNegativeFlag){
```
lindsayplatt, line 156: I think we want to get away from just using `is.null` and using `isEmptyOrBlank` to capture the multitude of things it might come through as.

R/dvhydrograph-data.R
```diff
@@ -126,17 +138,28 @@ parseDVMinMaxIVs <- function(reportObject, timezone, type, invertedFlag, exclude
...
+    stat[['points']] <- removeZeroNegative(stat[['points']])
+    est[['points']] <- removeZeroNegative(est[['points']])
+  }
+
+  est <- est[['points']][c('time', 'value')]
+  stat <- stat[['points']][c('time', 'value')]
```
lindsayplatt, line 162: what do these two lines do?

R/dvhydrograph-data.R
```diff
@@ -167,6 +190,7 @@ getEstimatedEdges <- function(stat, est){
 #' @param tsType the type of the TS (to use for the legend name)
 #' @param timezone the timezone to parse the times into
 #' @param inverted whether or not the TS is inverted
+#' @return the first min or max IV data point from the list of min max IVs
```
lindsayplatt, line 193: maybe add a comment about why only the first min or max are used

R/dvhydrograph-render.R
```diff
@@ -20,64 +15,65 @@ createDVHydrographPlot <- function(reportObject){
...
+  #Get Necessary Report Metadata
   if(validateFetchedData(metaData, "metadata", requiredMetadataFields)){
-    #Get Necessary Report Metadata
     timezone <- fetchReportMetadataField(reportObject, 'timezone')
     excludeZeroNegativeFlag <- fetchReportMetadataField(reportObject, 'excludeZeroNegative')
     excludeMinMaxFlag <- fetchReportMetadataField(reportObject, 'excludeMinMax')
     invertedFlag <- fetchReportMetadataField(reportObject, 'isInverted')
+    excludeZeroNegativeFlag <- ifelse(is.null(excludeZeroNegativeFlag), FALSE, TRUE)
+    excludeMinMaxFlag <- ifelse(is.null(excludeMinMaxFlag), FALSE, TRUE)
+    invertedFlag <- ifelse(is.null(invertedFlag), FALSE, TRUE)
```
lindsayplatt, line 32: should we use `isEmptyOrBlank` instead of `is.null`? Also, can these be moved so that are directly below their fetcher?

R/dvhydrograph-render.R
```diff
@@ -20,64 +15,65 @@ createDVHydrographPlot <- function(reportObject){
...
+  stat2TimeSeries <- parseDVTimeSeries(reportObject, 'secondDownChain', 'downChainDescriptions2', timezone, excludeZeroNegativeFlag)
+  stat2TimeSeriesEst <- parseDVTimeSeries(reportObject, 'secondDownChain', 'downChainDescriptions2', timezone, excludeZeroNegativeFlag, estimated=TRUE)
+  stat3TimeSeries <- parseDVTimeSeries(reportObject, 'thirdDownChain', 'downChainDescriptions3', timezone, excludeZeroNegativeFlag)
+  stat3TimeSeriesEst <- parseDVTimeSeries(reportObject, 'thirdDownChain', 'downChainDescriptions3', timezone, excludeZeroNegativeFlag, estimated=TRUE)
+  comparisonTimeSeries <- parseDVTimeSeries(reportObject, 'comparisonSeries', 'comparisonSeriesDescriptions', timezone, excludeZeroNegativeFlag)
+  comparisonTimeSeriesEst <- parseDVTimeSeries(reportObject, 'comparisonSeries', 'comparisonSeriesDescriptions', timezone, excludeZeroNegativeFlag, estimated=TRUE)
 
   #Validate Basic Plot Data
-  if(is.null(primaryPresentTS)){
+  if(is.null(c(stat1TimeSeries, stat1TimeSeriesEst, stat2TimeSeries, stat2TimeSeriesEst, stat3TimeSeries, stat3TimeSeriesEst))){
```
lindsayplatt: you probably need an `is.null(any(...))` and maybe `isEmptyOrBlank` ?! But I guess you know that these should only return NULL since that is how `parseDVTimeSeries` is setup.

R/dvhydrograph-render.R
```diff
@@ -20,64 +15,65 @@ createDVHydrographPlot <- function(reportObject){
...
+  
+  #Get Additional Plot Data
+  groundWaterLevels <- parseDVGroundWaterLevels(reportObject)
+  fieldVisitMeasurements <- parseDVFieldVisitMeasurements(reportObject)
+  minMaxIVs <- parseDVMinMaxIVs(reportObject, timezone, stat1TimeSeries[['type']], invertedFlag, excludeMinMaxFlag, excludeZeroNegativeFlag)
+  minMaxCanLog <- TRUE
+
+  if(!isEmptyOrBlank(minMaxIVs)){
+    minMaxLabels <- minMaxIVs[grepl("label", names(minMaxIVs))]
+    minMaxPoints <- minMaxIVs[!grepl("label", names(minMaxIVs))]
```
lindsayplatt, line 69: `minMaxLabels` and `minMaxPoints` need to be created outside of this if statement in case the statement is never entered

R/dvhydrograph-render.R
```diff
@@ -87,50 +83,58 @@ createDVHydrographPlot <- function(reportObject){
...
 
   #Legend
   legend_items <- plot_object$legend$legend.auto$legend
   ncol <- ifelse(length(legend_items) > 3, 2, 1)
-  leg_lines <- ifelse(ncol==2, ceiling((length(legend_items) - 6)/2), 0) 
-  legend_offset <- ifelse(ncol==2, 0.3+(0.05*leg_lines), 0.3)
-  plot_object <- legend(plot_object, location="below", cex=0.8, legend_offset=0.2, y.intersp=1.5, ncol=ncol)
+  leg_lines <- ifelse(ncol==2, ceiling((length(legend_items) - 6)/2), length(legend_items))
+  legend_offset <- ifelse(ncol==2, 0.09+(0.025*leg_lines), (0.025*leg_lines))
+  legend_offset <- ifelse(as.period(interval(startDate, endDate, tzone = attr(startDate, timezone))) < years(1), legend_offset+0.03, legend_offset+0.08)
```
lindsayplatt, line 130: add comment about why - since there will be year labels in addition to months or something

R/dvhydrograph-render.R
```diff
@@ -87,50 +83,58 @@ createDVHydrographPlot <- function(reportObject){
...
+  plot_object <- legend(plot_object, location="below", cex=0.8, legend_offset=legend_offset, y.intersp=1.5, ncol=ncol)
 
   #Add Min/Max labels if we aren't plotting min and max
-  minmax_labels <- append(dvData['max_iv_label'], dvData['min_iv_label'])
   line <- 0.33
-  for(ml in na.omit(names(minmax_labels))){
+  for(ml in na.omit(names(minMaxLabels))){
     #Extract Timezone
-    tzf <- format(as.POSIXct(dvData[[ml]]$time), "%z")
+    tzf <- format(as.POSIXct(minMaxLabels[[ml]]$time), "%z")
```
lindsayplatt: there is an object called `timezone` available to use or could we expect different mins and maxes to have different timezones?

R/dvhydrograph-render.R
```diff
@@ -87,50 +83,58 @@ createDVHydrographPlot <- function(reportObject){
...
     #Insert ":" before 2nd to last character
     tzf <- sub("([[:digit:]]{2,2})$", ":\\1", tzf)
-    formatted_label <- paste0(dvData[[ml]]$legend.name, data$firstDownChain$units, 
-                              format(as.POSIXct(dvData[[ml]]$time), " %b %d, %Y %H:%M:%S"), " (UTC ", tzf, ")")
+    formatted_label <- paste0(minMaxLabels[[ml]]$legend.name, stat1TimeSeries[['units']], 
+                              format(as.POSIXct(minMaxLabels[[ml]]$time), " %b %d, %Y %H:%M:%S"), " (UTC ", tzf, ")")
```
lindsayplatt: almost complete if removing `$`! same with timezone line

R/dvhydrograph-render.R
```diff
@@ -169,59 +173,61 @@ createDVHydrographRefPlot <- function(reportObject, series, descriptions) {
...
-  refData <- list()
-
-  refData[[series]] <- parseDVTimeSeries(reportObject, series, descriptions, timezone, excludeZeroNegativeFlag)
-  refData[[seriesEst]] <- parseDVTimeSeries(reportObject, series, descriptions, timezone, excludeZeroNegativeFlag, estimated=TRUE)
+  referenceSeries <- parseDVTimeSeries(reportObject, series, descriptions, timezone, excludeZeroNegativeFlag)
+  referenceSeriesEst <- parseDVTimeSeries(reportObject, series, descriptions, timezone, excludeZeroNegativeFlag, estimated=TRUE)
 
   #Validate Basic Plot Data
-  if(is.null(refData[[series]]) && is.null(refData[[seriesEst]])){
+  if(is.null(c(referenceSeries, referenceSeriesEst))){
```
lindsayplatt: I think you'll need an `is.null(any(...))` or `isEmptyOrBlank`

R/utils-plot.R
```diff
@@ -135,6 +136,90 @@ formatTimeSeriesForPlotting <- function(series, removeZeroNegativeFlag=NULL){
...
+    dataFrameList <- dataFrameList[unname(unlist(lapply(dataFrameList, function(e) {return(!is.null(e))})))]
+    return(dataFrameList)
+  }
+
+  return(NULL)
+}
+
+#' Plot Time Series
+#' 
+#' @description Function that takes a GS plot object and the necessary time series data and
```
lindsayplatt: GS plot = gsplot

R/utils-plot.R
```diff
@@ -135,6 +136,90 @@ formatTimeSeriesForPlotting <- function(series, removeZeroNegativeFlag=NULL){
...
+#' @param plot_object The GS Plot object to plot onto
+#' @param item The item to plot
+#' @param name The variable name to use for the item (used for style and config matching)
+#' @param configFunction The function to use for fetching the style and config data for this TS
+#' @param yLabel The label to put onto the Y-Axis for this item (default: "")
+#' @param isDV Whether or not the plot is a daily value plot (defulat: FALSE)
+plotItem <- function(plot_object, item, name, configFunction, yLabel="", isDV=FALSE){
+  if(!is.null(item) && anyDataExist(item)){
+    legendName <- item['legend.name']
+    plotItem <- configFunction(item, name, yLabel)
```
lindsayplatt, line 210: didn't know you could do this! I thought you'd have to to use `do.call(configFunction, ...)`

R/utils-validation.R
```diff
@@ -112,13 +112,11 @@ checkRequiredFields <- function(data, requiredFields){
...
       
-      naCols <- c(missingFields, partialFields)
-      
-      
+      naCols <- c(missingFields, partialFields)      
       return(naCols)
     }
 
-    return(list())
+    return(NULL)
```
lindsayplatt: try to only have on `return()`

R/utils-validation.R
```diff
@@ -151,8 +149,8 @@ validateFetchedData <- function(data, name, requiredFields, stopMissing=TRUE, st
   if(!isEmptyOrBlank(data) && !isEmptyOrBlank(requiredFields)){
     missingFields <- checkRequiredFields(data, requiredFields)
 
-    if(length(missingFields) > 0){
-      if(!stopPartial){
+    if(!is.null(missingFields) && length(missingFields) > 0){
```
lindsayplatt: `isEmptyOrBlank` could be substituted for both of these logical statements

lindsayplatt (commented):
nothing major, just some small suggestions

R/dvhydrograph-render.R
zmoore-usgs: These are addressed in my Five Year GW refactor. I added a parseReportMetadataField() function that can be supplied with a default value to return if the field is not found, so now instead of a fetch and then null check it just uses one parse line.

R/dvhydrograph-data.R
zmoore-usgs: I think they were necessary at one time because the points that this function was receiving had additional columns that I didn't want to include in the estimated edges data. I'll check and see if this is still the case, but I don't think it is so these can probably be removed.

R/dvhydrograph-render.R
zmoore-usgs: I'm not totally sure if we could expect them to have different timezones or not. The main thing is that Laura wanted this label to be in a specific UTC format and the timezone object is not in the same format and I wasn't sure how to convert it since it is just a string and not a date object.

lindsayplatt: Ah, I see. Then carry on :)

zmoore-usgs:
@lindsaycarr I made updates to everything except for stuff in dvhydrograph-data.R because that file has been significantly changed on my five year gw refactor branch. I will make the changes on that branch and they will be in the PR for that branch.

lindsayplatt (approved):

PR #557, opened 2017-01-31
UV refactor, secondary data broken up into smaller pieces.

thongsav-usgs:
Only partial refactor, the primary plot functions and data are still a WIP.


R/utils-read.R
```diff
@@ -334,8 +334,7 @@ readApprovalBar <- function(ts, timezone, legend_nm, snapToDayBoundaries=FALSE){
   if (!isEmptyOrBlank(ts$approvals$startTime) && !isEmptyOrBlank(ts$startTime)) {
     startTime <-
         flexibleTimeParse(ts$approvals$startTime, timezone = timezone)
-    chain.startTime <-
-        flexibleTimeParse(ts$startTime, timezone = timezone)
+    chain.startTime <- ts$startTime #start time must be preparsed, relies on readTimeSeries
```
lindsayplatt, line 337: trying to get away from using `$` and instead using `[[ ]]`

R/uvhydrograph-data.R
```diff
@@ -95,61 +98,74 @@ parsePrimaryUVData <- function(data, month) {
...
   } else {
-    #Upchain Time Series Data
-    approvals <- readApprovalBar(data[['upchainSeries']], timezone, 
-        legend_nm=getTimeSeriesLabel(data, "upchainSeries"))
+    uncorrected <- NULL
   }
   
-  effect_shift <- subsetByMonth(getTimeSeries(data, "effectiveShifts"), month)
-  gage_height <- subsetByMonth(readMeanGageHeights(data), month)
+  return(list(corrected=corrected, estimated=estimated, uncorrected=uncorrected, inverted=inverted))
```
lindsayplatt, line 120: I generally avoid having return() create the thing I'm returning. It's easier to debug if they are on separate lines

R/uvhydrograph-data.R
```diff
@@ -95,61 +98,74 @@ parsePrimaryUVData <- function(data, month) {
...
+readSecondaryUvHydroApprovalBars <- function(reportObject, timezone, month) {
+  if(hasReferenceSeries(reportObject) && !isPrimaryDischarge(reportObject)) {
+    #Reference Time Series Data
+    approvals <- readApprovalBar(readTimeSeries(reportObject, "referenceSeries", timezone), timezone, 
+        legend_nm=getTimeSeriesLabel(reportObject, "referenceSeries"))
+  } else if(hasUpchainSeries(reportObject)) {
+    #Upchain Time Series Data
+    approvals <- readApprovalBar(readTimeSeries(reportObject, "upchainSeries", timezone), timezone, 
+        legend_nm=getTimeSeriesLabel(reportObject, "upchainSeries"))
+  }
```
lindsayplatt, line 167: with two ifs, I think we need an `else` - so what is returned for `approvals` when neither of those ifs are TRUE

R/uvhydrograph-data.R
```diff
@@ -160,6 +176,22 @@ isPrimaryDischarge <- function(reportObject) {
   return(any(grepl("Discharge", fetchReportMetadataField(reportObject,'primaryParameter'))))
 }
 
+#' Has Refernce Series
+#' Determines if the report has a reference series attached
+#' @param reportObject UV Hydro report object
+#' @return true/false
+hasReferenceSeries <- function(reportObject) {
+  return(any(grepl("referenceSeries", names(reportObject))))
```
lindsayplatt, line 184: suggest moving the actions outside of the `return` (same in `hasUpchainSeries` below)

R/uvhydrograph-data.R
```diff
@@ -238,58 +270,33 @@ parseSupplementalPrimaryInfo <- function(data, pts, lims) {
...
+  return(list(label=label, units=units))
+}
+
+
+getSecondaryTimeSeriesUvInfo <- function(reportObject, timezone, month) {
+  if(hasReferenceSeries(reportObject) && !isPrimaryDischarge(reportObject)) {
+    infos <- getTimeSeriesUvInfo(reportObject, "referenceSeries")
+  } else if(hasUpchainSeries(reportObject)) {
+    infos <- getTimeSeriesUvInfo(reportObject, "upchainSeries")
+  }
```
lindsayplatt, line 298: I'd add an `else` just in case

R/uvhydrograph-render.R
```diff
@@ -219,78 +230,106 @@ createPrimaryPlot <- function(primaryData, primaryInfo, corrections, lims){
 #' TODO documentation
 #' @importFrom lubridate hours
 #' @importFrom lubridate minutes
-createSecondaryPlot <- function(secondaryData, secondaryInfo, secondarySeriesList, corrections, lims){
+createSecondaryPlot <- function(uvInfo, timeInformation, secondarySeriesList, 
+    approvalBars, effective_shift_pts, gw_level, 
+    meas_shift, gage_height,
+    corrections, lims, invertPlot, tertiary_label=""){
```
lindsayplatt, line 236: :+1:

R/uvhydrograph-render.R
```diff
@@ -219,78 +230,106 @@ createPrimaryPlot <- function(primaryData, primaryInfo, corrections, lims){
...
+                meas_shift$time, meas_shift$value,
+                startDate, endDate, 
+                uvInfo$label, tertiary_label, lims)
         )
   }
-
+  
+  if(!isEmptyVar(gw_level)){
+    plot_object <- 
+        AddToGsplot(plot_object,
```
lindsayplatt, line 319: can we update this function to not capitalize the first letter to match the rest of our style?

tests/testthat/test-utils-read.R
```diff
@@ -201,7 +201,7 @@ test_that('readApprovalPoints data returns as expected', {
 
 test_that('readApprovalBar data returns as expected', {
   library("jsonlite")
-  ts <- fromJSON('{
+  ts <- repgen:::readTimeSeries(fromJSON('{ "anExampleSeries" : {
```
lindsayplatt, line 204: I think you want to skip this and just manually create the things needed in `ts` for `readApprovalBar`? Because we aren't also trying to test `readTimeSeries` here

lindsayplatt (commented):
My comments that apply to a lot of things:

try to get `return()` statements only returning an object; anything that creates a list or performs some action should happen on a second line.

avoid `$` and use `[[ ]]` where possible!

PR #558, opened 2017-01-31
AQCU-992: Refactor Five Year GW Summary - Must be merged after AQCU-991

zmoore-usgs:
Note this PR also includes and is dependent on the changes from the PR for AQCU-991 (PR #556). This should not be merged until after that PR is merged.


R/dvhydrograph-render.R
```diff
@@ -249,7 +246,13 @@ getDVHydrographPlotConfig <- function(plotItem, plotItemName, yLabel="", ...){
 
   x <- plotItem[['time']]
   y <- plotItem[['value']]
-  legend.name <- plotItem[['legend.name']]
+
+  if('legend.name' %in% names(plotItem)){
+    legend.name <- plotItem$legend.name
+  } else {
+    legend.name <- ""
+  }
```
lindsayplatt: is there anyway to replace this with the `nullMask` function? 

`legend.name <- nullMask(plotItem$legend.name)`

R/dvhydrograph-render.R
```diff
@@ -278,30 +281,30 @@ getDVHydrographPlotConfig <- function(plotItem, plotItemName, yLabel="", ...){
...
-                           lty=ifelse(plotItem$newSet == "est", 1, 6), col=ifelse(plotItem$newSet == "est", "green", "red4")),
+      arrows = append(list(x0=plotItem[['time']], x1=plotItem[['time']], y0=plotItem[['y0']], y1=plotItem[['y1']], 
+                           lty=ifelse(plotItem[['newSet']] == "est", 1, 6), col=ifelse(plotItem[['newSet']] == "est", "green", "red4")),
                       styles$est_lines)
     ),
-    meas_Q = list(
+    fieldVisitMeasurements = list(
       points = append(list(x=x, y=y, legend.name="Measured Discharge"), styles$meas_q_points),
-      callouts = append(list(x=x, y=y, labels = plotItem$meas_Q$n), styles$meas_q_callouts)
+      callouts = append(list(x=x, y=y, labels = plotItem[['n']]), styles$meas_q_callouts)
```
lindsayplatt, line 300: just checking - we definitely don't need `plotItem[['meas_Q']][['n']]` right? Thought there might have been a find and replace error here

R/dvhydrograph-render.R
```diff
@@ -329,7 +332,13 @@ getDVHydrographRefPlotConfig <- function(plotItem, plotItemName, yLabel, ...){
...
   x <- plotItem[['time']]
   y <- plotItem[['value']]
-  legend.name <- plotItem[['legend.name']]
+  
+  if('legend.name' %in% names(plotItem)){
+    legend.name <- plotItem$legend.name
+  } else {
+    legend.name <- ""
+  }
+
```
lindsayplatt: use `nullMask` instead of if-else? Example given above

R/fiveyeargwsum-render.R
```diff
@@ -1,100 +1,151 @@
...
+  priorityStat <- getPriorityStat(reportObject)
+
+  #Get Basic Plot data
+  statTimeSeries <- parseTimeSeries(reportObject, priorityStat[['data_nm']], priorityStat[['descr_nm']], timezone, isDV=TRUE)
+  statTimeSeriesEst <- parseTimeSeries(reportObject, priorityStat[['data_nm']], priorityStat[['descr_nm']], timezone, estimated=TRUE, isDV=TRUE)
+
+  #Validate Basic Plot Data
+  if(is.null(c(statTimeSeries, statTimeSeriesEst))){
+    return(NULL)
+  }
```
lindsayplatt, line 58: I'm normally opposed to multiple return calls, but in this case :+1: because we'd want to skip all the other plotting stuff anyways. And early returns can be much cleaner than large if-else statements

R/fiveyeargwsum-render.R
```diff
@@ -1,100 +1,151 @@
...
+
+  #Add Min/Max labels if we aren't plotting min and max
+  line <- 0.33
+  for(ml in na.omit(names(minMaxLabels))){
+    #Extract Timezone
+    tzf <- format(as.POSIXct(minMaxLabels[[ml]][['time']]), "%z")
+    #Insert ":" before 2nd to last character
+    tzf <- sub("([[:digit:]]{2,2})$", ":\\1", tzf)
+    formatted_label <- paste0(minMaxLabels[[ml]][['legend.name']], statTimeSeries[['units']], 
+                              format(as.POSIXct(minMaxLabels[[ml]][['time']]), " %b %d, %Y %H:%M:%S"), " (UTC ", tzf, ")")
```
lindsayplatt: could we make all of this formatting label stuff into a function? Not really sure what file it would go in. Then we would just have:

```r
line <- 0.33
for(ml in na.omit(names(minMaxLabels))){
    formatted_label <- formatMinMaxLabel(ml, statTimeSeries[['units']])
    plot_object <- ....
    line <- line + 1
}
```

R/fiveyeargwsum-render.R
```diff
@@ -1,100 +1,151 @@
...
-  y <- reportObject[[1]]$value
-  legend.name <- reportObject[[1]]$legend.name
+  x <- plotItem$time
+  y <- plotItem$value
+
+  if('legend.name' %in% names(plotItem)){
+    legend.name <- plotItem$legend.name
+  } else {
+    legend.name <- ""
+  }
```
lindsayplatt: use `nullMask`?

lindsayplatt (changes requested):
Two main suggestions:
1. Replace occasions of "if something exists return the thing, else return an empty string" with nullMask.
2. Extract the min max label formatting into its own function?

R/dvhydrograph-render.R
zmoore-usgs: Just double checked and we do not need meas_Q any more, the list is now flattened when it comes in. Thanks for catching that though!

lindsayplatt (approved):

PR #560, opened 2017-02-01
Removed clean-up step due to Jenkins sequencing

ashalper-usgs:


ashalper-usgs:
@mhines-usgs merging this without review because it does less than before, and I'm in a hurry to test it.

PR #561, opened 2017-02-02
Added installPackages() line to install latest devtools.

ashalper-usgs:
Something that was being done apart from `installPackages.R` by the Jenkins job. Moved there for coherence.


mhines-usgs (commented):
seems OK to me.

PR #567, opened 2017-02-06
Package list derived from "Imports" section of DESCRIPTION file

ashalper-usgs:


mhines-usgs:
I'm not sure why we need the DOI cert? This is already installed on the servers by the SAs and we shouldn't necessarily have that here?

Otherwise I am interested to see how this works!

DESCRIPTION
```diff
@@ -2,7 +2,7 @@ Package: repgen
 Type: Package
 Title: USGS internal report generation
 Version: 1.1
-Date: 2016-11-22
+Date: 2017-02-06
 Author: Lindsay Carr, Andy Halper, Megan Hines, Zack Moore, Dominick Patterman, Jordan S Read, Phethala Thongsavanh
```
mhines-usgs: Since you're in here, would you mind adding Kathy Schoephoester to the authors?

inst/extdata/installPackages.R
```diff
@@ -21,7 +45,9 @@ installPackages <- function(pkgs, lib, repos = getOption("repos")) {
     },
     warning = function(w) {
       # if package p is not installed...
-      if (any(grepl("(DESCRIPTION file of package .+ is missing or broken|no package .+ was found)", w))) {
+      if (any(grepl(
```
mhines-usgs, line 48: does this handle the case where gsplot isn't in MRAN?

inst/extdata/installPackages.R
```diff
@@ -34,18 +60,59 @@ installPackages <- function(pkgs, lib, repos = getOption("repos")) {
   }
 }
 
-lib <- Sys.getenv("R_LIBS")
-if (nchar(lib) == 0) {
-  stop("Could not get a value for R_LIBS environment variable")
+nodename <- Sys.info()["nodename"]
```
mhines-usgs: not sure what this is for? can we divorce the local machine idea from what we're building on the server? this applies to the DOI root cert too.. I think you're mixing scope again.

inst/extdata/installPackages.R
```diff
@@ -34,18 +60,59 @@ installPackages <- function(pkgs, lib, repos = getOption("repos")) {
...
+pkgs <-
+  grep(
+    "gsplot", repgenImports(lib.loc = lib),
+    value = TRUE, fixed = TRUE, invert = TRUE
+  )
+
+# all packages are held back to older, MRAN versions
+installPackages(c(pkgs, "devtools"), lib, "https://mran.microsoft.com/snapshot/2016-03-31")
+
+if (nodename == "IGSWZTWWWSASHA") {
```
mhines-usgs: this isn't appropriate, since our scope right now is the server stack?

inst/extdata/installPackages.R
```diff
@@ -34,18 +60,59 @@ installPackages <- function(pkgs, lib, repos = getOption("repos")) {
...
+    file.remove(doiRootCA)
+  }
+}
 
-# devtools is only needed to install gsplot and repgen from source on GitHub;
-# it's not used by repgen in production
-installPackages(c("devtools"), lib, "https://cloud.r-project.org")
+# gsplot hasn't made it to an offical repository yet, so it needs to be
+# installed with devtools
+devtools::install_github("USGS-R/gsplot#420")
```
mhines-usgs, line 80: Once GRAN has the new version of gsplot (0.7.1, aka PR #420) we could install from there instead of with devtools, but this works anyway.

mhines-usgs (changes requested):

inst/extdata/installPackages.R
thongsav-usgs: yes, please remove infrastructure specific logic (as well as the root cert specific functionality)

inst/extdata/installPackages.R
ashalper-usgs: It's in there because it's a pain to develop this patch without it. I don't want an `R_LIBS` environment variable lurking around on my development machine (there was a time, in the not-so-distant past, when creating dependencies on environment variable values was considered poor form).

That said, I'll remove it.

DESCRIPTION
ashalper-usgs: yah sure.

inst/extdata/installPackages.R
ashalper-usgs: gsplot is installed on line 115, and I have a very tenuous understanding of gsplot's repositories. That line is really just copy-and-pasted from Zack's Slack recommendation.

inst/extdata/installPackages.R
ashalper-usgs: Right. But see comment above please.

inst/extdata/installPackages.R
ashalper-usgs: Let me know when it's ready.

inst/extdata/installPackages.R
mhines-usgs: gotcha. I understand your logic but I think, again, sorry to beat a horse, we just need to keep this focused on the server needs at this point

inst/extdata/installPackages.R
mhines-usgs: this will work for now, that can wait until they update gran. THANKS!!

mhines-usgs (approved):

PR #572, opened 2017-02-07
Removed installation of gsplot from R script to:

ashalper-usgs:
attempt to work around GitHub HTTPS throttling.


mhines-usgs (approved):
Looks good. We can wait for Travis then merge and test again.

PR #573, opened 2017-02-07
AQCU-1017: Added functionality for flipping min and max labels when…

zmoore-usgs:
…the parameter is on an inverted Y-Axis.


mhines-usgs:
Checking this out locally now

PR #580, opened 2017-02-11
Approval/DV pts refactored to return consolidated lists and added tests

thongsav-usgs:


R/utils-read.R
```diff
@@ -308,27 +308,35 @@ readApprovalPoints <- function(approvals, points, timezone, legend_nm, appr_var_
...
-        
-        approval_info[[i]] <- list(time = applicable_dates,
-            value = applicable_values,
-            legend.name = paste(sub_list["type"], legend_nm),
-            point_type = point_type)
-      }
+  for(level in date_index_list){
+    
+    appr_var <- appr_var_all[which(appr_type == level["type"])]
+    list <- level[[2]]
```
zmoore-usgs: Is there a way to replace this constant 2 with either a named constant somewhere or the actual name of the column (if it has a name)? Not a big deal, but could be confusing with just a 2.

R/utils-read.R
```diff
@@ -308,27 +308,35 @@ readApprovalPoints <- function(approvals, points, timezone, legend_nm, appr_var_
...
+    
+    size <- length(list)
```
zmoore-usgs: Is this used anywhere? It's certainly possible that I'm just missing it or that github isn't showing the line(s) it is used on.

zmoore-usgs (commented):
Just a couple of small comments, but nothing that would necessarily need to be changed before this is merged. Overall the changes look good to me.

R/utils-read.R
thongsav-usgs: Nope, left over from an old impl. Removing.

R/utils-read.R
thongsav-usgs: It's only related to line 309 so there isn't much outside of this function that would be able to change that constant. I may look at removing line 309 as well. I just modified the function but didn't look to see if that line should be changed.

R/uvhydrograph-render.R
```diff
@@ -506,49 +512,44 @@ calculateLimitsAndSides <- function(primarySeriesList, uvInfo, refInfo, compInfo
...
-          lines = append(list(x=x, y=y, legend.name=paste(styles[['est_UV_lbl']], primary_lbl)), styles[['est_UV_lines']])
+          lines = append(list(x=x, y=y, legend.name=paste(styles[['est_UV_lbl']], label)), styles[['est_UV_lines']])
           ),
       uncorrected = list(
-          lines = append(list(x=x, y=y, legend.name=paste(styles[['uncorr_UV_lbl']], primary_lbl)), styles[['uncorr_UV_lines']])
+          lines = append(list(x=x, y=y, legend.name=paste(styles[['uncorr_UV_lbl']], label)), styles[['uncorr_UV_lines']])
           ),
       comparison = list(
-          lines = append(list(x=x, y=y, ylim=dataLimits[['comparison']], side=dataSides[['comparison']], axes=compAxes, ylab=compLabel, ann=compAnnotations, legend.name=comp_lbl), styles[['comp_UV_lines']])
+          lines = append(list(x=x, y=y, ylim=ylim, side=dataSide, axes=compAxes, ylab=label, ann=compAnnotations, legend.name=label), styles[['comp_UV_lines']])
```
dpattermann-usgs, line 546: Did you mean to make all of these use the same label? Wouldn't this make the comparison series have the same legend name as the primary series?

R/utils-read.R
lindsayplatt: I would just suggest naming the second list item. `type` and `index` or something. Just makes it easier to follow along with

R/utils-read.R
```diff
@@ -308,27 +308,34 @@ readApprovalPoints <- function(approvals, points, timezone, legend_nm, appr_var_
   
   date_index_list <- list(list(type="Approved",approved_index), list(type="In Review",review_index), list(type="Working",working_index))
   
-  for(sub_list in date_index_list){
```
lindsayplatt: can we remove the semicolons from L301-303. They don't hurt, but they are also not being used in our repgen code anywhere else

R/utils-read.R
lindsayplatt: don't use `list` as your var name - that's the name of a function

R/utils-read.R
```diff
@@ -308,27 +308,34 @@ readApprovalPoints <- function(approvals, points, timezone, legend_nm, appr_var_
...
+    appr_var <- appr_var_all[which(appr_type == level["type"])]
+    list <- level[[2]]
+    
+    approval_info <- data.frame(time=.POSIXct(character()),
+      value=numeric(),
+      legend.name=character(),
+      point_type=numeric(),
+      stringsAsFactors=FALSE)
+
+    for(i in seq_along(list)){
```
lindsayplatt: I'm not really seeing the point of this inner loop. It seems like list at this time is a vector of indices. We should be able to do 

```
applicable_dates <- points[['time']]['list']
applicable_values <- points[['value']]['list']
```

R/utils-read.R
```diff
@@ -308,27 +308,34 @@ readApprovalPoints <- function(approvals, points, timezone, legend_nm, appr_var_
...
+          time=applicable_dates,
+          value=applicable_values,
+          legend.name=paste(level["type"], legend_nm),
+          point_type=point_type,
+          stringsAsFactors=FALSE))
       }
     }
-    approvals_all <- append(approvals_all, approval_info)
+    
+    approvals_all[[appr_var]] <- approval_info
```
lindsayplatt: my suggestion is to use an lapply to get rid of all that nested looping. Especially if my assumption that list is already just a vector.

```r
date_index_list <- list(list(type="Approved",index=approved_index), 
                        list(type="Working",index=working_index), 
                        list(type="In Review",index=review_index))

approvals_all <- lapply(date_index_list, function(level, points, legend_nm, point_type){
  
  d <- level[['index']]
  
  applicable_dates <- points[['time']][d]
  applicable_values <- points[['value']][d]
  
  if(any(!is.na(applicable_dates))) {
    approval_info_level <- data.frame(
      time=applicable_dates,
      value=applicable_values,
      legend.name=paste(level[["type"]], legend_nm),
      point_type=point_type,
      stringsAsFactors=FALSE)
  } else {
    approval_info_level <- data.frame(time=.POSIXct(character()),
                                      value=numeric(),
                                      legend.name=character(),
                                      point_type=numeric(),
                                      stringsAsFactors=FALSE)
  }
  
  return(approval_info_level)
}, points, legend_nm, point_type)

appr_type_ordered <- sapply(date_index_list, function(level){ level[['type']]})
names(approvals_all) <- appr_var_all[match(appr_type, appr_type_ordered)]
```

R/utils-read.R
```diff
@@ -308,27 +308,34 @@ readApprovalPoints <- function(approvals, points, timezone, legend_nm, appr_var_
...
-            value = applicable_values,
-            legend.name = paste(sub_list["type"], legend_nm),
-            point_type = point_type)
-      }
+  for(level in date_index_list){
+    
+    appr_var <- appr_var_all[which(appr_type == level["type"])]
+    list <- level[[2]]
+    
+    approval_info <- data.frame(time=.POSIXct(character()),
```
lindsayplatt: use `as.POSIXct(character())` when setting up empty dfs

lindsayplatt (approved):
would like to see some of those for loops change to lapplys before it's merged

R/utils-read.R
thongsav-usgs: For some reason this has different behavior. I'm having an issue where the original date format is not preserved if I do as.POSIXct(character()) (EG: DV's turn into UTC and get a 3am time)

The above preserves the original format of the incoming data.

lindsayplatt: interesting...because I was getting an error in some scenarios without the `as.`

PR #581, opened 2017-02-13
AQCU 1020 - Supposed gap error fix

dpattermann-usgs:
Fixed what I thought was a gap error fix but was actually a problem with reading non-estimated timeseries. Also, fixed a test that was broken for me in uvhydrograph-render-test and added another test that does a similar check.


R/utils-read.R
```diff
@@ -626,15 +626,15 @@ readEstimatedTimeSeries <- function(reportObject, seriesName, timezone, descript
...
-        } else {
-          estimatedSubset <- rbind(estimatedSubset, seriesData[['points']][seriesData[['points']][['time']] >= startTime & seriesData[['points']][['time']] < endTime,])
-        }
+        estimatedSubset <- rbind(estimatedSubset, seriesData[['points']][seriesData[['points']][['time']] >= startTime & seriesData[['points']][['time']] < endTime,])
       }
     }
     #Replace data with only saved data
-    seriesData[['points']] <- estimatedSubset
+    if(inverted){
+      seriesData[['points']] <- seriesData[['points']][ !(seriesData[['points']][['time']] %in% estimatedSubset[['time']]), ]
```
dpattermann-usgs: @lindsaycarr I tagged you for a review because I'm wondering if you know of a better way to do this than what I'm doing here (subsetting). I tried dplyr but couldn't get it to work, so I'm wondering if you have any advice.

lindsayplatt: I think you can just use `subset` to make it easier. It's what we use in `subsetByMonth`. Here's my reproducible example using test data:

```r
library(jsonlite)

reportObject <- fromJSON(system.file('extdata','testsnippets','test-time-series.json', package = 'repgen'))
tz <- repgen:::fetchReportMetadataField(reportObject, "timezone")
series <- repgen:::readTimeSeries(reportObject, "testSeries1", repgen:::fetchReportMetadataField(reportObject, "timezone"))

startE <- repgen:::flexibleTimeParse(series[['estimatedPeriods']][['startDate']], tz)
endE <- repgen:::flexibleTimeParse(series[['estimatedPeriods']][['endDate']], tz)


estTimeSeries <- subset(series[['points']], time >= startE & time <= endE)
nonEstTimeSeries <- subset(series[['points']], time < startE | time > endE)
```

I think your code could change to (except I don't really follow the circumstance surrounding the inverted scenario):
```r
for(i in 1:nrow(estimatedPeriods)) {
    p <- estimatedPeriods[i,]
    startTime <- p$start
    endTime <- p$end
    if(inverted){
      estimatedSubset_i <- subset(estimatedSubset, time >= startTime & time < endTime)
    } else {
      estimatedSubset_i <- ?
    }
    estimatedSubset <- rbind(estimatedSubset, estimatedSubset_i)
}
```

lindsayplatt: Actually, I kind of follow it. You're just pulling out the inverse scenario from the loop. So,

```r
for(i in 1:nrow(estimatedPeriods)) {
    p <- estimatedPeriods[i,]
    startTime <- p$start
    endTime <- p$end
    estimatedSubset_i <- subset(estimatedSubset, time >= startTime & time < endTime)
    estimatedSubset <- rbind(estimatedSubset, estimatedSubset_i)
}

if(inverted){
  nonEstimatedSubset <- subset(series[['points']], !time %in% estimatedSubset[['time']])
  series[['points']] <- nonEstimatedSubset
} else {
  series[['points']] <- estimatedSubset
}
```

lindsayplatt: :point_up: I know I've got a lot of intermediate objects there which seems unnecessary. Good commenting can probably eliminate the need for those

dpattermann-usgs: Yeah, the 'inverse' flag is kind of weird and I'm not huge fan of it, but your second one should work. I'll try that out and hopefully that makes this more readable.

lindsayplatt: yeah, `inverted` just doesn't seem like the right word. It is also used in `createGapsFromEstimatedPeriods` because we were "inversing" what the function returned.

R/utils-read.R
```diff
@@ -626,15 +626,16 @@ readEstimatedTimeSeries <- function(reportObject, seriesName, timezone, descript
         p <- estimatedPeriods[i,]
         startTime <- p$start
         endTime <- p$end
-        if(inverted){
-          estimatedSubset <- rbind(estimatedSubset, seriesData[['points']][seriesData[['points']][['time']] < startTime | seriesData[['points']][['time']] >= endTime,])
-        } else {
-          estimatedSubset <- rbind(estimatedSubset, seriesData[['points']][seriesData[['points']][['time']] >= startTime & seriesData[['points']][['time']] < endTime,])
-        }
+        estimatedSubset <- rbind(estimatedSubset, seriesData[['points']][seriesData[['points']][['time']] >= startTime & seriesData[['points']][['time']] < endTime,])
```
lindsayplatt: I would also use `subset` here for readability

lindsayplatt: rather than having a million `[[ ]]`, which is confusing to look at

dpattermann-usgs: Yeah, I'll change that up, too. When I was first looking at this it really threw me for a loop.

zmoore-usgs (approved):
Changes look good to me. I agree that the inverted/non-inverted stuff in readEstimatedTimeSeries is not very clean, but the changes you made definitely make it more readable.

lindsayplatt (approved):

PR #582, opened 2017-02-14
Continued UV hydro refactor/testing

thongsav-usgs:
- Testing around secondary plot toggling
- Minimal case tests for primary/secondary plot and some method signature refactoring


R/uvhydrograph-render.R
```diff
@@ -22,7 +22,9 @@ uvhydrographPlot <- function(reportObject) {
     for (month in months) {
       primary <- getPrimaryReportElements(reportObject, month, timezone, excludeZeroNegativeFlag)
       
-      if(!is.null(primary[['plot']])){
+      
+      
+      if(!is.null(primary[['plot']]) && useSecondaryPlot(reportObject)){
```
thongsav-usgs: secondary plotting toggle moved up to here.

R/uvhydrograph-render.R
```diff
@@ -100,32 +115,31 @@ getSecondaryReportElements <- function(reportObject, month, timezone, excludeZer
...
-          corrections, 
-          secondaryLims,
-          timezone, 
-          excludeZeroNegativeFlag, 
-          tertiary_label=getTimeSeriesLabel(reportObject, "effectiveShifts"))
-      secondaryTable <- parseCorrectionsAsTable(corrections)
-    } else {
-      secondary_status_msg <- paste('Corrected data missing for', fetchReportMetadataField(reportObject, 'secondaryParameter'))
-    }
+  if(hasReferenceSeries) {
```
thongsav-usgs: Those whole section used to be nested in an if block that toggled secondary plotting on/off. It's indented back because that logic has moved up a level. No functional changes here.

tests/testthat/test-uvhydrograph-render.R
```diff
@@ -3,6 +3,86 @@ context("uvhydrograph-render tests")
...
+  expect_equal(length(renderList[['1206']]), 6) #2 plots, 2 corrections tables, and 2 status messages
+  expect_false(is.null(renderList[['1206']][['plot1']]))
+  expect_true(is.null(renderList[['1206']][['plot2']])) #skipped
+  expect_false(is.null(renderList[['1206']][['table1']]))
+  expect_true(is.null(renderList[['1206']][['table2']])) #skipped
+  expect_true(is.null(renderList[['1206']][['status_msg1']]))
+  expect_true(is.null(renderList[['1206']][['status_msg2']]))
+})
+
+test_that("useSecondaryPlot correctly flags when to use a secondary plot",{
```
thongsav-usgs, line 80: We need to verify the requirements on when to turn on/off secondary plot and add to this test.

R/uvhydrograph-data.R
```diff
@@ -25,6 +25,21 @@ parseCorrectionsByMonth <- function(reportObject, seriesName, month) {
...
+#' @description depending on the report configuration, corrections might come from a reference or upchain series
+#' @param reportObject the report to render
+#' @return corrections list from the correct series
+parseSecondaryCorrectionsByMonth <- function(reportObject, month) {
+  hasReferenceSeries <- hasReferenceSeries(reportObject)
+  hasUpchainSeries <- hasUpchainSeries(reportObject)
+  
+  if(hasReferenceSeries) {
+    corrections <- parseCorrectionsByMonth(reportObject, "referenceSeriesCorrections", month)
+  } else {
```
zmoore-usgs, line 38: I think this should be an `else if(hasUpchainSeries)` in case it also does not have an upchain series. Or if that's handled by the parseCorrectionsByMonth function then perhaps this entire function could be reduced to something like:

```
corrections <- parseCorrectionsByMonth(reportObject, "referenceSeriesCorrections", month)

if(isEmptyOrBlank(corrections)){
    corrections <- parseCorrectionsByMonth(reportObject, "upchainSeriesCorrections", month)
}
```

R/uvhydrograph-data.R
```diff
@@ -25,6 +25,21 @@ parseCorrectionsByMonth <- function(reportObject, seriesName, month) {
...
+    corrections <- parseCorrectionsByMonth(reportObject, "upchainSeriesCorrections", month)
+  }
+}
```
zmoore-usgs: Lindsay usually recommends having an explicit return at the end of functions instead of using the implicit one.

R/uvhydrograph-render.R
zmoore-usgs: I've been in the habit of replacing pretty much all `is.null()` calls with `isEmptyOrBlank` calls, though if this should only happen if it is null and not if it is just empty then it should be kept as is.

R/uvhydrograph-render.R
```diff
@@ -43,6 +45,19 @@ uvhydrographPlot <- function(reportObject) {
...
 
+#' Use Secondary Plot
+#' @description Determines where or not a report should use a secondary plot 
+#' @param reportObject the report to be rendered
+#' @return boolean true or false of whether a secondary series is appropriate for report
+useSecondaryPlot <- function(reportObject) {
+  hasReferenceSeries <- hasReferenceSeries(reportObject)
+  hasUpchainSeries <- hasUpchainSeries(reportObject)
+  
+  useSecondaryPlot <- (hasReferenceSeries && !isPrimaryDischarge(reportObject)) || hasUpchainSeries
```
zmoore-usgs: For consistency maybe the `isPrimaryDischarge()` call should be pulled out to a variable like `hasReferenceSeries` and `hasUpchainSeries`, or those 2 calls could be embedded into the this statement?

zmoore-usgs (commented):
Nothing major, just little tweaks / comments.

PR #583, opened 2017-02-14
Bug fix for AQCU-1021.

dpattermann-usgs:
Caused by an error in expandYLims that actually shrunk Y-lims.


R/utils-plot.R
```diff
@@ -105,11 +105,13 @@ extendYaxisLimits <- function(gsplot, error_bar_args){
...
+  ##Find the max and min of errors with bars.
   lowest_error_bar <- min(error_bar_args[['y']] - error_bar_args[['y.low']])
-  lowest_y <- min(ylim(gsplot, side=side)[1], lowest_error_bar)
-  
   highest_error_bar <- max(error_bar_args[['y']] + error_bar_args[['y.high']])
-  highest_y <- max(ylim(gsplot, side=side)[2], highest_error_bar)
+  
+  ##Compare error bar extrema with current plot lims.
+  lowest_y <- min( min(ylim(gsplot, side=side)), lowest_error_bar)
+  highest_y <- max( max(ylim(gsplot, side=side)), highest_error_bar)
```
lindsayplatt: if the axis is reversed, wouldn't lowest_y be wrong now? I think that's why we were using `ylim(...)[1]`

dpattermann-usgs:
@zmoore-usgs You wanna look at this for a sanity check, too? I want to make sure that this makes sense.

zmoore-usgs (approved):
Changes look good to me! That added documentation makes this much clearer as to what it's doing.

R/utils-plot.R
```diff
@@ -98,23 +98,44 @@ DelineateYearBoundaries <- function(object, years) {
...
-#' @description Extends the Y-Axis limits if the error bars extend further than the current limits
+#' @description Extends the Y-Axis limits if the error bars extend further than the current limits.
+#' Sets the y-limits to the smallest (and largest) value between the error bars and current plot limits.
 #' @param gsplot The gsplot object to extend limits for
-#' @param error_bar_args the error bar arguments to check the limits against
-extendYaxisLimits <- function(gsplot, error_bar_args){
-  side <- ifelse(!is.null(error_bar_args[['side']]), error_bar_args[['side']], 2)
+#' @param limits The minimum and maximum points to compare to Y-axis limits
+#' @param side The axis to be changed/compared to (default is 2, the y-axis).
+extendYaxisLimits <- function(gsplot, limits, side){
```
lindsayplatt: can we name the limits argument, `newLimits` or something? To make it more clear. Also, in the description, be sure to say that is is a list of length 2, where names are minPoint and maxPoint. Or make it simpler by just having it be a vector of form `c(minimum, maximum)`. Then update `getErrorBarYLims` to return limits as a vector of length 2

R/utils-plot.R
```diff
@@ -98,23 +98,44 @@ DelineateYearBoundaries <- function(object, years) {
...
   side_nm <- paste0('side.', side)
   
-  lowest_error_bar <- min(error_bar_args[['y']] - error_bar_args[['y.low']])
-  lowest_y <- min(ylim(gsplot, side=side)[1], lowest_error_bar)
-  
-  highest_error_bar <- max(error_bar_args[['y']] + error_bar_args[['y.high']])
-  highest_y <- max(ylim(gsplot, side=side)[2], highest_error_bar)
+  ##Compare error bar extrema with current plot lims.
+  lowest_y <- min( min(ylim(gsplot, side=side)), limits[['minPoint']])
+  highest_y <- max( max(ylim(gsplot, side=side)), limits[['maxPoint']])
```
lindsayplatt: need the interior min and max calls?

lindsayplatt (approved):

PR #584, opened 2017-02-16
Improvements

thongsav-usgs:
1) Limits expansion bug for multiple error
2) Upped testing coverage
3) Removed deprecated getTimeSeries function


R/utils-limits.R
```diff
@@ -9,8 +9,17 @@ getErrorBarYLims <- function(error_bar_args){
   side <- ifelse(!is.null(error_bar_args[['side']]), error_bar_args[['side']], 2)
   
   ##If uncertainty does not exist, replace with 0.
-  ylow <- ifelse(!isEmptyOrBlank(error_bar_args[['y.low']]), abs(error_bar_args[['y.low']]), 0)
-  yhigh <- ifelse(!isEmptyOrBlank(error_bar_args[['y.high']]), abs(error_bar_args[['y.high']]), 0)
+  if(isEmptyOrBlank(error_bar_args[['y.low']])) {
+    ylow <- 0
+  } else {
+    ylow <- ifelse(lapply(error_bar_args[['y.low']], isEmptyOrBlank), 0, error_bar_args[['y.low']])
```
dpattermann-usgs, line 15: Why not just immediately run this (it will still catch the first case, right?)

thongsav-usgs: The first check tests if the variable is null. The second test if individual items inside the non-null list are null/empty.

thongsav-usgs: One of the tests breaks without the first check.

dpattermann-usgs: Ah, I see it now. It runs differently than I expected (it fills them in element by element, pretty useful). 👍

PR #585, opened 2017-02-17
AQCU-986: Refactor Corr Report

zmoore-usgs:


thongsav-usgs (approved):

lindsayplatt:
@thongsav-usgs was going to review today :( you beat me to the merge

thongsav-usgs:
Pretty sure @zmoore-usgs is still working and could still use more review. I merged to get this out on DEV for testing since we may need to cut a release ASAP.

lindsayplatt:
Ok, I will still review!!!

R/correctionsataglance-data.R
```diff
@@ -1,357 +1,539 @@
...
+    approvals[['description']] <- paste("Approval:", approvals[['description']])
+    colors <- c()
     
-    ytop <- vector(mode = "numeric", length = len_data)
-    ybottom <- ytop
+    labels <- format(dateSeq, "%m/%Y")
     
-    ytop[] <- startH
-    ybottom[] <- startH - height
+    for (i in 1:length(approvals[['level']])) {
```
zmoore-usgs, line 152: I think this should probably be changed to an sapply, though the number of approvals is so small that it probably wouldn't make much of a difference. I can definitely make the change along with any other suggested changes though.

R/correctionsataglance-data.R
```diff
@@ -1,357 +1,539 @@
...
-  }
-
-  return(typeData)
+#' Get a list of end dates
+#' @description Given a startSeq date list and end date and whether 
+#' or not the end date is the first of the month, provides a list of 
+#' YYYY-MM-DD used in other functions
+#' @param startSeq the start date
+#' @param endD the end date
+#' @param firstOfMonth_end TRUE or FALSE whether the end date is the first of the month
```
lindsayplatt, line 39: @zmoore-usgs `firstOfMonth_end` is not an argument used here. Should it be added? Or should it be deleted?

zmoore-usgs: @lindsaycarr Whoops, no that's leftover from before I refactored it. There might actually be several leftover param tags that aren't relevant anymore. I didn't think to look for tags to remove, only tags that needed to be added. I'll take a look around for more

R/correctionsataglance-data.R
```diff
@@ -1,357 +1,539 @@
...
+      endDates <- periods[['endTime']]
+      value <- periods[['referenceValue']]
+      suppressData <- periods[['suppressData']]
       
+      formattedData <- rbind(formattedData, data.frame(type, startDates, endDates, value, suppressData))
     }
     
-    allData[[d]][['ytop']] <- ytop
-    allData[[d]][['ybottom']] <- ybottom
+    formattedData <- formattedData[which(formattedData[['suppressData']]),]
```
lindsayplatt, line 203: wouldn't you want to exclude suppressed data?

R/correctionsataglance-render.R
```diff
@@ -1,150 +1,214 @@
...
+
+  #Parse Basic Plot Data
+  primarySeriesRequiredFields <- c('approvals','qualifiers','grades','notes')
+  primarySeries <- readTimeSeries(reportObject, 'primarySeries', timezone, descriptionField = 'primaryParameter', requiredFields=primarySeriesRequiredFields)
+  preData <- parseCorrProcessingCorrections(reportObject, "pre", timezone)
+  normalData <- parseCorrProcessingCorrections(reportObject, "normal", timezone)
+  postData <- parseCorrProcessingCorrections(reportObject, "post", timezone)
+
+  #Parse Optional Plot Data
+  fieldVisitData <- tryCatch({
```
lindsayplatt, line 23: should this be in its own parse function to follow all the other patterns?

R/correctionsataglance-render.R
```diff
@@ -1,150 +1,214 @@
...
+                          laneData = thisLane,
+                          laneName = laneName,
+                          dateRange = dateRange,
+                          rectHeight = rectHeight)
   }
 
-  return(list(timeline = timeline, tableOfLabels = parseData$tableData))
-  
+  #Create Label Table 
+  if(!isEmptyOrBlank(tableLabels)){
```
lindsayplatt, line 109: would you want this logic inside of `createLabelTable`?

lindsayplatt (commented):
should there be new tests for the new functions in utils-time?

R/correctionsataglance-data.R
zmoore-usgs: That's what I thought as well but it was originally implemented to include only suppressed data and exclude data that wasn't suppressed so I kept the same behavior. I'm thinking "suppress data" must mean something other than what it seems to mean at face value?

lindsayplatt: Hmm yeah, must be. @thongsav-usgs can you clarify "suppressData"?

PR #587, opened 2017-02-21
AQCU-961: Restored missing dvhydro & five YR approvals explanation label

zmoore-usgs:


zmoore-usgs:
@thongsav-usgs Could you review this real quick? Just a small change restoring a label that got lost during my refactor.

R/dvhydrograph-render.R
```diff
@@ -122,13 +122,16 @@ createDVHydrographPlot <- function(reportObject){
...
+  plot_object <- mtext(plot_object, text = "Displayed approval level(s) are from the source TS that statistics are derived from.", side=3, cex=0.85, line=0.33, adj=1, axes=FALSE)
+
   #Legend
   #If the time period is greater than 1 year additional x-axis labels are added so we must move the legend down further
   #Legend offset also needs to be calculated based on the number of lines and columns to be consistent in position
   legend_items <- plot_object$legend$legend.auto$legend
   ncol <- ifelse(length(legend_items) > 3, 2, 1)
   leg_lines <- ifelse(ncol==2, ceiling((length(legend_items) - 6)/2), length(legend_items))
-  legend_offset <- ifelse(ncol==2, 0.09+(0.025*leg_lines), (0.025*leg_lines))
+  legend_offset <- ifelse(ncol==2, 0.1+(0.025*leg_lines), 0.05+(0.025*leg_lines))
```
thongsav-usgs: do we want these numbers pulled into styles?

R/fiveyeargwsum-render.R
```diff
@@ -104,14 +104,17 @@ createfiveyeargwsumPlot <- function(reportObject){
   # Add space to the top of the Y Axis
   plot_object <- RescaleYTop(plot_object)
 
+  #Add approval explanation label to the top of the plot
+  plot_object <- mtext(plot_object, text = "Displayed approval level(s) are from the source TS that statistics are derived from.", side=3, cex=0.6, line=0.1, adj=1, axes=FALSE)
+
   #Add Min/Max labels if we aren't plotting min and max
-  line <- 0.33
+  line <- 0.1
```
thongsav-usgs: same question about styles.

R/dvhydrograph-render.R
zmoore-usgs: I'm not sure if the legend_offset numbers would really fit in styles or not since there are actual calculations done to obtain them which wouldn't fit as cleanly into the CSS-like model we were using for styles.

R/fiveyeargwsum-render.R
zmoore-usgs: These numbers I could see possibly being pulled into styles since they aren't really calculated on, though this one is modified as part of the loop so I don't know if it specifically would fit? The numbers within the mtext() gsplot call would probably make sense in styles though

zmoore-usgs: I suppose I could just add a modifier based on the loop iteration each time rather than modifying the line value itself and then it might fit better into the styles.

PR #593, opened 2017-02-22
Separated vdiagram data into two distinct types.

thongsav-usgs:


dpattermann-usgs (approved):
Looks fine to me, as long as the hasEnoughData function is right in what we do and don't need. That would be the only thing I would double check.

PR #594, opened 2017-02-23
Fixed in min data detection and updated documentation

thongsav-usgs:


thongsav-usgs:
@dpattermann-usgs you called it in the last PR, the logic for hasEnoughData was incorrect. I am going to merge this as soon as CI passes to get DEV cleared but would like a review after the fact.

PR #596, opened 2017-02-23
AQCU-1005: UV Hydro: Show readings for the TS on the secondary plot

zmoore-usgs:
Should be merged at the same time as https://github.com/USGS-CIDA/AQCU/pull/664


R/uvhydrograph-data.R
```diff
@@ -262,11 +262,12 @@ readSecondaryUvHydroApprovalBars <- function(reportObject, timezone) {
 #' @description Read readings subsetted by month and separated by type for UV Hydrograph
 #' @param reportObject entire UV Hydro report object
 #' @param month subset only into this month
+#' @param derivation [DEFAULT: "primary"] The derivation chain position to get the readings from 
 #' @return named list of readings by type and subsetted by month
-readAllUvReadings <- function(reportObject, month) {
-  ref_readings <- subsetByMonth(readReadings(reportObject, "reference"), month)
-  csg_readings <- subsetByMonth(readReadings(reportObject, "crestStage"), month)
-  hwm_readings <- subsetByMonth(readReadings(reportObject, "waterMark"), month)
+readAllUvReadings <- function(reportObject, month, derivation="primary") {
```
thongsav-usgs: EG: readingsJsonField="primaryReadings" here.

R/utils-read.R
```diff
@@ -635,12 +635,14 @@ readMeanGageHeights<- function(reportObject){
...
 #' @param filter optional filter to restrict to reading types (reference, crestStage, or waterMark)
+#' @param derivation [DEFAULT: "primary"] The derivation chain position to get the readings from 
 #' @return data frame of reading records
-readReadings <- function(reportObject, filter="") {
-  time <- as.POSIXct(strptime(reportObject[['readings']][['time']], "%FT%T"))
-  value <- as.numeric(reportObject[['readings']][['value']])
-  type <- reportObject[['readings']][['type']]
-  uncertainty <- as.numeric(reportObject[['readings']][['uncertainty']])
+readReadings <- function(reportObject, filter="", derivation="primary") {
+  jsonField <- paste0(derivation, "Readings")
```
thongsav-usgs: I don't think we should do this hidden convention and instead force all callers to pass in the full field name. I'm not sure "derivation" is the correct term. I'd go with "readingsFieldName" to be explicit.

thongsav-usgs (changes requested):

PR #605, opened 2017-03-10
AQCU-1035 CORR- Always show the threshold lane

mhines-usgs:
-creating an empty threshold row with the start/end dates from reportMetadata and a null label; if there are threshold data these are overwritten with actual data, if not, an empty row appears.

### DIFF
--- a/R/correctionsataglance-render.R
+++ b/R/correctionsataglance-render.R
@@ -41,8 +41,7 @@ correctionsataglanceReport <- function(reportObject) {
   dataLanes <- allLaneData[['dataLanes']]
   processOrderLabelYPos <- mean(c(dataLanes[['preData']][["laneNameYPos"]],
                                   dataLanes[['normalData']][["laneNameYPos"]],
-                                  dataLanes[['postData']][["laneNameYPos"]],
-                                  dataLanes[['thresholdData']][["laneNameYPos"]]))
+                                  dataLanes[['postData']][["laneNameYPos"]]))
   
   #Create Base Plot
   timeline <- gsplot() %>% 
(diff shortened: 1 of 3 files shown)


mhines-usgs:
image

zmoore-usgs (changes requested):
These changes seem to work but I think there is a somewhat cleaner way to get the lane to always show up. If you look in the render file the call to `createPlotLanes` takes arguments of `optionalData` and `requiredData`. I think if you move the threshold stuff from optional to required it should make the lane always show. You might have to adjust the math for `rectHeight` in `createPlotLanes` though because I think it's based on the number of required/optional lanes.

mhines-usgs:
Thanks @zmoore-usgs -- checking that out. I think I've got it working how you suggest with the few examples from the JIRA now, but the height is still a little funky. Will keep looking.

zmoore-usgs:
Woah, my comment ended up in some weird place where it doesn't show up in the files changed view... 😕  I'll try again

R/correctionsataglance-render.R
```diff
@@ -41,7 +41,8 @@ correctionsataglanceReport <- function(reportObject) {
   dataLanes <- allLaneData[['dataLanes']]
   processOrderLabelYPos <- mean(c(dataLanes[['preData']][["laneNameYPos"]],
                                   dataLanes[['normalData']][["laneNameYPos"]],
-                                  dataLanes[['postData']][["laneNameYPos"]]))
+                                  dataLanes[['postData']][["laneNameYPos"]],
+                                  dataLanes[['thresholdData']][["laneNameYPos"]]))
```
zmoore-usgs: This part of the code is adjusting the y position of the "processing order" label to be in the middle of the processing order lanes. I'm not sure if threshold data is supposed to go under the "Processing Order" label or not? I don't think it was under it before when the lane show up, though I'm not totally sure.

zmoore-usgs (approved):
Changes look good, just a question about the processing order label. If threshold data is supposed to be under it then I think this looks good!

PR #608, opened 2017-03-14
AQCU-1031 revisions to uvhydro styles take I

mhines-usgs:


R/uvhydrograph-styles.R
```diff
@@ -4,45 +4,75 @@
 #' @importFrom grDevices rgb
 getUvStyles <- function() {
   styles=list(
+      #raw signal should be green
+      uncorr_UV_lbl="Uncorrected UV",
+      uncorr_UV_lines=list(col="green", lty=4),
+      uncorr_UV2_lines=list(col="green", lty=4),
```
thongsav-usgs: Is there a reason we need separate styles if both are the same?

thongsav-usgs (approved):
Looks fine to me.

R/uvhydrograph-styles.R
mhines-usgs: I can look into the render file to see tomorrow, not sure...

mhines-usgs: Yep, it was only there because it was a different color before, but now I have merged the colors and so the style and renderer file were updated to use the not 2 style.

PR #615, opened 2017-03-17
AQCU-994: Modified reading of SVP field visits to move associated IV …

zmoore-usgs:
…information to the highest-value reading per visit for visits that have mutliple readings.


mhines-usgs:
@zmoore-usgs it looks like its reverse sorting (latest -> earliest) now, is that an oversight or you wanted that change?

image

zmoore-usgs:
@mhines-usgs Woah, that's weird. I'm not seeing that locally. Could you send me the json you're using?

mhines-usgs:
Looks like this now, seems correct? 
image

mhines-usgs (approved):
Seems to work correctly, and it looks like there was a test that caught the output incorrect? :+1:

zmoore-usgs:
Yeah, there was already a test in-place testing the order of the rows, but it passed locally since I didn't have the newest version of dplyr.

PR #618, opened 2017-03-20
AQCU-1046: Extremes: Last occurrence of peak is showing, not first

zmoore-usgs:


R/extremes-data.R
```diff
@@ -240,16 +240,16 @@ createDataRows <-
...
+              duplicateRows <- dataRows %>% arrange(related) %>% group_by(related) %>% arrange(date, time) %>% ungroup() %>% as.data.frame()
             }
           } else {
             #Keep Maximum or Minimum based on current param
             if(param == "max"){
-              duplicateRows <- dataRows[order(dataRows$date, dataRows$primary, dataRows$time, decreasing = TRUE),]
+              duplicateRows <- dataRows %>% arrange(desc(primary)) %>% group_by(primary) %>% arrange(date, time) %>% ungroup() %>% as.data.frame()
             } else  if (param == "min"){
-              duplicateRows <- dataRows[order(dataRows$date, dataRows$primary, dataRows$time, decreasing = FALSE),]
+              duplicateRows <- dataRows %>% arrange(primary) %>% group_by(primary) %>% arrange(date, time) %>% ungroup() %>% as.data.frame()
```
lindsayplatt, line 252: the first occurrence of `arrange()` won't matter because then you group by primary. You should be able to leave that off.

lindsayplatt: I see other instances of using arrange before group_by, you can delete all of those and just do group_by. You use arrange right after to get the correct order anyways

zmoore-usgs: The reasoning behind using arrange before the group_by is because I want the group with the highest (or lowest) value for its parameter (primary or related) to be at the top of the resulting data frame. Is that done by the group_by implicitly?

lindsayplatt (approved):
just delete unnecessary bits

R/extremes-data.R
lindsayplatt: Ahhhhhh yeah, you are right. I did not give that enough thought. I guess it could be cleaner by putting all of the arrange stuff in one call, but that's more an aesthetics thing.

zmoore-usgs: Ah okay, gotcha. I wasn't sure how to put it all into one call since the second arrange needs to happen within the groups created by the group_by. 

Also not sure why that test is failing on Travis. Just ran check locally and everything passed... 😕

tests/testthat/test-extremes.R
```diff
@@ -390,7 +390,7 @@ test_that("proper number of rows are created based on data",{
   library(jsonlite)
   library(dplyr)
   reportObject <- fromJSON(system.file('extdata','extremes','extremes-multiple-min-max-test.json',package = 'repgen'))
-  expect_true(NROW(createDataRows(reportObject[[which(names(reportObject) %in% c("upchain"))]], "max", "Max", TRUE)[[1]]) == 2)
+  expect_true(NROW(createDataRows(reportObject[[which(names(reportObject) %in% c("upchain"))]], "max", "Max", TRUE)[[1]]) == 3)
   expect_true(NROW(createDataRows(reportObject[[which(names(reportObject) %in% c("primary"))]], "max", "Max", FALSE)[[1]]) == 2)
   expect_true(NROW(createDataRows(reportObject[[which(names(reportObject) %in% c("upchain"))]], "min", "min", TRUE)[[1]]) == 1)
```
lindsayplatt: this line returns 2 rows now

zmoore-usgs: Huh, on my machine I still seem to only be getting 1.. 😕

R/extremes-data.R
lindsayplatt: right, you're arranging the groups, right? Should be fine:

```
df <- data.frame(a=1:6, b=6:1, c=2:1)
df %>% group_by(c) %>% arrange(c, b)
```

tests/testthat/test-extremes.R
lindsayplatt: weird

lindsayplatt: I also get 
```
Quitting from lines 41-47 (extremes.Rmd) 
Error in function_list[[i]](value) : could not find function "group_by"
```
but I don't know where that's coming from

R/extremes-data.R
zmoore-usgs: Ah cool. I was thinking that the arrange that happened after the group_by would be completely within each group and so arranging by the parameter that was grouped by would have no affect. Thanks!

PR #623, opened 2017-03-28
Gsplot upgrade to 0.8.0

lindsayplatt:
NOT `v0.8.0`, just `0.8.0`. Messed up first release, so had to rename.

Actual work to update to the new version:
- fix tests that count the elements in a view because `error_bar` is now passed all the way through so there were 2 arrow calls for what is now 1 error bar 
- fixed how `vdiagram` axes were added. It was incorrectly grouping them all which causes each side to have the same ticks (e.g. `gsplot() %>% points(1:4,1:4) %>% axis( side = c(1, 2, 4), at = c(1:2, 2:4, 1:2))` sets sides 1,2, and 4 ticks to the vector `1 2 2 3 4 1 2`)
- adjusted how to add axis labels to the more accepted gsplot way
- removed unneccessary ablines (grid adds those, so abline was a duplicate) AND removed the unused abline styles
- updated the gsplot version used in `.travis.yml`

There are a few things that I changed along with this PR that bothered me during the `Check`:
- eliminated notes about "missing global variables" (changes to `extremes-data.R` and `utils-read.R`)
- eliminated notes about a code line that was too long (see `sig-dvhydrograph.R`)
- eliminated notes about using `::` vs `:::` (see `utils-read.R` -- @zmoore-usgs `:::` are used when the function is not exported from the package)

I also corrected a typo that has been bothering me (changes to `dvhydrograph-render.R` and `uvhydrograph-data.R`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - Rscript -e 'install.packages(c("devtools", "testthat"))'
 
   #Install packages from github
-  - Rscript -e 'devtools::install_github("usgs-r/gsplot#420")'
+  - Rscript -e 'devtools::install_github("usgs-r/gsplot#469")'
   - Rscript -e 'devtools::install_github("jimhester/covr")'
   - Rscript -e 'Sys.setenv(TZ="UTC")'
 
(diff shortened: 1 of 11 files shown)


dpattermann-usgs:
Seems fine to me.

zmoore-usgs (approved):

PR #624, opened 2017-03-29
AQCU-964: Add discrete WQ data to DV Hydro

zmoore-usgs:


R/dvhydrograph-render.R
```diff
@@ -293,6 +295,9 @@ getDVHydrographPlotConfig <- function(plotItem, plotItemName, yLabel="", maxIvLa
     groundWaterLevels = list(
       points = append(list(x=x, y=y, legend.name="Measured Water Level (GWSI)"), styles$gw_level_points)
     ),
+    waterQualityData = list(
```
thongsav-usgs, line 298: Will the NULL case break this or was there already a null check somewhere not part of the PR?

zmoore-usgs: I think it should be handled but I'll double check

zmoore-usgs: Yeah NULLs are handled by the plotItem function and if the item to be plotted is null it is skipped and the fetch style function is never called.

PR #625, opened 2017-03-29
AQCU-1049 Add check to value in threshold so that NULL's are changed to NA

dpattermann-usgs:
For AQCU-1049. @zmoore-usgs Not sure if this is the best way to do this, this seem right to you?

### DIFF
--- a/R/correctionsataglance-data.R
+++ b/R/correctionsataglance-data.R
@@ -203,7 +203,7 @@ parseCorrThresholds <- function(reportObject, timezone){
       type <- rep(threshold[['type']], times=nrow(periods))
       startDates <- periods[['startTime']]
       endDates <- periods[['endTime']]
-      value <- periods[['referenceValue']]
+      value <- ifelse(isEmptyOrBlank(periods[['referenceValue']]), NA, periods[['referenceValue']])
       suppressData <- periods[['suppressData']]
       
       formattedData <- rbind(formattedData, data.frame(referenceCode, type, startDates, endDates, value, suppressData))
--- a/R/correctionsataglance-data.R
+++ b/R/correctionsataglance-data.R
@@ -203,7 +203,7 @@ parseCorrThresholds <- function(reportObject, timezone){
       type <- rep(threshold[['type']], times=nrow(periods))
       startDates <- periods[['startTime']]
       endDates <- periods[['endTime']]
-      value <- ifelse(isEmptyOrBlank(periods[['referenceValue']]), NA, periods[['referenceValue']])
+      value <- if(isEmptyOrBlank(periods[['referenceValue']])){NA} else{periods[['referenceValue']]}
       suppressData <- periods[['suppressData']]
       
       formattedData <- rbind(formattedData, data.frame(referenceCode, type, startDates, endDates, value, suppressData))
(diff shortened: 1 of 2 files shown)


zmoore-usgs (approved):
Looks good to me, that's what I'd do. But maybe @mhines-usgs or @lindsaycarr has a better suggestion?

R/correctionsataglance-data.R
```diff
@@ -203,7 +203,7 @@ parseCorrThresholds <- function(reportObject, timezone){
       type <- rep(threshold[['type']], times=nrow(periods))
       startDates <- periods[['startTime']]
       endDates <- periods[['endTime']]
-      value <- periods[['referenceValue']]
+      value <- if(isEmptyOrBlank(periods[['referenceValue']])){NA} else{periods[['referenceValue']]}
```
lindsayplatt: generally when you use the `if() {} else {}` syntax, you don't want it in a single line (meaning, you don't want to set that equal to an object). You set the object value inside the if else statement. So you would need:

```r
if(isEmptyOrBlank(periods[['referenceValue']]) {
    value <- NA
} else {
    value <- periods[['referenceValue']]
}
```

If you know that `periods[['referenceValue']]` is only ever going to have one value in it, you can do 
```r
value <- ifelse(isEmptyOrBlank(periods[['referenceValue']]), NA, periods[['referenceValue']])
```

lindsayplatt: the `ifelse` function WILL NOT WORK if you have something > length 1!!!

dpattermann-usgs: That was my problem. `periods[['referenceValue']]` is a vector, and when I used ifelse, it caused it to only use the first one.
I'm open to changing it to multi-line if else, I just thought it looked weird assigning that to a variable.

dpattermann-usgs: Also, ifelse can work on things  > length 1, can't it? Apparently it returns a vector the same length as what you ask it to check? (although it doesn't work in this case) http://stackoverflow.com/questions/1335830/why-cant-rs-ifelse-statements-return-vectors

mhines-usgs: I'd defer to Lindsay, not sure if the ifelse is a must have but if the if / else works... :shipit:

R/correctionsataglance-data.R
```diff
@@ -203,7 +203,11 @@ parseCorrThresholds <- function(reportObject, timezone){
       type <- rep(threshold[['type']], times=nrow(periods))
       startDates <- periods[['startTime']]
       endDates <- periods[['endTime']]
-      value <- periods[['referenceValue']]
+      if(isEmptyOrBlank(periods[['referenceValue']])){
+        value <- NA
+      } else {
+        value <- periods[['referenceValue']]
+      }
```
dpattermann-usgs, line 210: @lindsaycarr This is what you meant?

lindsayplatt: @dpattermann-usgs yep! Just not normal R syntax to do it in one line, and I don't know if that has implications down the road.

R/correctionsataglance-data.R
lindsayplatt: @dpattermann-usgs the statement you're using to get T/F would need to return a vector the same length. But `isEmptyOrBlank` returns just one TRUE or FALSE so it wouldn't work in this particular scenario.

PR #626, opened 2017-04-03
AQCU-975: Changed the "waterdata.usgs.gov URL" text to "NWISWeb URL"

zmoore-usgs:
The added .Rd file was missing before, but the function it is representing was not modified.


dpattermann-usgs:
👍

mhines-usgs (approved):

PR #627, opened 2017-04-10
AQCU-1052 first pass fix for extremes report

mhines-usgs:
error seen, when primary and related upchain number of values mismatch an internal error occurred
now, it merges the two datasets into one dataframe and assigns NAs to values missing in one or the other.


R/extremes-data.R
```diff
@@ -210,7 +212,27 @@ createDataRows <-
         {
           relatedValue <- relatedSet$value
         }
+        
+        if (!isEmptyOrBlank(x$relatedPrimary)) {
+          if (nrow(x$relatedPrimary) > nrow(x$points) ||
+              nrow(x$points) > nrow(x$relatedPrimary)) {
```
dpattermann-usgs: Could these two if statements be simplified to nrow(x$related{Primary or Upchain}) != nrow(x$points) ?

mhines-usgs: is that a thing?! i'm actually writing a whole separate function now to reduce the duplicity of these two sets of statements as well and will try that..

dpattermann-usgs: Ha, I just mean replace each of those statements, it won't just be one statement.

dpattermann-usgs: Like (if(nrow(x$relatedPrimary) != nrow(x$points){ blah }
       if(nrow(x$relatedUpchain) != nrow(x$points)){ blah }
Sorry for the confusion, I don't have any secret R magic to get rid of the duplication off the top of my head.

zmoore-usgs (changes requested):
The approach looks good to me, though I'd suggest instead of giving each row the footnote string perhaps "footnote" should just hold TRUE or FALSE and then instead of including the if/else in the for loop in extremesTable I think you could just do return(list(toRet=toRet, footnote=any(dataRows$footnote))

Then in the RMD file add the actual footnote text if footnote=true, similar to how the other notes are being added about the '*' characters

mhines-usgs:
thanks @zmoore-usgs now that I got all the tests working again, I'll give that a try.

zmoore-usgs:
I took this ticket from @mhines-usgs after she left and made a few tweaks. New PR: https://github.com/USGS-R/repgen/pull/637

PR #631, opened 2017-04-10
AQCU-899: Ability to log different sides

zmoore-usgs:


mhines-usgs:
Will pull this locally and give it a whirl in a moment!

mhines-usgs (approved):

PR #633, opened 2017-04-11
AQCU-899: Ability to log different sides

zmoore-usgs:


.travis.yml
```diff
@@ -18,7 +18,7 @@ install:
   - Rscript -e 'install.packages(c("devtools", "testthat"))'
 
   #Install packages from github
-  - Rscript -e 'devtools::install_github("usgs-r/gsplot#469")'
+  - Rscript -e 'devtools::install_github("USGS-R/gsplot", ref="v0.8.1")'
```
lindsayplatt: use PR number instead - #473

R/uvhydrograph-render.R
```diff
@@ -180,47 +180,44 @@ createPrimaryPlot <- function(
...
   if((limsAndSides[['sides']][['reference']] == 4) || (limsAndSides[['sides']][['comparison']] == 4)){
-    plot_object <- lines(plot_object, x=0, y=0, side = 4, reverse = primarySeriesList[['inverted']]) %>%
+     plot_object <- lines(plot_object, x=0, y=0, side = 4, reverse = primarySeriesList[['inverted']]) %>%
         axis(side = 4, las = 0, reverse = primarySeriesList[['inverted']])
   }
   
+  #uncorrected data
+  if(!isEmptyOrBlank(primarySeriesList[['uncorrected']]) && !isEmptyVar(primarySeriesList[['uncorrected']][['points']])) {
+    plot_object <- plotTimeSeries(plot_object, primarySeriesList[['uncorrected']], "uncorrected", 
+                                  timezone, getPrimaryPlotConfig, list(uvInfo[['label']], limsAndSides$ylims[['primary']], limsAndSides$sides[['primary']]), excludeZeroNegativeFlag)
```
lindsayplatt, line 190: confused about how `doLog` ever makes it to `getPrimaryPlotConfig`. It should be in the list, but I see labels, limits, and sides added. Is it stored in `limsAndSides$sides[['primary']]`?

zmoore-usgs: The `doLog` that's used in `getPrimaryPlotConfig` is calculated within `plotTimeSeries` (utils-plot, line 194). 

When `logPrimary` is calculated (line 200) that is used for the section further down that limits other items being added on that axis to only having positive values (line 222)

lindsayplatt: Ohhhh there we go. I even looked at that already!

lindsayplatt (approved):
Looks good :+1:

PR #639, opened 2017-04-18
AQCU 1071: SRS Rendering error

dpattermann-usgs:
Error was caused by null elements in the list. Fixed by using dplyr to bind the non-empty rows together.

Unsure if the import in the sig/rmarkdown files are necessary.
Feel free to give alternate implementations if you have a better way.


inst/sensorreading/sensorreading.Rmd
```diff
@@ -62,8 +62,11 @@ if(!identical(tbl, "The dataset requested is empty.")){
 ***
 
 ```{r,echo=FALSE,result='asis',comment=NA,warning=FALSE}
+library(dplyr)
```
lindsayplatt, line 65: you might not need this....you need either this one or the `importFrom` line but not both. Try deleting this and seeing if it still works.

zmoore-usgs (approved):

inst/sensorreading/sensorreading.Rmd
dpattermann-usgs: So, it looks like I need this (or something like it). I kind of wonder if I can put it somewhere else though. Does the importFrom line actually import anything or is it just used for documentation?

lindsayplatt: it works for functions specifically, so that function can use it, but since the function calls an .Rmd, it starts in a fresh environment. So keep `library(dplyr)` and remove `importFrom`

inst/sensorreading/sensorreading.Rmd
```diff
@@ -62,8 +62,11 @@ if(!identical(tbl, "The dataset requested is empty.")){
...
 if(!identical(tbl, "The dataset requested is empty.")){
-  quals <- as.data.frame(data$readings$qualifiers)
+  quals <- bind_rows(lapply(data$readings$qualifiers, function(qualRow) {
+                    as.data.frame(Filter(Negate(is.null), qualRow))
```
lindsayplatt: instead of `Negate`, we use `!is.null()` (I believe that's what Negate is doing)

lindsayplatt:
woo! :tada:

PR #640, opened 2017-04-18
AQCU-1070: DEV CORR Report Rendering Error

zmoore-usgs:


dpattermann-usgs (approved):
Should we at all consider having flexibleTimeParse handle vectors or just leave it and use apply functions when this crops up? Something that we should probably just keep an eye on. Code change looks fine though.

zmoore-usgs:
^ That's a good question, I'm not sure. @lindsaycarr?

lindsayplatt:
I'm pretty sure you can already do that: 

```r
x <- seq(as.Date("2010-10-01"), as.Date("2015-10-01"), by="month")
flexibleTimeParse(x, timezone="UTC")
```

zmoore-usgs:
@lindsaycarr The implementation using the sapply ended up significantly slowing down UV Hydros and DV Hydros so I looked up how to vectorize parse_date_time and went with that (leaving out the case for DVs that need to shift to noon). It seems to work now and all tests are passing so :crossed_fingers: . Let me know if you have any other suggestions / comments!

R/utils-time.R
```diff
@@ -4,6 +4,8 @@
 #' 
 #' @description A convienence function that will attempt to parse a (day
 #'   point-type) date, UTC time, or offset time extremes JSON.
+#'   NOTE: This will behave more accruately than before (where the first time in x set the format)
+#'   but does not handle cases where DVs and UVs are mixed in x
```
lindsayplatt: good to know :)

R/utils-time.R
```diff
@@ -13,18 +15,11 @@
 #' @importFrom lubridate hours
 #' @export
 flexibleTimeParse <- function(x, timezone, shiftTimeToNoon = TRUE) {
+  format_str <- c("Ymd HMOS z", "Ymd T* z*", "Ymd HMS")
```
lindsayplatt, line 16: still unsure about this - if x is length 5, this will use `format_str[1]` to parse the first date, then `format_str[2]` to parse the second, then `format_str[3]` for the third date, then `format_str[1]` again for the 4th date, and `format_str[2]` for the fifth date. I don't think this is the behavior we want. We want it to test which format is most accurate.....

Perhaps something similar to this:
```r
library(lubridate)
timezone <- "UTC"
shiftTimeToNoon <- TRUE

# example has DV first (should be formatted in third statement)
# T* version second (should be formatted in second statement)
# third is still a wrong format so should remain NA in final time vector
x <- c("2010-10-01", "2017-01-01T12:12:13Z", "2017-01-01 12:12:55 UTC")

# first if statement becomes:
format_str <- "Ymd HMOS z"
time <- parse_date_time(x,format_str, tz=timezone,quiet = TRUE)
which.NA.1 <- sapply(time, repgen:::isEmptyOrBlank)

# second if statement becomes:
format_str <- "Ymd T* z*"
time[which.NA.1] <- parse_date_time(x[which.NA.1],format_str, tz=timezone,quiet = TRUE)
which.NA.2 <- sapply(time, repgen:::isEmptyOrBlank)

# third if statement becomes:
format_str <- "Ymd"
time[which.NA.2] <- parse_date_time(x[which.NA.2],format_str, tz=timezone,quiet = TRUE)
if (shiftTimeToNoon) {
  time[which.NA.2] <- time[which.NA.2] + hours(12)
}

time
```

R/utils-time.R
```diff
@@ -4,7 +4,7 @@
 #' 
 #' @description A convienence function that will attempt to parse a (day
 #'   point-type) date, UTC time, or offset time extremes JSON.
-#' @param x A character vector of the date/time.
+#' @param x A vector of character vectors of the dates/times.
```
lindsayplatt: the prev documentation did make sense - "a vector of vectors" is a little more confusing. To be completely clear, I might say "A character vector of length one or greater of the dates/times."

R/utils-time.R
```diff
@@ -13,30 +13,21 @@
...
-  #first attempt utc
-  format_str <- "Ymd HMOS z"
   time <- parse_date_time(x,format_str, tz=timezone,quiet = TRUE)
   
-  #then attempt an offset time
-  if(isEmptyOrBlank(time)) {
-    format_str <- "Ymd T* z*"
-    time <- parse_date_time(x,format_str, tz=timezone, quiet = TRUE)
-  }
+  dvTimes <- x[which(is.na(time))]
```
lindsayplatt, line 20: since you extract DVs here and then put them back in with `time[which(is.na(time))] <- dv`, it would be simpler to just perform the operations on the NAs. Also, I think `isEmptyOrBlank` should still be used, but we want a specific T/F for each element so we use sapply. Then, you can eliminate the if structure. I think the entire function could become this:

```r
# Try these formats on each of the dates. It will try all formats given in 
# format_str. If none work, it will return NA.
format_str <- c("Ymd HMOS z", "Ymd T* z*", "Ymd HMS")
time <- parse_date_time(x,format_str, tz=timezone,quiet = TRUE)

#Handle DVs
#Try the DV format for dates that are still NA
#Add noon as the time if necessary
which.still.NA <- sapply(time, repgen:::isEmptyOrBlank)
format_str_DV <- "Ymd"
time[which.still.NA] <- parse_date_time(x[which.still.NA],format_str_DV, tz=timezone,quiet = TRUE)
if (shiftTimeToNoon) {
    time[which.still.NA] <- time[which.still.NA] + hours(12)
}

return(time)
```

lindsayplatt: If you think the sapply is slowing things down, we could just use `is.na`. The times should be returned as `NA` from the `parse_date_time` function, so that could be a sure thing. If that's the case, replace the sapply line with `which.still.NA <- is.na(time)`

I think that is good assumption considering the follow still produce NA:
```r
library(lubridate)
format_str <- c("Ymd HMOS z", "Ymd T* z*", "Ymd HMS")

# empty string
time <- parse_date_time("",format_str, tz="UTC",quiet = TRUE)
time
[1] NA

# NULL
time <- parse_date_time(NULL,format_str, tz="UTC",quiet = TRUE)
time
character(0)

# the vectorized check could be:
which.still.wrong <- is.na(time) & length(time) > 0

# though, that seems to be an issue when any are NULL because it doesn't keep it in the final variable
time <- parse_date_time(c("", NULL, "2017-01-01T12:12:13Z"),format_str, tz="UTC",quiet = TRUE)
time
[1] NA                        "2017-01-01 12:12:13 UTC"

# although neither does isEmptyOrBlank
# not sure if that's a scenario we need to prevent, but it might have to go at the beginning because you would end up returning a vector that is a different length of the original.
```

tests/testthat/test-correctionsataglance.R
```diff
@@ -70,9 +75,9 @@ test_that("calcStartSeq properly calculates the sequence of month start dates",
   expect_equal(length(startSeq4), 3)
 
   expect_equal(startSeq1, startDate1)
-  expect_equal(as.numeric(startSeq2), unlist(list(startDate1, monthDate2+startTime1, monthDate3+startTime1)))
+  expect_equal(as.numeric(startSeq2), unname(unlist(list(startDate1, monthDate2+startTime1, monthDate3+startTime1))))
   expect_equal(startSeq3, startDate2)
-  expect_equal(startSeq4, c(startDate2, monthDate2+startTime2, monthDate3+startTime2))
+  expect_equal(unname(startSeq4), unname(c(startDate2, monthDate2+startTime2, monthDate3+startTime2)))
```
lindsayplatt: why do these need to be unnamed?

zmoore-usgs: These needed to be unnamed in one of the earlier committed versions because it was returning named vectors with the input dates as the names. The `unname` calls could go away now I think

lindsayplatt (approved):
my only reservation is what will happen with that NULL case?

zmoore-usgs:
Tested in the NULL case, empty vector case, and empty list case and it returns `character(0)` for all of those

PR #646, opened 2017-04-26
Add up to 4 stat-derived TS plotting to 5YrGW reports.

kmschoep-usgs:


R/fiveyeargwsum-render.R
```diff
@@ -44,21 +44,28 @@ createfiveyeargwsumPlot <- function(reportObject){
...
-
+  
+  #Find the highest priority TS that has data
+  priorityTS <- list(stat1TimeSeries, stat2TimeSeries, stat3TimeSeries, stat4TimeSeries, stat1TimeSeriesEst, stat2TimeSeriesEst, stat3TimeSeriesEst, stat4TimeSeriesEst)
+  priorityTS <- priorityTS[unlist(lapply(priorityTS, function(ts){!isEmptyOrBlank(ts)}))][[1]]
+  
   #Get Additional Plot Data
   groundWaterLevels <- parseGroundWaterLevels(reportObject)
-  minMaxIVs <- parseMinMaxIVs(reportObject, timezone, statTimeSeries[['type']], invertedFlag, excludeMinMaxFlag, FALSE)
+  minMaxIVs <- parseMinMaxIVs(reportObject, timezone, stat1TimeSeries[['type']], invertedFlag, excludeMinMaxFlag, FALSE)
```
dpattermann-usgs: Just to check, do we want to use stat1TS or do we want to use priorityTS to get the type?

kmschoep-usgs: Excellent question!  I updated a lot of the 5Yr to behave like DV Hydro, and DV Hydro does use stat1TimesSeries.  That doesn't rule out DV Hydro from being incorrect, however.  I'll do more research on those two reports.

kmschoep-usgs: Good catch!  Turns out it still plots the right thing, but the label doesn't display the time series type, which looks weird.  This will have to be fixed in DV as well.

PR #647, opened 2017-04-27
add shift remarks to V Diagram table, reordered columns

kmschoep-usgs:


mhines-usgs:
Checking this out now :D

mhines-usgs (approved):

PR #649, opened 2017-05-01
add fourth stat-derived TS to DV Hydro report.

kmschoep-usgs:


R/dvhydrograph-render.R
```diff
@@ -35,17 +35,19 @@ createDVHydrographPlot <- function(reportObject){
...
+  stat4TimeSeriesEst <- parseTimeSeries(reportObject, 'fourthStatDerived', 'fourthStatDerivedLabel', timezone, estimated=TRUE, isDV=TRUE)
 
   #Validate Basic Plot Data
   if(all(isEmptyOrBlank(c(stat1TimeSeries, stat1TimeSeriesEst, stat2TimeSeries, stat2TimeSeriesEst, stat3TimeSeries, stat3TimeSeriesEst)))){
     return(NULL)
   }
 
   #Find the highest priority TS that has data
-  priorityTS <- list(stat1TimeSeries, stat2TimeSeries, stat3TimeSeries, stat1TimeSeriesEst, stat2TimeSeriesEst, stat3TimeSeriesEst)
+  priorityTS <- list(stat1TimeSeries, stat2TimeSeries, stat3TimeSeries, stat4TimeSeries, stat1TimeSeriesEst, stat2TimeSeriesEst, stat3TimeSeriesEst, stat3TimeSeriesEst)
```
lindsayplatt: you forgot to change the last one to stat4; there are two stat3 est

R/dvhydrograph-render.R
```diff
@@ -158,7 +164,8 @@ createDVHydrographRefPlot <- function(reportObject, series, descriptions) {
   series_number <- switch(series,
     'firstReferenceTimeSeries' = '1',
     'secondReferenceTimeSeries' = '2',
-    'thirdReferenceTimeSeries' = '3'
+    'thirdReferenceTimeSeries' = '3',
+    'thirdReferenceTimeSeries' = '4'
```
lindsayplatt: need to change to 'fourthReferenceTimeSeries'

R/dvhydrograph-render.R
```diff
@@ -95,16 +98,19 @@ createDVHydrographPlot <- function(reportObject){
...
+  plot_object <- plotTimeSeries(plot_object, stat4TimeSeriesEst, 'stat4TimeSeriesEst', timezone, getDVHydrographPlotConfig, list(yLabel=yLabel), excludeZeroNegativeFlag, isDV=TRUE)
   plot_object <- plotTimeSeries(plot_object, comparisonTimeSeries, 'comparisonTimeSeries', timezone, getDVHydrographPlotConfig, list(yLabel=yLabel), excludeZeroNegativeFlag, isDV=TRUE)
   plot_object <- plotTimeSeries(plot_object, comparisonTimeSeriesEst, 'comparisonTimeSeriesEst', timezone, getDVHydrographPlotConfig, list(yLabel=yLabel), excludeZeroNegativeFlag, isDV=TRUE)
 
   #Plot Other Items
   plot_object <- plotItem(plot_object, estimated1Edges, getDVHydrographPlotConfig, list(estimated1Edges, 'estimated1Edges'), isDV=TRUE)
   plot_object <- plotItem(plot_object, estimated2Edges, getDVHydrographPlotConfig, list(estimated2Edges, 'estimated2Edges'), isDV=TRUE)
   plot_object <- plotItem(plot_object, estimated3Edges, getDVHydrographPlotConfig, list(estimated3Edges, 'estimated3Edges'), isDV=TRUE)
+  plot_object <- plotItem(plot_object, estimated4Edges, getDVHydrographPlotConfig, list(estimated3Edges, 'estimated4Edges'), isDV=TRUE)
```
lindsayplatt: need `estimated4Edges` - there are two 3's

R/dvhydrograph-render.R
```diff
@@ -287,6 +300,11 @@ getDVHydrographPlotConfig <- function(plotItem, plotItemName, yLabel="", minMaxE
                            lty=ifelse(plotItem[['newSet']] == "est", 1, 6), col=ifelse(plotItem[['newSet']] == "est", "orange", "red3")),
                       styles$est_lines)
     ),
+    estimated4Edges = list(
+      arrows = append(list(x0=plotItem[['time']], x1=plotItem[['time']], y0=plotItem[['y0']], y1=plotItem[['y1']], 
+                           lty=ifelse(plotItem[['newSet']] == "est", 1, 7), col=ifelse(plotItem[['newSet']] == "est", "black", "red3")),
```
lindsayplatt: isn't the color defined in dvhydrograph-styles.R? I don't think we need to specify it in both places, right?

R/dvhydrograph-render.R
kmschoep-usgs: gah, thanks!

R/dvhydrograph-render.R
kmschoep-usgs: good catch, I actually need to delete the fourth Reference TS, since there are still only 3...

R/dvhydrograph-render.R
kmschoep-usgs: I'm not sure.  The "ifelse"s in there made me think it's some dynamic styling, but I could be wrong.

zmoore-usgs: Yeah in this spot we pulled the color out of styles because it is dynamic (creating the vertical lines connecting estimated/non-estimated series and choosing whether or not the line should be the estimated color or non-estimated color). If there is a clean way to move the color back into the styles that would be cool, but I don't think we came up with anything at the time this was originally added.

Do you have any ideas @lindsaycarr ?

zmoore-usgs (commented):
Looks good to me aside from the things that Lindsay mentioned!

R/dvhydrograph-render.R
lindsayplatt: I believe, `styles[[plotStylesName]]` would do the trick. That correctly has whether it's estimated and red3 or not estimated and black, correct?

lindsayplatt: Maybe I'm misinterpreting...

zmoore-usgs: Ah I see what you mean I think. So essentially we'd add 8 new entries to the styles:

estimated1EdgesEst(lty = 1, col = "blue", lwd=1, code=1, length = 0)
estimated1EdgesStat(lty = 2, col = "red1", lwd=1, code=1, length = 0)
estimated2EdgesEst(lty=1, col = "maroon", lwd=1, code=1, length = 0)
estimated2EdgesStat(lty=3, col = "red2", lwd=1, code=1, length = 0)
estimated3EdgesEst...
estimated3EdgesStat...
estimated4EdgesEst...
estimated4EdgesStat...

Then in the config function have it be something like
`estimated1Edges = list(
      arrows = append(list(x0=plotItem[['time']], x1=plotItem[['time']], y0=plotItem[['y0']], y1=plotItem[['y1']],
                           ), ifelse(plotItem[['newSet']] == "est", styles$estimated1EdgesEst, styles$estimated1EdgesStat))
    ),`

?

zmoore-usgs: @kmschoep-usgs Do you have any thoughts on this? @lindsaycarr and I just talked and think this is probably the cleaner way to do it ^

kmschoep-usgs: I can go ahead and implement that.

kmschoep-usgs: So, in styles, I have:

estimated1EdgesEst = list(lty=1, col = "blue", lwd=1, code=1, length = 0),
estimated1EdgesStat = list(lty=2, col = "red", lwd=1, code=1, length = 0),

and in render I have: 
```
estimated1Edges = list( 
      arrows = append(list(x0=plotItem[['time']], x1=plotItem[['time']], y0=plotItem[['y0']], y1=plotItem[['y1']],  
                      ifelse(plotItem[['newSet']] == "est", styles$estimated1EdgesEst, styles$estimated1EdgesStat)), styles$est_lines) 
    ),
```

but I get an error when running: 

Error in (function (x0, y0, x1 = x0, y1 = y0, length = 0.25, angle = 30,  : 
  invalid arrow head angle
 attempting to run: try({dvhydrograph(fromJSON("report-26e4c9bb-5389-4df2-9343-dd241b4cf8dd.json"), "kmschoep")},silent=TRUE)

zmoore-usgs: In the render function I think you should take that ifelse out of the `list()` call and replace `styles$est_lines` with it. The last 3 parameters in the styles entries come from styles$est_lines so that can be removed completely from the styles and its entry in the render function can be replaced with the ifelse.

kmschoep-usgs: this change:

```
estimated1Edges = list( 
      arrows = append(list(x0=plotItem[['time']], x1=plotItem[['time']], y0=plotItem[['y0']], y1=plotItem[['y1']] ), 
                      ifelse(plotItem[['newSet']] == "est", styles$estimated1EdgesEst, styles$estimated1EdgesStat)) 
      ),
```

resulted in the same invalid arrow head angle error.

zmoore-usgs: Hmm, I wonder if order of the parameters int he styles matters? This worked for me:

```
  estimated1Edges = list(
      arrows = append(list(x0=plotItem[['time']], x1=plotItem[['time']], y0=plotItem[['y0']], y1=plotItem[['y1']]),styles$est_lines)
    )
```

and in styles:

```
 est_lines1_est = list(lwd=1, code=1, length = 0, lty=1, col="blue"),
    est_lines1_est = list(lwd=1, code=1, length = 0, lty=2, col="red")
```

zmoore-usgs: Oh whoops, I forgot to change the styles entry in the render and now it doesn't work. Thinking back I actually think I remember @lindsaycarr mentioning something about `ifelse` not playing nice with lists?

lindsayplatt: Yeah, `ifelse` only likes returning one value.

This returns the first list item, so `list(2)` not `list(2,2,2)`.
```
ifelse(1==1, list(2,2,2), list(3,3,3))
```

lindsayplatt: :thinking: about the best way to do this

lindsayplatt: It's almost like `plotItemName` needs to be changed to `estimated1EdgesEst` or `estimated1EdgesStat` instead of just `estimated1Edges` prior to entering this function. It seems like the function grabbing the configs shouldn't have to deal with that kind of logic - it should happen upstream.

This might be getting to be a bigger task to clean that up. Could be a new ticket instead.

lindsayplatt: :point_up: that's the kind of stuff that bit us before. We had a bunch of logic going on really far downstream and it got very confusing.

lindsayplatt:
Let's make a new ticket to clean up the styling stuff.

PR #650, opened 2017-05-01
AQCU 1077: DV-Hydro Legend overlap

dpattermann-usgs:
Text change so DV-Hydro plots with less than 1 year of data don't have their legend overlap the dates for the plot.


R/dvhydrograph-render.R
```diff
@@ -398,8 +398,8 @@ plotDVHydroLegend <- function(plot_object, startDate, endDate, timezone, initial
   legend_offset <- ifelse(ncol==2, initialOffset+(0.025*leg_lines), initialOffset/2+(0.025*leg_lines))
   legend_offset <- legend_offset * modOffset
 
-  #If the time period is greater than 1 year additional x-axis labels are added so we must move the legend down further
-  legend_offset <- ifelse(as.period(interval(startDate, endDate, tzone = attr(startDate, timezone))) < lubridate::years(1), legend_offset+0.03, legend_offset+0.08)
+  #Add space to legend so that it doesn't overlap with dates.
+  legend_offset <- legend_offset + 0.8
```
zmoore-usgs: Does this not need to be dynamic anymore? I think it was dynamic because the extra labels that caused the overlap are only added on plot ranges of 1 year or greater and we didn't want the extra space to be there when it didn't need to be, but if this isn't an issue anymore then I think this looks good!

dpattermann-usgs: I can post some screenshots quick, but it looks like it doesn't need to change when it's 1 year or greater. The 1 year plots I have seen don't overlap and (with this change), neither do the less than 1 year? Unless the overlap is caused by how many items are in the legend?

zmoore-usgs: Gotcha, sounds good to me then! I'm not sure if the number of items in the legend would have anything to do with it or not. @lindsaycarr ?

zmoore-usgs (approved):

R/dvhydrograph-render.R
lindsayplatt: Number of legend items shouldn't make a difference - the offset defines the top of the legend and then everything else is pushed under it. 

We just discussed this - if the > 1 year plots look fine w/ 0.08, then this fix is good. If not, @dpattermann-usgs will change the offset values and clean up that if-else so it's not so giant

lindsayplatt: I'm a liar. It seems like the offset is offsetting the centroid of the legend.

```r
gs <- gsplot(mar=c(15,4,4,2)) %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    points(1,2,legend.name="one") %>% 
    legend(location="below", cex=0.8, ncol=2,
           legend.offset=1)
gs
```

dpattermann-usgs:
@lindsaycarr I haven't seen any instances where the legend overlaps with the graph using these parameters. The reference series overlap issue was because the initial offset for reference series was set to 0 (while the primary is set to 0.1). Could potentially consolidate to one variable if we won't need both. I'm going to try this with a comparison time series to triple check that this is enough.

PR #652, opened 2017-05-05
AQCU-1095: clean up getDVHydrographPlotConfig

mhines-usgs:
and move styles to dvhydrograph-styles


zmoore-usgs (changes requested):
This section of the styles is used to draw the vertical edges connecting the estimated and non-estimated periods of dvhydro and the logic is that if we are going from a non-estimated to an estimated series the vertical line should appear like it is part of the non-estimated series, and if we are going from estimated to non-estimated it should appear as part of the estimated series. Currently this information comes into the configuration function through the `plotItem[['newSet']]` parameter where `newSet` represents the series type we are transitioning to (`stat` or `est`). 

We still need the idea of the `ifelse` to exist somewhere in the code in order to differentiate the styles based on what series we are transitioning to, but @lindsaycarr and I discussed it and decided that it would be cleaner to not have that logic within this configuration function, but instead have it somewhere higher up in the code, so that was the main idea behind this ticket (though it wasn't super clear in the ticket).

lindsayplatt:
Sorry - lack of clarity in the ticket would be my fault. Thought I explained it all! 

The whole point is that we want to avoid logic down in these config functions -- those were the kinds of things we did prior to the refactor and partially the reason we had the refactor. The config function shouldn't need to know logic about the est --> nonest versus nonest --> est. It should get a style type and then return the configs. So, this ticket was meant to pull out that logic and move it up one function-level so that it wasn't embedded in the config function.

mhines-usgs:
No problem. I'm looking more closely now at the report generation to identify the right place. It feels like the dvhydro-data getEstimatedEdges function could be providing that information back to the dvhydro-render function.. ?

lindsayplatt:
So, `plotItem` function in `dvhydro-render` uses the config function and originally sets up the plotItemName. The `getDVHydrographPlotConfig` needs the newly added `plotItemName` for these edge scenarios. Logic in `dvhydro-data` instead of render makes sense, but I think the edges aren't calculated until `dvhydro-render` using `getEstimatedEdges`. 

```r
# getEstimatedEdges already has this functionality:
estimated1Edges[['newSet']]
[1] 'est' # OR returns "stat"

# we could just tack that on as the name to use for the styles list:
plotItemName <- paste('esimated1Edges', estimated1Edges[['newSet']])

# if newSet is NULL, the name should stay the same
plot_object <- plotItem(plot_object, estimated1Edges, getDVHydrographPlotConfig, list(estimated1Edges, plotItemName), isDV=TRUE)
```

lindsayplatt:
:point_up: that's not tested, just an idea that I was thinking might work.

mhines-usgs:
Argghghhhh, rewrote things and came up with the same error.

> Quitting from lines 53-65 (dvhydrograph.Rmd) 
> Error in switch(plotItemName, stat1TimeSeries = list(lines = append(list(x = x,  : 
>   EXPR must be a length 1 vector

I dont think it likes having multiple styles for the plotItem given to it.. but I'm not seeing how to change that since the edges are together in one object but they are mixed as far as style needs go.

mhines-usgs:
Working on separating the edges into two separate objects/edge type to see if it'll work that way..

mhines-usgs:
Error with last commit = 

> Error in append(list(x0 = plotItem[["time"]], x1 = plotItem[["time"]],  : 
>   argument "values" is missing, with no default

mhines-usgs:
Might have just figured the last error out.

mhines-usgs:
Eh, nope. still :bomb:

mhines-usgs:
Hahahah hooo lol. I had a parenthesis in the wrong spot. :(

zmoore-usgs (approved):

zmoore-usgs:
Tested locally and it looks good!

PR #653, opened 2017-05-11
AQCU-1089: V-Diagram- Times show in shift table not displaying UTC offset

zmoore-usgs:
Should be merged at the same time as: https://github.com/USGS-CIDA/AQCU/pull/695


R/utils-time.R
```diff
@@ -145,3 +145,20 @@ boundDate <- function(date, dateRange, padDays=1){
...
+#' Format Time as a label with the UTC Offset
+#'
+#' @description Formats a time datum as a text
+#' label including the UTC offset at the end.
+#' @param time The time datum to format as a text label
+formatUTCTimeLabel <- function(time){
+  formatted_time <- ""
+  
+  if(!isEmptyOrBlank(time)){
+    tzf <- format(as.POSIXct(time), "%z")
```
mhines-usgs, line 157: did you try this for 9999 or does it never encounter that because it's just the start date?

zmoore-usgs: It shouldn't ever encounter 9999 in this instance but it's possible that if we use that function elsewhere it might so it's probably a good idea to add a test for it, just in case. I'll do that now

mhines-usgs: ok, good idea

mhines-usgs: seems like it works ok when I manually tested, so great!

mhines-usgs (approved):

PR #659, opened 2017-05-19
Whisker based report building structure with example

thongsav-usgs:
See the README.md file in the PR for the overall process of creating a report.


R/utils-render.R
```diff
@@ -57,4 +70,185 @@ startRender <- function(data, author, reportName){
...
+#' 
+#' Peform the rendering of subfragments for a given report, should return a list of html fragments to 
+#' be available to inject into mustach templates
+#'
+#' @param reportData typed report json, the type should be the name of the report
+#' @return vector of rendered HTML fragments that will be included in the data available to the final report template. The fragments will be
+#' available to the templates as renderedFragments.html
+#' 
+#' @export
+renderCustomFragments <- function(reportData) UseMethod("renderCustomFragments")
```
jiwalker-usgs: I showed you S3 objects, but this project already has S4 in place, it is probably best not to mix the two.

thongsav-usgs: 👍 I'll make the conversion to S4.

R/02-templating-class.R
```diff
@@ -0,0 +1,26 @@
...
+#' 
+#' Peform the rendering of subfragments for a given report, should return a list of html fragments to 
+#' be available to inject into mustach templates
+#'
+#' @param reportData typed report json, the type should be the name of the report
+#' @return vector of rendered HTML fragments that will be included in the data available to the final report template. The fragments will be
+#' available to the templates as renderedFragments.html
+#' 
+#' @export
+renderCustomFragments <- function(reportData) 0
```
jiwalker-usgs, line 11: Is this what they suggest for stubbing in the function?

R/utils-render.R
```diff
@@ -27,16 +45,11 @@ cleanTempSpace <- function() {
...
 #' 
 startRender <- function(data, author, reportName){
-  data <- data 
-  
-  wd <- getwd()
-  tmp_folder_name <- paste0('tmp-', reportName)
+  warning("This rendering style is deprecated, convert to using whisker style and startTemplatedRender")
   
-  output_dir <- paste0(wd, "/", tmp_folder_name)
+  data <- data
```
jiwalker-usgs, line 50: does this actually do anything?

jiwalker-usgs (commented):
S4 is a bit goofy this way, which is partially why I prefer S3, but I think this looks alright.

zmoore-usgs (commented):
Looks good to me, just a couple of comments:

1. Not sure if any of the functions added to utils-render are really testable, but if they are we should probably add tests for them.

2. Since the JSON parameter names for the header are now in a shared header we should probably go back and make sure that we have those names consistent in the report choreographers because I'm not sure if they are. Perhaps it would be good to add some sort of shared function for generating the report metadata or a shared list of the required metadata parameters for every report so that we don't have inconsistencies across the reports.

R/utils-render.R
thongsav-usgs: I think it puts data in the env so it's available to rmarkdown. I never quite wrapped my head around why but this is going away soon.

thongsav-usgs:
@zmoore-usgs, so far the new functions are only tested via the integration test in test-renderReport, but I will look at adding unit tests.

The reportMetadata is definitely NOT consistent across reports, in the header you'll see there are template fragments for UV Hydro. We will do consolidation as we convert reports from rmarkdown to whisker.

R/02-templating-class.R
thongsav-usgs: Yeah, this was suggested as the base function. Without it the setMethod/setGeneric complains that there isn't a function yet defined.

PR #670, opened 2017-06-05
AQCU-1136: Fix for erroring/plotting on dvhydro when empty l…

mhines-usgs:
…ist elements are sent to plotItem function. Introduced with the est/stat edges work done a few weeks ago.

### DIFF
--- a/R/utils-plot.R
+++ b/R/utils-plot.R
@@ -217,6 +217,9 @@ plotTimeSeries <- function(plot_object, ts, name, timezone, configFunction, conf
 #' @param configFunctionParams list of params to call pall configFunctionWith
 #' @param isDV Whether or not the plot is a daily value plot (defulat: FALSE)
 plotItem <- function(plot_object, item, configFunction, configFunctionParams, isDV=FALSE){
+  #if item has list elements that are empty, filter them out to avoid error
+  item <- Filter(length, item)
+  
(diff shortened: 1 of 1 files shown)


zmoore-usgs (approved):
Seems like a good fix to me, though not 100% sure if this would have any unintended side-effects or not. I think @lindsaycarr is pretty busy right now so it's probably okay to go ahead and merge but might be a good idea to have her look at this at some point, just to be safe.

mhines-usgs:
Thanks, will do!

lindsayplatt:
Added it to my list for Friday!

mhines-usgs:
@lindsaycarr nice, thank you, wasn't sure if I should make an issue or just ask in standup, so THANKS! :pray:

lindsayplatt (commented):
It seems like it would be a good idea to add some tests for this so it doesn't pop up again. I'm going to make a few and do a PR soon.

lindsayplatt:
I'm starting to think that we need a slightly different approach. If `item` has any empty fields, it might cause an error downstream since fields in `item` are describing the actual plotting info like x and y values. If those are missing, it could balk somewhere else, right? Should we be skipping an item entirely if it has empty fields?

@mhines-usgs @zmoore-usgs

mhines-usgs:
thanks @lindsaycarr !!

PR #673, opened 2017-06-05
AQCU-1129: Fixed an issue where report sections with multiple tables …

zmoore-usgs:
…would not properly be marked as empty.


mhines-usgs (approved):
looks/works great.

PR #674, opened 2017-06-06
AQCU-1131: Fixed an issue where child dataframes with different…

zmoore-usgs:
…structures would cause the attachFullDataToSubFrame function to fail.


mhines-usgs:
:+1: on the tweak related to the error reported by Laura.

mhines-usgs:
pulling again to poke.

mhines-usgs (approved):

PR #675, opened 2017-06-07
AQCU-1143: changed sort order of extremes values before selecting val…

mhines-usgs:
…ue, need to select first occurrence whereas previously was selecting last


zmoore-usgs:
Testing this locally now

zmoore-usgs (approved):
This works well locally for me!

PR #678, opened 2017-06-08
Aqcu 1135 addition -- adding additional sorting

mhines-usgs:


R/timeseriessummary-data.R
```diff
@@ -159,6 +160,7 @@ parseTSSRatingCurves <- function(reportData, timezone){
     curves[['applicablePeriods']] <- lapply(curves[['applicablePeriods']], function(p){
       p[['startTime']] <- formatOpenDateLabel(flexibleTimeParse(p[['startTime']], timezone))
       p[['endTime']] <- formatOpenDateLabel(flexibleTimeParse(p[['endTime']], timezone))
+      p <- p[order(p[['startTime']]),]
```
zmoore-usgs: I think this sort will only make the order of the applicable periods of the current curve be in order, while the order of the curves themselves may not be. Each curve has a `startOfPeriod` parameter that may also need to be used for sorting outside of this lapply in order to sort the curves themselves.

R/timeseriessummary-data.R
```diff
@@ -85,6 +85,7 @@ parseTSSThresholds <- function(reportData, timezone){
     thresholds[['periods']] <- lapply(thresholds[['periods']], function(p){
       p[['startTime']] <- formatOpenDateLabel(flexibleTimeParse(p[['startTime']], timezone))
       p[['endTime']] <- formatOpenDateLabel(flexibleTimeParse(p[['endTime']], timezone))
+      p <- p[order(p[['startTime']]),]
```
zmoore-usgs: Reviewed this from the bottom up for some reason, but I think this will run into the same issue as the curves where the periods of the threshold will be sorted but the thresholds themselves will not be. 

Unfortunately thresholds don't have an overall periodStartTime like curves do which makes this one a bit trickier to sort. Perhaps you could use dplyr or some other method to add a periodStartTime value to each threshold which is equal to the start time of the earliest period of that threshold and then sort based on those? 

I'm not sure if that's the best way to do this or not. Perhaps @lindsaycarr would have some other suggestions tomorrow when she's available?

zmoore-usgs (commented):
Changes look good for the most part, just a couple of comments related to sorting the items that have multiple time periods per item (thresholds and rating curves).

R/timeseriessummary-data.R
mhines-usgs: yeah I wasn't really sure what made sense and if I should sort before getting into the lapply. I can try mocking up some data to verify and try sorting ahead of time. Thanks!

R/timeseriessummary-data.R
mhines-usgs: Ok, I can try that and if Lindsay has other ideas, I'll gladly adapt :D

mhines-usgs:
So fumbling a bit with this but this sometimes works but sometimes doesn't, depending on the threshold data.. 

```
#' Parse TSS Thresholds
#'
#' @description Given the full report JSON object reads the
#' thresholds and handles read errors.
#' @param reportObject the full report JSON object
#' @param timezone the timezone of the report
parseTSSThresholds <- function(reportData, timezone){
  thresholds <- tryCatch({
    readThresholds(reportData)
  }, error=function(e){
    warning(paste("Returning NULL for TSS thresholds. Error:", e))
    return(list())
  })
  
  if(!isEmptyOrBlank(thresholds)){
    thresholds <- mutate(thresholds, sortTime = sapply(thresholds$periods, "[[", "startTime"))
    thresholds <- thresholds[order(thresholds[['sortTime']]),]
    thresholds[['periods']] <- lapply(thresholds[['periods']], function(p){
      p[['startTime']] <- formatOpenDateLabel(flexibleTimeParse(p[['startTime']], timezone))
      p[['endTime']] <- formatOpenDateLabel(flexibleTimeParse(p[['endTime']], timezone))
      return(p)
    })
    
    thresholds <- attachFullDataToSubFrame(thresholds, 'periods')
  }
  
  return(thresholds)
}
```

zmoore-usgs:
Hmm perhaps if you did the thresholds sort time mutation and ordering after the ordering within the periods? It might be that the first period isn't necessarily the earliest so when you pull it out to sort by it you might not end up sorting by the earliest? So if you add back the sorting of the periods within the lapply and then do the outside threshold sorting and mutating after that it could help?

The other thing is that there are situations when the threshold could have multiple periods and I'm not totally sure how that would end up with your sapply? I'm not too familiar with how sapply works so I could be wrong but that might end up giving you a list of all of the start times for the sortTime and then the sort after that might be sorting the overall thresholds based on a list for within each row and I'm not sure how that works

mhines-usgs:
Yeah, still trying to wrap my head around the variations across the test datasets. Trying something else now!

mhines-usgs:
maybe I'm crazy, but I just used this data for thresholds, and these seem to be sorting without me doing anything already.. 
```
 "thresholds": [
        {
            "severity": "0.0",
            "name": "Operational limit (maximum) Equipment 2010",
            "description": "Values above a specified threshold limit",
            "periods": [
                {
                    "startTime": "2015-10-07T00:00:00-05:00",
                    "endTime": "9999-12-31T23:59:59.9999999Z",
                    "appliedTime": "2017-02-02T21:27:53.3106Z",
                    "referenceValue": 19.2,
                    "secondaryReferenceValue": 17.28,
                    "suppressData": true
                }
            ],
            "referenceCode": "Operational limit - high-Public",
            "type": "ThresholdAbove"
        },
        {
            "severity": "0.0",
            "name": "Operational limit (minimum) DCP offset 2016",
            "description": "Values below a specified threshold limit",
            "periods": [
                {
                    "startTime": "2015-10-02T00:00:00-05:00",
                    "endTime": "9999-12-31T23:59:59.9999999Z",
                    "appliedTime": "2017-02-02T21:27:53.1881899Z",
                    "referenceValue": -0.07,
                    "secondaryReferenceValue": 0.13,
                    "suppressData": true
                },
                {
                    "startTime": "2015-10-05T00:00:00-05:00",
                    "endTime": "9999-12-31T23:59:59.9999999Z",
                    "appliedTime": "2017-02-02T21:27:53.1881899Z",
                    "referenceValue": -0.07,
                    "secondaryReferenceValue": 0.13,
                    "suppressData": true
                },
                {
                    "startTime": "2015-10-06T00:00:00-05:00",
                    "endTime": "9999-12-31T23:59:59.9999999Z",
                    "appliedTime": "2017-02-02T21:27:53.1881899Z",
                    "referenceValue": -0.07,
                    "secondaryReferenceValue": 0.13,
                    "suppressData": true
                }
            ],
            "referenceCode": "Operational limit - low-Public",
            "type": "ThresholdBelow"
        },
		{
            "severity": "0.0",
            "name": "Operational limit (minimum) DCP offset 2016",
            "description": "Values below a specified threshold limit",
            "periods": [
                {
                    "startTime": "2015-10-04T00:00:00-05:00",
                    "endTime": "9999-12-31T23:59:59.9999999Z",
                    "appliedTime": "2017-02-02T21:27:53.1881899Z",
                    "referenceValue": -0.07,
                    "secondaryReferenceValue": 0.13,
                    "suppressData": true
                }
            ],
            "referenceCode": "Operational limit - low-Public",
            "type": "ThresholdBelow"
        },
		{
            "severity": "0.0",
            "name": "Operational limit (minimum) DCP offset 2016",
            "description": "Values below a specified threshold limit",
            "periods": [
                {
                    "startTime": "2015-10-03T00:00:00-05:00",
                    "endTime": "9999-12-31T23:59:59.9999999Z",
                    "appliedTime": "2017-02-02T21:27:53.1881899Z",
                    "referenceValue": -0.07,
                    "secondaryReferenceValue": 0.13,
                    "suppressData": true
                }
            ],
            "referenceCode": "Operational limit - low-Public",
            "type": "ThresholdBelow"
        },
		{
            "severity": "0.0",
            "name": "Operational limit (minimum) DCP offset 2016",
            "description": "Values below a specified threshold limit",
            "periods": [
                {
                    "startTime": "2015-10-01T00:00:00-05:00",
                    "endTime": "9999-12-31T23:59:59.9999999Z",
                    "appliedTime": "2017-02-02T21:27:53.1881899Z",
                    "referenceValue": -0.07,
                    "secondaryReferenceValue": 0.13,
                    "suppressData": true
                }
            ],
            "referenceCode": "Operational limit - low-Public",
            "type": "ThresholdBelow"
        }
    ],
```
image

zmoore-usgs (approved):
Huh yeah it certainly seems to be working! Not totally sure where that sorting is happening, but evidently it is at some point! I'd say this is good to go then!

mhines-usgs:
Thanks.. yeah. I might test the other sections when I'm bored to see if they also don't need any explicit sorts. Probably some other function handled it here.

R/timeseriessummary-data.R
```diff
@@ -155,7 +155,7 @@ parseTSSRatingCurves <- function(reportData, timezone){
   
   if(!isEmptyOrBlank(curves)){
     colnames(curves)[which(colnames(curves) == 'remarks')] <- "curveRemarks"
-    
+    curves <- curves[order(curves[['startOfPeriod']]),]
```
lindsayplatt, line 158: In the first test for this, the class of `curves[['startOfPeriod']]` is character. So using `order` could be a problem. It appears to do it correctly, but makes me nervous that it doesn't throw a warning or something. 

`order(c("2010-10-01", "2011-10-01", "2010-10-02"))`

lindsayplatt (commented):
Using `order` works well! You could also use dplyr's `arrange` function and eliminate all of the parens and brackets.

`shifts <- arrange(shifts, applicableStartDateTime)`

lindsayplatt:
It would be worthwhile to throw in tests for the `parseTSS...` functions for ordering.

mhines-usgs:
Yeah I think the check to see if the dates are open ended (formatOpenDateLabel) transforms the dates into characters. I'll look at adding some tests. Thanks! Would you think arrange is the better solution here instead of order?

zmoore-usgs:
Yeah that function transforms them to characters because if they are open-ended it needs to replace them with the word "Open" and I was getting errors when the data frame was mixed with dates and strings

lindsayplatt:
@mhines-usgs I don't think it's necessary. `dplyr` can be faster depending on how large the data frames are because it's running C (C++ ?) under the hood.  In this case, it's mostly a style thing.

lindsayplatt:
@zmoore-usgs yeah, you can't have a column with mixed values - it's a vector so it needs one class. I guess ordering by characters should work, but just feels strange. Could the conversion to character happen after the ordering?

zmoore-usgs:
It could yeah, it just needs to happen before the call to formatOpenDateLabel

mhines-usgs:
Yep, I can move the calls above that part. No problem!

mhines-usgs:
adds tests and moves calls up above: https://github.com/USGS-R/repgen/pull/690

PR #681, opened 2017-06-09
AQCU-1145: Add correct reportMetadata.timeseriesLabel field for 

kmschoep-usgs:
derivation chain report

### DIFF
--- a/inst/templates/common/header.mustache
+++ b/inst/templates/common/header.mustache
@@ -10,7 +10,9 @@
 <div class="header-labels">
 <div class="header-col-left">
 <p><strong>Location:</strong> {{reportMetadata.stationId}} - {{reportMetadata.stationName}}</p>
+<!-- Only DVHydro and 5YR GW use primarySeriesLabel -->
 {{#reportMetadata.primarySeriesLabel}}<p><strong>Primary Time Series:</strong> {{reportMetadata.primarySeriesLabel}}</p>{{/reportMetadata.primarySeriesLabel}}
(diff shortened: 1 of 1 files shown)


zmoore-usgs (approved):
Might be good to eventually document which reports use which parameter name, though that's outside of the scope of this ticket.

mhines-usgs (approved):
:+1:

inst/templates/common/header.mustache
```diff
@@ -10,7 +10,9 @@
 <div class="header-labels">
 <div class="header-col-left">
 <p><strong>Location:</strong> {{reportMetadata.stationId}} - {{reportMetadata.stationName}}</p>
+<!-- Only DVHydro and 5YR GW use primarySeriesLabel -->
 {{#reportMetadata.primarySeriesLabel}}<p><strong>Primary Time Series:</strong> {{reportMetadata.primarySeriesLabel}}</p>{{/reportMetadata.primarySeriesLabel}}
+{{#reportMetadata.timeseriesLabel}}<p><strong>Primary Time Series:</strong> {{reportMetadata.timeseriesLabel}}</p>{{/reportMetadata.timeseriesLabel}}
```
thongsav-usgs, line 15: Instead of adding to header I'd fix on the java side to be consistent with other reports. I think "primarySeriesLabel" is the most widely used.

thongsav-usgs (changes requested):

inst/templates/common/header.mustache
kmschoep-usgs: It's all over the place: DvHydro uses primarySeriesLabel, and SRS, SVP and DC use timeseriesLabel, Extremes uses primaryLabel, corrections and UvHydro uses primaryParameter.  I can change derivation chain, but if we go to whisker templates for all reports, a lot of the other reports will need to be changed.

PR #687, opened 2017-06-09
AQCU-1149: Add note on CORR and TSS reports about end dates

zmoore-usgs:


inst/correctionsataglance/correctionsataglance.Rmd
```diff
@@ -36,6 +36,8 @@ cat(getLogo())
 
 ***
 
+Note: If the displayed end date is in the future, it's application is actually open-ended.
```
mhines-usgs: grammar nerd comment, I think it's should be its

inst/templates/timeseriessummary/body.mustache
```diff
@@ -1,5 +1,11 @@
 <div id="tss-report">
 
+  <div class="api-notice">
+    Note: If the displayed end date is in the future, it's application is actually open-ended.
```
mhines-usgs: it's -> its

mhines-usgs (approved):
Approve with minor edit!

zmoore-usgs:
Good catch!

PR #690, opened 2017-06-09
AQCU-1135-addition2: moving sort above open date check which converts…

mhines-usgs:
… dates to characters; adding tests for all the things


zmoore-usgs (approved):

lindsayplatt (commented):
Two questions: 

1. Do we not need to change anything for `parseTSSRatingCurves` ? That's where I first noticed the character dates.
2. Is there anyway to share or cut down some of that example JSON in the tests?

mhines-usgs:
@lindsaycarr checking out the curves now, but that was already sorting ahead of going into the 'check for open date' function. Also will trim down the json as much as possible for the tests. thanks!

mhines-usgs:
Interesting, they're characters already coming into the function. They are character-ified in the fetch rating curve function, actually. I guess if it's not making a difference in the sort/work at this point, it's OK..

PR #692, opened 2017-06-12
Aqcu 1144: format date/time in period section of header of new templated reports

mhines-usgs:


R/utils-render.R
```diff
@@ -110,6 +110,10 @@ startTemplatedRender <- function(reportJson, author){
   #call custom render function for the report, it should turn a list of HTML fragments that will available to the templates
   templateData[["renderedFragments"]] <- renderCustomFragments(reportJson)
   
+  #adding appropriately formatted start and end dates to reportJSON
+  templateData[['reportMetadata']][['displayStartDate']] <- format(flexibleTimeParse(templateData[['reportMetadata']][['startDate']], templateData[['reportMetadata']][['timezone']]), "%Y-%m-%d")
```
mhines-usgs, line 114: instead of changing startDate and endDate here, because I'm not sure if other downstream functions require those to include the times, I created a new displayStartDate and displayEndDate

inst/templates/common/header.mustache
```diff
@@ -22,7 +22,7 @@
 {{#reportMetadata.comparisonSeries}}<strong>Comparison Time Series:</strong> {{reportMetadata.comparisonSeries}}{{/reportMetadata.comparisonSeries}}
 </div>
 <div class="header-col-right">
-<p><strong>Period:</strong> {{reportMetadata.startDate}} to {{reportMetadata.endDate}}</p>
+<p><strong>Period:</strong> {{reportMetadata.displayStartDate}} to {{reportMetadata.displayEndDate}}</p>
```
mhines-usgs, line 25: note that just as before my changes, for derivation chain, this appears as just Period: to

if it makes sense to do a check here for whether or not the values are empty and whether or not to show this Period section at all, I can try to do that as a part of this task since this didn't take me very long.

R/utils-render.R
zmoore-usgs: From my understanding at this point everything in templateData is used strictly for rendering directly into the HTML so it should be okay to directly modify the startDate and endDate fields. But I'm not absolutely sure that is the case ( @thongsav-usgs ?) and I'm also not against putting it into a new field in order to keep the originally formatted data in case it does need to be rendered into another location or something.

inst/templates/common/header.mustache
zmoore-usgs: I'm pretty sure there's a ticket to fix the period on the DC report so I think it's safe to assume that the period is mandatory for every report so I'm not sure if a check is really necessary. If the period is not present it's a data error so it should be okay for it to render a bit strangely to make it obvious that there is an error.

zmoore-usgs (approved):

PR #693, opened 2017-06-13
AQCU-1125: adding legend to derivation chain report including the pro…

mhines-usgs:
…cessor types


R/utils-render.R
```diff
@@ -98,6 +98,9 @@ startTemplatedRender <- function(reportJson, author){
   #load header image
   templateData[["usgsLogoBase64"]] <- loadUsgsLogoBase64()
   
+  #load symbology processor images for derivation chain report
```
thongsav-usgs: This should not be in the shared global function. There is a place for it in derivationchain-data.R.

R/utils-render.R
```diff
@@ -194,6 +197,24 @@ loadUsgsLogoBase64 <- function() {
   return(base64encode(system.file('shared','usgs_logo.jpg', package = 'repgen')))
 }
 
+#' load legend symbology for derivation chain report
+#' @description will load the legend icons needed and convert to base64 text
+#' @return base64 string representation of legend images
+#' @importFrom base64enc base64encode
+loadLegendSymbologyDerivationChain <- function() {
```
thongsav-usgs: I'd also move this function to derivationchain-data.R

inst/templates/derivationchain/body.mustache
```diff
@@ -1,8 +1,20 @@
 <div id="cy"></div>
 <br/>
 <p><i>Time-Series processor icon images courtesy of Aquatic Informatics.</i></p>
+<div id="legendContainer"><strong>Legend</strong><br/><br/>
+<span style="text-decoration:underline">Processor Types:</span> <br/>
+<img src="data:image/jpeg;base64,{{derivationChainSymbols.ratingModelDerived}}" title="Rating derived" alt="Rating derived"> Rating derived<br/>
```
thongsav-usgs: Once you move the functions to derivationchain-data.R, you should be able to access the same information with reportData.derivationChainSymbols.ratingModelDerived

thongsav-usgs (changes requested):

R/utils-render.R
mhines-usgs: :+1:

R/utils-render.R
mhines-usgs: duh, woops. :+1:

inst/templates/derivationchain/body.mustache
mhines-usgs: hah, i missed this but you're correct :) :+1:

thongsav-usgs (approved):

PR #704, opened 2017-06-15
AQCU-1146: for templated reports where they do not have startDate or …

mhines-usgs:
…endDate in their reportMetadata, display "Period: Dynamic based on selection" instead of "Period: 2016-05-01 to 2017-05-30" for example

### DIFF
--- a/R/utils-render.R
+++ b/R/utils-render.R
@@ -112,8 +112,9 @@ startTemplatedRender <- function(reportJson, author){
   
   #adding appropriately formatted start and end dates to reportJSON
   if (anyDataExist(reportJson[['reportMetadata']][['startDate']]) && anyDataExist(reportJson[['reportMetadata']][['endDate']])) {
-    templateData[['reportMetadata']][['displayStartDate']] <- format(as.Date(fetchReportMetadataField(reportJson,'startDate')), "%Y-%m-%d")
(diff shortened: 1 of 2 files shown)


R/utils-render.R
```diff
@@ -112,8 +112,9 @@ startTemplatedRender <- function(reportJson, author){
   
   #adding appropriately formatted start and end dates to reportJSON
   if (anyDataExist(reportJson[['reportMetadata']][['startDate']]) && anyDataExist(reportJson[['reportMetadata']][['endDate']])) {
-    templateData[['reportMetadata']][['displayStartDate']] <- format(as.Date(fetchReportMetadataField(reportJson,'startDate')), "%Y-%m-%d")
-    templateData[['reportMetadata']][['displayEndDate']] <- format(as.Date(fetchReportMetadataField(reportJson,'endDate')), "%Y-%m-%d")
+    templateData[['reportMetadata']][['periodDisplay']] <- paste(format(as.Date(fetchReportMetadataField(reportJson,'startDate')), "%Y-%m-%d")," to ", format(as.Date(fetchReportMetadataField(reportJson,'endDate')), "%Y-%m-%d"))
+  } else {
```
zmoore-usgs, line 116: This logic may need to change in the future if we ever have another report that also doesn't include a time period but I don't see that happening any time soon so I think this is good.

zmoore-usgs (approved):

R/utils-render.R
mhines-usgs: Yeah, we might need to rethink it later, agreed.

PR #707, opened 2017-06-15
Aqcu 1153: Add parameters data to correction type field 

mhines-usgs:


mhines-usgs:
We probably don't want to merge this until we resolve AQCU-1158

zmoore-usgs (commented):
Looking at the example JSON that Laura provided I'm thinking it's likely that this might need some additional processing before outputting it to get it to look the way Laura wants (I'm not sure how Whisker handles outputting a list of data.frames which is what I think the "parameters" will likely get parsed into). 

I would guess the JSON that gets sent to repgen will end up looking very similar to the parameters JSON in Laura's example so you could probably use that for testing / working on this if you wanted (just pull out the parameters JSON from the correction entries from AQ and plug it into the individual correction entries in the JSON sent to repgen)

mhines-usgs:
Good idea, I'll fake her details into the json I grabbed from the server. Thanks for the suggestions!

R/timeseriessummary-data.R
```diff
@@ -388,4 +389,86 @@ parseTSSGapTolerances <- function(reportData, timezone){
   }
   
   return(gapTolerances)
-}
\ No newline at end of file
+}
+
+#' Format TSS Correction Parameters
+#'
```
mhines-usgs: oops, workin on this too :)

R/timeseriessummary-data.R
```diff
@@ -388,4 +389,86 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+#' Format TSS Correction Parameters
+#' @description formats the parameters so they are prettier to print
+#' @param corrections the corrections for the report
+#' @param timezone the timezone to parse data into
+#' @return corrections with formattedParameters field to print in report
+formatCorrectionParameters <- function(corrections, timezone) {
+  
+  corrections[["formattedParameters"]] <- ""
+  
+  if(!isEmptyOrBlank(corrections) && nrow(corrections)>0) {
```
lindsayplatt: `isEmptyVar` might be more robust than `nrow() > 0`

mhines-usgs: Thanks, sometimes I have a hard time figuring out which of the validation utils to use ;)! I like them all tho!

R/timeseriessummary-data.R
```diff
@@ -388,4 +389,86 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+    
+    for (i in seq(nrow(corrections))) {
```
lindsayplatt: thinking we can clean up this giant for loop...brainstorming now

mhines-usgs: Yeah, I redid this like three times, but the data come through oddly and this seemed to work alright ...

R/timeseriessummary-data.R
```diff
@@ -388,4 +389,86 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+#' @param corrections the corrections for the report
+#' @param timezone the timezone to parse data into
+#' @return corrections with formattedParameters field to print in report
+formatCorrectionParameters <- function(corrections, timezone) {
+  
+  corrections[["formattedParameters"]] <- ""
+  
+  if(!isEmptyOrBlank(corrections) && !isEmptyVar(corrections)) {
+    
+    for (i in seq(nrow(corrections))) {
```
lindsayplatt: bleh can't figure out how to vectorize this because the format is so funky. 

However, I think we should get rid of all those if-else statments and separate this out into many smaller, formatting functions. Then we can use a switch statement. This would be the only thing inside the for loop.

```
corrections[i, "formattedParameters"] <- 
        switch(corrections[i, 'type'],
               "Offset" = formatCorrectionsParamOffset(corrections[i, 'parameters'], timezone),
               "Drift" = formatCorrectionsParamDrift(corrections[i, 'parameters'], timezone),
               "SinglePoint" = formatCorrectionsParamSinglePoint(corrections[i, 'parameters'], timezone),
               "USGSMultiPoint" = formatCorrectionsParamUSGSMultiPoint(corrections[i, 'parameters'], timezone),
               "AdjustableTrim" = formatCorrectionsParamAdjustableTrim(corrections[i, 'parameters'], timezone),
               "FillGaps" = formatCorrectionsParamFillGaps(corrections[i, 'parameters'], timezone),
               "Deviation" = formatCorrectionsParamDeviation(corrections[i, 'parameters'], timezone),
               corrections[i, 'type']) #anything that doesn't match the previous words will get this as output
```

lindsayplatt: then we can write a few simple tests for each one of those - sorry, I know this creates a lot more work. If you don't think it's really adding anything, don't worry about it. I just get intimated by giant if-else kind of functions

mhines-usgs: No problem, I don't think this adds too much more work. I just hope when AQCU-1158 gets worked out, the JSON we mocked up holds true.  I'll work on it in the meantime. Thanks!

lindsayplatt: we could make this super clean if parameters wasn't a nested dataframe and was just the respective columns, e.g. `parmOffset` is a column of `corrections` instead of `offset` being a column of `parameters` within `corrections`. We could eliminate the whole for loop.

```
corrections_formatted <- corrections %>% 
  rowwise() %>% 
  mutate(timezone = timezone) %>% 
  mutate(
    formattedParameters = switch(corrections[i, 'type'],
                                 "Offset" = formatCorrectionsParamOffset(corrections[i, 'parameters'], timezone),
                                 "Drift" = formatCorrectionsParamDrift(corrections[i, 'parameters'], timezone),
                                 "SinglePoint" = formatCorrectionsParamSinglePoint(corrections[i, 'parameters'], timezone),
                                 "USGSMultiPoint" = formatCorrectionsParamUSGSMultiPoint(corrections[i, 'parameters'], timezone),
                                 "AdjustableTrim" = formatCorrectionsParamAdjustableTrim(corrections[i, 'parameters'], timezone),
                                 "FillGaps" = formatCorrectionsParamFillGaps(corrections[i, 'parameters'], timezone),
                                 "Deviation" = formatCorrectionsParamDeviation(corrections[i, 'parameters'], timezone),
                                 corrections[i, 'type'])) %>% 
  ungroup()
```

mhines-usgs: Yeah, I'm not entirely sure if the format will stay as it is, that's the format we see it coming through as from Aquarius, but there is a bug right now in our AQCU java app that pumps out the json so I just found the location where the parameters were showing up (empty), added Laura's data from Aquarius, and then made Repgen work... but maybe we can change how it works on the java end to support this.

lindsayplatt: Wait, what if we just unnested it and then the entire `formatCorrectionParameters` function would just contain:

```r
# unnesting parameters
parms <- corrections$parameters
corrections$parameters <- NULL

# adding params df as columns to corrections df
corrections <- cbind(corrections, parms)

# adding the formatted parameters as a column
corrections_formatted <- corrections %>% 
  rowwise() %>% 
  mutate(timezone = timezone) %>% 
  mutate(
    formattedParameters = switch(type,
                                 "Offset" = formatCorrectionsParamOffset(offset, tz),
                                 "Drift" = formatCorrectionsParamDrift(driftPoints, tz),
                                 "SinglePoint" = formatCorrectionsParamSinglePoint(value, tz),
                                 "USGSMultiPoint" = formatCorrectionsParamUSGSMultiPoint(startShiftPoints, endShiftPoints, usgsType, tz),
                                 "AdjustableTrim" = formatCorrectionsParamAdjustableTrim(upperThresholdPoints, tz),
                                 "FillGaps" = formatCorrectionsParamFillGaps(resamplePeriod, gapLimit, tz),
                                 "Deviation" = formatCorrectionsParamDeviation(deviationValue, deviationType, tz),
                                 type)) %>% 
  ungroup()
```

lindsayplatt:
wooo! You might need to add those dummy variables `timezone <- '.dplyr'` because of the weird dplyr NSE notes that pop-up in check. H

mhines-usgs:
Working on new tests to take advantage of these changes.

mhines-usgs:
Found a bug in our logic while working on tests, where the parameter types that don't have any parameter fields (where parameters are empty) (DeleteRegion, CopyPaste, Freehand, and RevertToRaw) on their own causes this to error. Working on a fix.

lindsayplatt:
@mhines-usgs :+1:  also, I just merged Zack's PR, so you might have to do some conflict merging.

R/timeseriessummary-data.R
```diff
@@ -413,8 +413,14 @@ unNestCorrectionParameters <- function(corrections, timezone) {
   deviationType <- ".dplyr"
   windowSizeInMinutes <- ".dplyr"
   ungroup <- ".dplyr"
+  
+  if (corrections[["type"]]=="CopyPaste" || corrections[["type"]]=="Freehand" || corrections[["type"]]=="DeleteRegion" || corrections["type"]=="RevertToRaw") {
+    corrections[["parameters.dummy"]] <- ""
```
mhines-usgs: @lindsaycarr how bad of a taste does this leave in your mouth? the problem is there aren't any parameters.X fields in these types of corrections, so without this dummy, the cbind fails.

lindsayplatt: I would do an if-else kind of thing down by cbind instead. Then corrections would remain the same if params was empty.

```r
if(!isEmptyOrBlank(params)){
    corrections <- cbind(corrections, params)
} 
```

R/timeseriessummary-data.R
```diff
@@ -525,14 +525,12 @@ unNestCorrectionParameters <- function(corrections, timezone) {
...
   
-  if (corrections[["type"]]=="CopyPaste" || corrections[["type"]]=="Freehand" || corrections[["type"]]=="DeleteRegion" || corrections["type"]=="RevertToRaw") {
-    corrections[["parameters.dummy"]] <- ""
-  }
-  
   params <- corrections$parameters
   corrections$parameters <- NULL
   
-  corrections <- cbind(corrections, params)
+  if (length(params)>0) {
```
lindsayplatt: I think `isEmptyVar` is what you are looking for - but this works, too

mhines-usgs:
grr, I'm having trouble sorting out changes now that this has been open so long, going to re-do the PR :/

lindsayplatt:
gah sorry!

mhines-usgs:
Nah, it's my own inability to figure out what I did wrong :( you have only been helpful!!

PR #708, opened 2017-06-19
AQCU-1151 + new class repgendate + fixing broken tests

lindsayplatt:
Addressing AQCU-1151 required creating a new class so that we could add new print methods to get the `00:00:00` to show up. With the new class, we needed new helper functions associated with it. 

Added tests for these new functions, plus updated tests that dealt with dates. If they used `expect_equal` it no longer worked because there was an additional class attached (so used expect_true & `==`) or because the dates now had "00:00:00" in them. These should all be fixed now.

Running `Check` has apparently been letting broken tests through because code above the actual tests in the file failed, but didn't cause the tests themselves to fail (it most likely exited and never made it to the tests). Running `Test Package` (aka CTRL + SHIFT + T) made these obvious. They should be fixed now.


zmoore-usgs (approved):
Code changes look good to me. Pulled down and tested locally and all tests passed in both check() and test()

lindsayplatt:
Yay!

PR #709, opened 2017-06-22
AQCU-1159: Add "Time Series Details" section to report

zmoore-usgs:
Should be merged with https://github.com/USGS-CIDA/AQCU/pull/738


zmoore-usgs:
I'm not sure if I got all of the new stuff covered with my tests so if the coverage goes down significantly this PR and the corresponding AQCU PR can wait on merging and I can work on upping the test coverage when I get back.

zmoore-usgs:
I also attached some example JSON that would be produced by the corresponding AQCU PR to the JIRA ticket (also timeseriessummary-example5.json in the extdata/timeseriessummary directory in Repgen)

R/timeseriessummary-data.R
```diff
@@ -44,7 +44,14 @@ parseCustomDataElementsForTemplateForTimeSeriesSummary <- function(reportData) {
...
   approvalsTable <- formatDataTable(parseTSSApprovals(reportData, timezone))
   
+  tsDetailsTable <- list()
+  tsDetails <- constructTSDetails(reportData, timezone)
+  tsDetailsTable[['tsAttrs']] <- formatDataTable(tsDetails[['tsAttrs']])
+  tsDetailsTable[['tsExtAttrs']] <- formatDataTable(tsDetails[['tsExtAttrs']])
+  addChangeNote <- tsDetails[['changeNote']]
+  
   return(list(
+      tsDetails = list(hasData=TRUE, data=tsDetailsTable, addChangeNote=addChangeNote),
```
lindsayplatt, line 54: Wouldn't you want to follow the same pattern and do `hasData = !isEmptyOrBlank(tsDetailsTable)` ?

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+  methodData <- parseTSSMethods(reportData, timezone)
+  itData <- parseTSSInterpolationTypes(reportData, timezone)
+  processorData <- parseTSSProcessors(reportData, timezone)
+  changeNote <- FALSE
+  
+  if(!is.null(metadata) && !isEmptyOrBlank(metadata)){
+    #Time Series Attributes
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Label", value=metadata[['identifier']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Parameter", value=metadata[['parameter']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Units", value=metadata[['unit']], indent=8, stringsAsFactors = FALSE))
```
lindsayplatt: I was going to suggest to do this in one step rather than 3 different rbinds, but I actually like how clean your implementation is even if it's lengthier. It's easier to see what value goes with what label. 

```r
tsAttrs <- rbind(tsAttrs,
                 data.frame(label=c("Label", "Parameter", "Units"), 
                            value=c(metadata[['identifier']], metadata[['parameter']], 
                                    metadata[['unit']]), 
                            indent=8, 
                            stringsAsFactors = FALSE)
```

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+  tsExtAttrs <- data.frame(stringsAsFactors = FALSE)
+  
+  
+  metadata <- parseTSSPrimaryTsMetadata(reportData)
+  methodData <- parseTSSMethods(reportData, timezone)
+  itData <- parseTSSInterpolationTypes(reportData, timezone)
+  processorData <- parseTSSProcessors(reportData, timezone)
+  changeNote <- FALSE
+  
+  if(!is.null(metadata) && !isEmptyOrBlank(metadata)){
```
lindsayplatt: `!isEmptyOrBlank(metadata)` checks with `!is.null` so you don't need both

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+    #Time Series Attributes
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Label", value=metadata[['identifier']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Parameter", value=metadata[['parameter']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Units", value=metadata[['unit']], indent=8, stringsAsFactors = FALSE))
+    if(!is.null(itData) && !isEmptyOrBlank(itData) && nrow(itData) > 0){
```
lindsayplatt: don't need `is.null` with `isEmptyOrBlank`, also you could use `isEmptyVar` for checking the rows

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+      mIt <- FALSE
+      if(nrow(itData) > 1){
+        changeNote <- TRUE
+        mIt <- TRUE
+      }
+      tsAttrs <- rbind(tsAttrs, data.frame(label="Interpolation", value=ifelse(mIt,  paste(itData[1,][['type']], "*"), itData[1,][['type']]), indent=8, stringsAsFactors = FALSE))
```
lindsayplatt: this needs a comment - what is the `ifelse` doing?

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+    }
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Sub-Location", value=metadata[['sublocation']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="UTC Offset", value=metadata[['utcOffset']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Computation", value=metadata[['computation']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Period", value=metadata[['period']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Publish", value=metadata[['publish']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Description", value=metadata[['description']], indent=8, stringsAsFactors = FALSE))
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Comments", value=metadata[['comment']], indent=8, stringsAsFactors = FALSE))
+    if(!is.null(methodData) && !isEmptyOrBlank(methodData) && nrow(methodData) > 0){
```
lindsayplatt: `isEmptyOrBlank` does `is.null` already; use `isEmptyVar` instead of `nrow() > 0`

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+      mMethod <- FALSE
+      if(nrow(methodData) > 1){
+        changeNote <- TRUE
+        mMethod <- TRUE
+      }
+      
+      tsAttrs <- rbind(tsAttrs, data.frame(label="Measurement Method", value=ifelse(mMethod,  paste(methodData[1,][['methodCode']], "*"), methodData[1,][['methodCode']]), indent=8, stringsAsFactors = FALSE))
```
lindsayplatt: might be worth pulling these `ifelse`'s out in their own line and commenting them?

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+      tsAttrs <- rbind(tsAttrs, data.frame(label="Method Start Time", value=methodData[1,][['startTime']], indent=26, stringsAsFactors = FALSE))
+    }
+    if(!is.null(processorData) && !isEmptyOrBlank(processorData) && nrow(processorData) > 0){
```
lindsayplatt: ~`is.null`~ 
~`nrow() > 0`~ --> `isEmptyVar`

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+      mProcessor <- FALSE
+      if(nrow(processorData) > 1){
+        changeNote <- TRUE
+        mProcessor <- TRUE
+      }
+      tsAttrs <- rbind(tsAttrs, data.frame(label="Processing Type", value=ifelse(mProcessor, paste(processorData[1,][['processorType']], "*"), processorData[1,][['processorType']]), indent=8, stringsAsFactors = FALSE))
```
lindsayplatt: comments on `ifelse` again

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+      tsAttrs <- rbind(tsAttrs, data.frame(label="Period Start Time", value=processorData[1,][['startTime']], indent=26, stringsAsFactors = FALSE))
+      tsAttrs <- rbind(tsAttrs, data.frame(label="Period End Time", value=processorData[1,][['endTime']], indent=26, stringsAsFactors = FALSE))
+    }
+    
+    #Time Series Extended Attributes
+    extAttrs <- metadata[['extendedAttributes']]
+    if(!is.null(extAttrs) && !isEmptyOrBlank(extAttrs)){
```
lindsayplatt: ~`is.null`~

R/timeseriessummary-data.R
```diff
@@ -71,6 +78,100 @@ formatDataTable <- function(inputData){
...
+      accessValue <- ifelse(is.null(extAttrs[['ACCESS_LEVEL']]), " ", extAttrs[['ACCESS_LEVEL']])
```
lindsayplatt: rather than `is.null` for these, consider `isEmptyOrBlank` instead. We created that because it's a more robust way to check for missing information in our data.

R/timeseriessummary-data.R
```diff
@@ -388,4 +489,89 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+#' @description TSS wrapper for the readProcessors function
+#' that handles errors thrown and returns the proper data
+#' @param reportData The full report JSON object
+#' @param timezone The timezone to parse data into
+parseTSSProcessors <- function(reportData, timezone){
+  processors <- tryCatch({
+    readProcessors(reportData, timezone)
+  }, error=function(e){
+    warning(paste("Returning NULL for processors. Error:", e))
+    return(NULL)
```
lindsayplatt, line 577: do you need to return NULL or an empty list? 

`return(list())` is used in `parseTSSPrimaryTsMetadata`

lindsayplatt (commented):
If you add new tests, I would suggest adding some that test when the `fetch`, `read`, `parse`, etc functions "fail" and return empty objects because they are missing a critical field or something along those lines.

R/timeseriessummary-data.R
zmoore-usgs: In this case Larua wanted the section to always be open-able even when there is no data

lindsayplatt: Ah gotcha!

zmoore-usgs:
Not sure why but this PR doesn't seem to be picking up my latest commit to this branch. https://github.com/USGS-R/repgen/compare/master...zmoore-usgs:feature/timeseries-details?expand=1

(14 commits listed in the comparison linked to this PR but only 13 in the PR itself)

I'm going to give it a bit longer and then I might close this PR and open a new one.

mhines-usgs:
@zmoore-usgs Ive been having issues with github today too, not sure what's up. It took a while for some commits to show up on existing PRs, as well as wanting to open a new PR but my fork didn't notice I had one there..  I had to go digging. You might try closing and re-opening this one if you haven't tried that first.

zmoore-usgs:
Looks like it finally picked it up 😕

zmoore-usgs:
@lindsaycarr I think I've gotten the test coverage to a good spot now. I think I addressed all of your other feedback as well so if you want to take one last quick look at this it should be stable now.

R/timeseriessummary-data.R
lindsayplatt: @zmoore-usgs what was the verdict on why you need `NULL` and not `list()`?

PR #710, opened 2017-06-22
AQCU-1176: UV Hydro: Estimated corrected values are not plotting

kmschoep-usgs:
called readTimeSeries for primarySeries (corrected TS) instead of
filtering by non-estimated points.


dpattermann-usgs (approved):

mhines-usgs:
Seems legit to me, if it's working as expected. Could have just been a typo/mixup?

mhines-usgs (approved):

PR #711, opened 2017-06-26
AQCU-1125: adding in blue/green basic/processor derived ts descriptio…

mhines-usgs:
…n to legend. also attempt to add in the noprocessor since seems like that got missed. will have marty check that out.


inst/templates/derivationchain/body.mustache
```diff
@@ -29,4 +31,5 @@ var statsImage = '{{reportData.derivationChainSymbols.statDerived}}';
 var fillMissingDataImage = '{{reportData.derivationChainSymbols.fillMissingData}}';
 var conditionalDataImage = '{{reportData.derivationChainSymbols.conditionalFill}}';
 var verticalDatumConversionImage = '{{reportData.derivationChainSymbols.verticalDatumConversion}}';
+var noProcessorImage = '{{reportData.derivationChainSymbols.noProcessor}}';
```
mhines-usgs, line 34: @mwernimont it looks like we missed this one? or maybe it was intentional/no examples to test with? anyway, added it while I was in here, let me know if it's unknown yet and i should remove this and other changes related if so. thanks :)

inst/templates/derivationchain/custom.js
```diff
@@ -9,7 +9,8 @@ var colorMap = {
 		"statistics" : '#FFA500',
 		"calculation" : '#000080',
 		"fillmissingdata" : '#000080',
-		"correctedpassthrough" : "#3CB371"
+		"correctedpassthrough" : "#3CB371", 
+		"noprocessor" : '#000080'
```
mhines-usgs, line 13: guessed on color here..

inst/templates/derivationchain/body.mustache
mwernimont: Well, there was a discussion on how it would be so rare for that icon to ever appear that I kinda just didn't do it, but doesn't hurt to have it I suppose. Either just that it was rare or Aquarius was doing away with it, but again doesn't hurt to have it.

mwernimont (approved):

mwernimont:
Looks like you covered everything @mhines-usgs

PR #714, opened 2017-06-27
AQCU-1159: Updates to ensure that all ts attribute rows show even when no data is present

zmoore-usgs:


R/timeseriessummary-data.R
```diff
@@ -90,11 +90,13 @@ constructTSDetails <- function(reportData, timezone){
   changeNote <- FALSE
   
   if(!isEmptyOrBlank(metadata)){
-    #Time Series Attributes
+    #Time Series Attributes - Note: The order that these are added matters in order to have proper ordering on the report
```
lindsayplatt, line 93: :+1:

R/timeseriessummary-data.R
```diff
@@ -103,8 +105,10 @@ constructTSDetails <- function(reportData, timezone){
         changeNote <- TRUE
         itValue <- paste(itValue, "*")
       }
-      tsAttrs <- rbind(tsAttrs, data.frame(label="Interpolation", value=itValue, indent=8, stringsAsFactors = FALSE))
     }
+    tsAttrs <- rbind(tsAttrs, data.frame(label="Interpolation", value=itValue, indent=8, stringsAsFactors = FALSE))
+    
+    #Continued Time Series Attributes
```
lindsayplatt: just being picky about language - this makes it sounds like it might be referencing "continuous time series" but I think you're just commenting that the following code is just continuing with time series attributes?

tests/testthat/test-timeseriessummary.R
```diff
@@ -1459,14 +1459,110 @@ test_that('constructTSDetails properly constructs the two tables for the TSS det
...
   expect_equal(tsDetails[['changeNote']], TRUE)
   expect_equal(nrow(tsDetails[['tsAttrs']]), 16)
   expect_equal(nrow(tsDetails[['tsExtAttrs']]), 10)
+  
+  expect_is(tsDetails2[['tsAttrs']], 'data.frame')
+  expect_is(tsDetails2[['tsExtAttrs']], 'data.frame')
+  expect_equal(length(tsDetails2), 3)
+  expect_equal(tsDetails2[['changeNote']], FALSE)
+  expect_equal(nrow(tsDetails2[['tsAttrs']]), 13)
+  expect_equal(nrow(tsDetails2[['tsExtAttrs']]), 10)
```
lindsayplatt, line 1564: is there a way we can add a test to look for the actual empty data? Or something along those lines? Someone could change the JSON that lives here later without realizing that is specifically for testing whether the empty fields still are added to the table.

```r
row_i <- which(tsDetails2[['tsAttrs']][['labels']] == "Measurement Method")
expect_equal(tsDetails2[['tsAttrs']][['value']], "")
```

zmoore-usgs: Seems like a good idea to me, will do!

R/timeseriessummary-data.R
zmoore-usgs: Yeah, that's what I'm referring to. I can tweak it to be a bit more clear.

lindsayplatt (approved):
changes look good! I'll merge once tests pass

PR #716, opened 2017-06-27
AQCU-1153: adding parameter details to corrections sections of report

mhines-usgs:
See https://github.com/USGS-R/repgen/pull/707 for long convo on approach and various reviews


mhines-usgs:
ok to merge @lindsaycarr ? :) oh, just realized I didn't do the documentation here. brb.

R/timeseriessummary-data.R
```diff
@@ -524,6 +525,151 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+  gapLimit <- ".dplyr"
+  deviationValue <- ".dplyr"
+  deviationType <- ".dplyr"
+  windowSizeInMinutes <- ".dplyr"
+  ungroup <- ".dplyr"
+  
+  params <- corrections$parameters
+  corrections$parameters <- NULL
+  
+  if (length(params)>0) {
```
lindsayplatt: this was where I made a comment before that might have been lost in the shuffle - could use `isEmptyVar` here. Up to you! I don't remember if we were trying to use those functions everywhere or not.

mhines-usgs: Sure - I'll try it out here locally!

mhines-usgs: Huh, I thought it worked before I pushed it up here but I guess I didn't build and reload again. I'm going to change it back, getting errors on isEmptyVar

mhines-usgs: Hah, no, I missed the !

R/timeseriessummary-data.R
```diff
@@ -551,7 +551,7 @@ unNestCorrectionParameters <- function(corrections, timezone) {
   params <- corrections$parameters
   corrections$parameters <- NULL
   
-  if (isEmptyVar(params)) {
+  if (!isEmptyVar(params)) {
```
lindsayplatt: :100:

mhines-usgs:
Based on PT's PR for AQCU-1158, going to update some names to be StartShiftPoints instead of startShiftPoints (.NET style over camelcase style we usually see)

and fix some tests that have bad json snippets
and modify the existing json examples to match the field names/formats, too.

mhines-usgs:
Making good progress here. After I got PT's changes for 1158, I'm getting more examples to improve the work already here, such as where I made assumptions on only getting x# of values.

mhines-usgs:
Alright.. I think I'm done now.. @lindsaycarr ! :grinning:

lindsayplatt:
@mhines-usgs I will review now that checks have passed!

R/timeseriessummary-data.R
```diff
@@ -524,6 +525,178 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+#' @param driftPoints the driftPoints values in a list including
+#' date/times and the difference values in feet
+#' @param timezone the timezone for the drift parameters
+#' @return formatted string of drift parameters for report display
+formatCorrectionsParamDrift <- function(driftPoints, timezone) {
+  formattedParameters <- ""
+  if (!isEmptyOrBlank(driftPoints) && !isEmptyOrBlank(timezone)) {
+    driftPoints <- as.data.frame(driftPoints)
+    if (any(names(driftPoints) %in% c("Value","Offset"))) {
+        for (i in 1:nrow(driftPoints)) {
```
lindsayplatt: to simplify the if statements, you could do `driftpoints <- as.data.frame(driftpoints)` outside of the first if, right below `formattedParameters <- ""`. Then you just have one if statement: `if (!isEmptyOrBlank(driftPoints) && !isEmptyOrBlank(timezone) && any(names(driftPoints) %in% c("Value","Offset")) ) {`

lindsayplatt: What if "value" was in the data, but "Offset" wasn't? This would fail since you are saying either to get into the if code, but only "offset" is used

R/timeseriessummary-data.R
```diff
@@ -524,6 +525,178 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+#' @param startShiftPoints startShift points values and offset as a list
+#' @param endShiftPoints endShift points values and offset as a list
+#' @param usgsType the USGSMultiPoint type
+#' @return formatted string of USGSMultiPoint parameters for report display
+formatCorrectionsParamUSGSMultiPoint <- function(startShiftPoints, endShiftPoints, usgsType) {
+  formattedParameters <- "USGSMultiPoint "
+  if (!isEmptyOrBlank(startShiftPoints) && !isEmptyOrBlank(endShiftPoints) && !isEmptyOrBlank(usgsType)) {
+    startShiftPoints <- as.data.frame(startShiftPoints)
+    endShiftPoints <- as.data.frame(endShiftPoints)
+    if (any(names(startShiftPoints) %in% c("Value","Offset"))) {
```
lindsayplatt: you use `any` here for start and end points, but do you need it to include both? Both are used in the paste functions. Maybe you need `all` instead? Then you just switch the order: `all(c("Value","Offset") %in% names(startShiftPoints))`

R/timeseriessummary-data.R
```diff
@@ -524,6 +525,178 @@ parseTSSGapTolerances <- function(reportData, timezone){
...
+#' @param upperThresholdPoints upper threshold points date/times and values as a list
+#' @param lowerThresholdPoints lower threshold points date/times and values as a list
+#' @param timezone the timezone for the drift parameters
+#' #' @return formatted string of adjustable trim parameters for report display
+formatCorrectionsParamAdjustableTrim <- function(upperThresholdPoints, lowerThresholdPoints, timezone) {
+  formattedParameters <- "Adjustable trim "
+  if (!isEmptyOrBlank(upperThresholdPoints) || !isEmptyOrBlank(lowerThresholdPoints) && !isEmptyOrBlank(timezone)) {
+    upperThresholdPoints <- as.data.frame(upperThresholdPoints)
+    lowerThresholdPoints <- as.data.frame(lowerThresholdPoints)
+    if (any(names(upperThresholdPoints) %in% c("Value","Time"))) {
```
lindsayplatt: same concern about using `any` here and for the lowerThresholdPoints as in the last function

tests/testthat/test-timeseriessummary.R
```diff
@@ -1587,5 +1581,555 @@ test_that('constructTSDetails properly constructs the two tables for the TSS det
...
+                           "url": "https://cida-eros-aqcudev.er.usgs.gov:8444/aqcu-webservice/service/reports/correctionsataglance/?endDate=2017-02-28Z&station=01069500&startDate=2017-01-01Z&primaryTimeseriesIdentifier=efb88ed6b5dc41dcaa33931cd6c144a2"
+                           }
+                           }
+}')
+  corrections <- repgen:::parseTSSProcessingCorrections(deviationJson, "normal", timezone)
+  expect_equal(corrections[['formattedParameters']], "Deviation type DeviationFromMinimum; value: 0.01, window size 15 minutes")
+  expect_equal(names(corrections),c("appliedTimeUtc","comment","startTime","endTime","type","user","processingOrder","DeviationValue","DeviationType","WindowSizeInMinutes","timezone","formattedParameters"))
+})
+
+test_that('unNestCorrectionParameters correctly formats empty parameters data ', {
```
lindsayplatt, line 2095: should we also add a test for the unknown parameter type? The switch statement for `formattedParameters` just returns whatever appears in `type`

mhines-usgs:
Thanks, reworking things now!

R/timeseriessummary-data.R
mhines-usgs: Oops, this is actually checking for the wrong data, should be Time and Offset..

mhines-usgs:
Addressed all the feedback and added a few more tests!

tests/testthat/test-timeseriessummary.R
mhines-usgs: Let me know if I added what you were thinking, if not, let me know..

mhines-usgs: https://github.com/USGS-R/repgen/pull/716/commits/dda79f3405236b2433fc985cd4c5a191ecb2b379#diff-ab088e65e082708b4dea806d74c49e80R2132

lindsayplatt:
Looks good! Merging

PR #717, opened 2017-06-28
AQCU-1150: Added a generic note explaining that gaps during the start…

zmoore-usgs:
… or end of the report period will not be listed.


zmoore-usgs:
Since Travis seems to be failing for unrelated reasons, and these changes are trivial HTML changes that wouldn't affect our tests anyway, I'm going to go ahead and merge this.

PR #718, opened 2017-06-28
Update travis.yml because of a knitr failure

mhines-usgs:
adding dist: trusty


mhines-usgs:
That didn't help. Closing.

PR #719, opened 2017-06-29
AQCU-1188: adding fixed table layout and widths for columns in the TS…

mhines-usgs:
…S table. It works for all report sections right now, if we add another section we may need to review.


zmoore-usgs:
Pulling this down to test now!

zmoore-usgs (approved):
Looks good to me!

PR #721, opened 2017-06-29
AQCU-1178: changes to how we log axes

lindsayplatt:
Plot axes were not properly set up for logging (`gsplot(ylog=TRUE)` has no effect) but data was being added as logged. Proper way to log is to use `view(log='y')` when the y axis should be logged and `view(log='')` when the axis should not be logged.

Fixed tests and removed some unused code from #631/#633.


zmoore-usgs (approved):

mhines-usgs:
Tested this locally and it looks better!

mhines-usgs (approved):

PR #722, opened 2017-06-29
AQCU-1179: handle values of zero in field visit measurements

kmschoep-usgs:
Zeros in field visit measurements would break the log axis
functionality.  Fix was two-fold: added a check for zeros in field visit
measurement data when excludeZeroNegative was false or missing. Then
stripped zeros/negatives from field visit measurement data when
excludeZeroNegative was true.

### DIFF
--- a/R/dvhydrograph-render.R
+++ b/R/dvhydrograph-render.R
@@ -74,7 +74,15 @@ createDVHydrographPlot <- function(reportObject){
   primarySeriesApprovals <- parsePrimarySeriesApprovals(reportObject, startDate, endDate)
   primarySeriesLegend <- fetchReportMetadataField(reportObject, 'primarySeriesLabel')
   approvals <- readApprovalBar(primarySeriesApprovals, timezone, legend_nm=primarySeriesLegend, snapToDayBoundaries=TRUE)
-  logAxis <- isLogged(priorityTS[['points']], priorityTS[['isVolumetricFlow']], excludeZeroNegativeFlag) && minMaxCanLog
+  primaryTSCanLog <- isLogged(priorityTS[['points']], priorityTS[['isVolumetricFlow']], excludeZeroNegativeFlag)
+  if (!isEmptyOrBlank(fieldVisitMeasurements)) {
+    fieldVisitMeasurementsCanLog <- isLogged(fieldVisitMeasurements, TRUE, excludeZeroNegativeFlag) 
+  }
+  else {
+    fieldVisitMeasurementsCanLog <- TRUE
+  }
+  
+  logAxis <- primaryTSCanLog && fieldVisitMeasurementsCanLog && minMaxCanLog
   yLabel <- paste0(priorityTS[['type']], ", ", priorityTS[['units']])
 
   #Get Estimated / Non-Estimated Edges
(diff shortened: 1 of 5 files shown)


R/dvhydrograph-render.R
```diff
@@ -89,6 +97,11 @@ createDVHydrographPlot <- function(reportObject){
   estimated3EdgesDf <- as.data.frame(estimated3Edges, stringsAsFactors = FALSE)
   estimated4EdgesDf <- as.data.frame(estimated4Edges, stringsAsFactors = FALSE)
   comparisonEdgesDf <- as.data.frame(comparisonEdges, stringsAsFactors = FALSE)
+  
+  #remove zeros from field measurements, if present:
+  if (excludeZeroNegativeFlag && !isEmptyOrBlank(fieldVisitMeasurements)) {
+    fieldVisitMeasurements <- removeZeroNegative(fieldVisitMeasurements)
+  }
```
lindsayplatt: I think you only need to check `isLogged` in the if statement since `excludeZeroNegativeFlag` and `!isEmptyOrBlank(fieldVisitMeasurements)` were checked in order to create `logAxis`.

R/dvhydrograph-render.R
```diff
@@ -74,7 +74,15 @@ createDVHydrographPlot <- function(reportObject){
...
   primarySeriesLegend <- fetchReportMetadataField(reportObject, 'primarySeriesLabel')
   approvals <- readApprovalBar(primarySeriesApprovals, timezone, legend_nm=primarySeriesLegend, snapToDayBoundaries=TRUE)
-  logAxis <- isLogged(priorityTS[['points']], priorityTS[['isVolumetricFlow']], excludeZeroNegativeFlag) && minMaxCanLog
+  primaryTSCanLog <- isLogged(priorityTS[['points']], priorityTS[['isVolumetricFlow']], excludeZeroNegativeFlag)
+  if (!isEmptyOrBlank(fieldVisitMeasurements)) {
+    fieldVisitMeasurementsCanLog <- isLogged(fieldVisitMeasurements, TRUE, excludeZeroNegativeFlag) 
+  }
+  else {
+    fieldVisitMeasurementsCanLog <- TRUE
+  }
```
lindsayplatt: I think lines 78-83 should go inside of `parseFieldVisitMeasurements()` much like they do for `parseMinMaxIVs()`.

R/dvhydrograph-render.R
kmschoep-usgs: This actually removes the zeros from the data if excludeZeroNegativeFlag = true; otherwise that doesn't happen and the report doesn't render if excludeZeroNegativeFlag = true and the zeros and negatives are not removed.

EDIT:  Ohhhh, I think I know what you mean now.  duh.  Yes.

lindsayplatt: Right, but `isLogged` will only return true if it won't cause an error, and that function checks `excludeZeroNegativeFlag` inside. So this `if` can just become 

```
if (logAxis) {
    fieldVisitMeasurements <- removeZeroNegative(fieldVisitMeasurements)
}
```

lindsayplatt: if there aren't zero's, this won't do anything.
if there are zero's, but `excludeZeroNegativeFlag = FALSE`, then `logAxis` will be FALSE
if there are zero's and `excludeZeroNegativeFlag = TRUE`, then `logAxis` will be TRUE

kmschoep-usgs: Yep, gotcha-- that works!

zmoore-usgs (approved):
Changes you made from Lindsay's feedback look good to me, but you might want to wait for @lindsaycarr to look at them as well.

R/utils-parse.r
```diff
@@ -157,6 +158,18 @@ parseFieldVisitMeasurements <- function(reportObject){
...
+  #Check if the field visit measurements (if they exist) allow for a log axis or not, and remove zeros/negative values if indicated
+  if (!isEmptyOrBlank(meas_Q)) {
+    if(!isEmptyOrBlank(meas_Q[['value']]) && meas_Q[['value']] <= 0 && isEmptyOrBlank(excludeZeroNegativeFlag)){
+      meas_Q[['canLog']] <- FALSE
+    } else if(!isEmptyOrBlank(meas_Q[['value']]) && meas_Q[['value']] <= 0 && !isEmptyOrBlank(excludeZeroNegativeFlag) && excludeZeroNegativeFlag ) {
+      meas_Q[['canLog']] <- TRUE
+    }
+    else if(!isEmptyOrBlank(meas_Q[['value']]) && meas_Q[['value']] > 0 ) {
+      meas_Q[['canLog']] <- TRUE
+    }
```
lindsayplatt, line 171: kind of wondering if `meas_Q[['canLog']] <- isLogged(meas_Q, isVolFlow=TRUE, excludeZeroNegativeFlag)` can be used in lieu of all these if-elses (same change could be used in `parseMinMaxIVs`). Probably out of scope, so approving!

lindsayplatt (approved):

PR #723, opened 2017-06-29
AQCU-1191: sort methods, interpolation and processors by end date rat…

mhines-usgs:
…her than start date; change * message

### DIFF
--- a/R/timeseriessummary-data.R
+++ b/R/timeseriessummary-data.R
@@ -119,10 +119,10 @@ constructTSDetails <- function(reportData, timezone){
     
     #Measurement Methods
     methodValue <- ""
-    methodStartTime <- NULL #Null so that it is hidden if there is no method
+    methodEndTime <- NULL #Null so that it is hidden if there is no method
     if(!isEmptyOrBlank(methodData) && !isEmptyVar(methodData)){
       methodValue <- methodData[1,][['methodCode']]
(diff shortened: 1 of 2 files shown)


R/timeseriessummary-data.R
```diff
@@ -119,10 +119,10 @@ constructTSDetails <- function(reportData, timezone){
     
     #Measurement Methods
     methodValue <- ""
-    methodStartTime <- NULL #Null so that it is hidden if there is no method
+    methodEndTime <- NULL #Null so that it is hidden if there is no method
     if(!isEmptyOrBlank(methodData) && !isEmptyVar(methodData)){
       methodValue <- methodData[1,][['methodCode']]
-      methodStartTime <- methodData[1,][['startTime']]
+      methodEndTime <- methodData[1,][['endTime']]
```
zmoore-usgs: I agree with your interpretation of the ticket that Method Start time should also be replaced on the report by end time but after thinking about it some more you might just want to confirm that with Laura (or perhaps she might even want both the start and end times listed?). Unless you already did that, in which case 👍

zmoore-usgs (approved):
Looks good. Tested and it works well! Just one clarifying comment

R/timeseriessummary-data.R
mhines-usgs: Thanks, I will follow up with her!

PR #724, opened 2017-07-05
AQCU-1143: Fixing order behavior by reformatting dates.

mhines-usgs:
Instead of using decreasing=TRUE or FALSE, seeing correct results with this sort change for two test cases I had locally. I'll test more tomorrow. 
Maybe it's because these are character date/times, or the fact that order doesn't behave how we think - it might be keying on the index of the value, rather than the actual content. In some cases, we get lucky and the data are sorted correctly, others not?


lindsayplatt:
I'm a little confused about the issue because `order` does order dates as I expect:

```r
somedates <- sample(seq(as.Date('1900/01/01'), as.Date('2000/01/01'), by="day"), 10)
somedates
 [1] "1982-02-05" "1964-09-15" "1978-04-17" "1955-04-20" "1952-12-19" "1978-12-05"
 [7] "1902-05-03" "1947-09-19" "1973-03-20" "1969-04-04"
order(somedates)
 [1]  7  8  5  4  2 10  9  3  6  1
order(somedates, decreasing = TRUE)
 [1]  1  6  3  9 10  2  4  5  8  7
```

mhines-usgs:
So [this change had fixed one case](https://github.com/USGS-R/repgen/pull/675/files) Laura described on JIRA, but not all of them for whatever reason. However these new changes do.. I don't really get why it would sometimes work and sometimes not? These aren't real dates, too, FYI, these are just characters. Not sure if that's a clue?

lindsayplatt:
ahh I vaguely remember a conversation about this when it was implemented...but we thought it was working

mhines-usgs:
I'll try to simplify when it's not working here in the morning!

mhines-usgs:
So, here is where we were, before this ticket, all the orders were like
```
dataRows <- dataRows[with(dataRows, order(dataRows$date, dataRows$time, decreasing = FALSE)),]
dataRows <- dataRows[order(dataRows$date, decreasing = FALSE),]
dataRows <- dataRows[order(dataRows$date, dataRows$time, decreasing = FALSE),]
```
I think this is a simple example with the reason for the 1143 task originally:
```
temp <- c("05-24-2017", "05-24-2017", "10-16-2016")
order(temp, decreasing=FALSE)
```
It comes through with 05-24-2017 as the first row - which gets passed to filterAndMarkDuplicates, and then that becomes the selected row since it's grabbing the first row there.

So I changed all the decreasing=FALSE to true, which is not quite right, but it fixed this example. They provided two examples that were negatively affected by that change.

So, this latest PR works for all those scenarios. I'm not sure why the original decreasing=FALSE isn't working, though.

lindsayplatt:
So the weird thing with those dates is that they are not in the expected scheme - YYYY-MM-DD, they are in MM-DD-YYYY. I know they are character at this point, but I'm wondering when they would ever be reformatted to having the years at the end.

If they were in the expected order, it would work:
```
temp <- c("2017-05-24", "2017-05-24", "2016-10-16")
order(temp, decreasing=FALSE)
[1] 3 1 2
order(temp, decreasing=TRUE)
[1] 1 2 3
```

mhines-usgs:
Another option (per searching stackoverflow) seems to be
```
temp <- c("05-24-2017", "05-24-2017", "10-16-2016")
order(-rank(temp), decreasing=FALSE)
```

when the temp values are character, this makes it behave

mhines-usgs:
Interesting, I can track that back and see why we're formatting them this way. Guessing it would be OK to format them the YYYY-MM-DD way

mhines-usgs:
Updated the title of the task to reflect what was eventually merged. Cleared it with Laura.

lindsayplatt:
wunderbar!  :de:

PR #726, opened 2017-07-06
AQCU-1176: UV Hydro: Estimated corrected values are not plotting

kmschoep-usgs:


R/uvhydrograph-data.R
```diff
@@ -111,15 +111,17 @@ parseUvEstimatedSeries <- function(reportObject, seriesName, month, timezone) {
...
 parsePrimarySeriesList <- function(reportObject, month, timezone) {
-  correctedSeries <- readTimeSeries(reportObject, "primarySeries", timezone, onlyMonth=month)
+  correctedSeries <- readNonEstimatedTimeSeries(reportObject, "primarySeries", timezone, onlyMonth=month)
   estimatedSeries <- readEstimatedTimeSeries(reportObject, "primarySeries", timezone, onlyMonth=month)
   uncorrectedSeries <- readTimeSeries(reportObject, "primarySeriesRaw", timezone, onlyMonth=month)
   
   inverted <- isTimeSeriesInverted(correctedSeries)
   excludeZeroNegatives <- fetchReportMetadataField(reportObject, 'excludeZeroNegative')
   
+  useEstimated <- (isEmptyOrBlank(correctedSeries) || isEmptyVar(correctedSeries[['points']])) && (!isEmptyOrBlank(estimatedSeries) && !isEmptyVar(estimatedSeries[['points']]))
```
lindsayplatt, line 123: :+1:

R/uvhydrograph-data.R
```diff
@@ -409,8 +414,10 @@ hasUpchainSeries <- function(reportObject) {
...
 #' @return the lims object that to be applied to the primary plot
 calculatePrimaryLims <- function(primarySeriesList, isDischarge) {
-  if(!is.null(primarySeriesList[['corrected']])){
+  if(!isEmptyOrBlank(primarySeriesList[['corrected']][['points']][['value']])){
     lims <- calculateLims(primarySeriesList[['corrected']][['points']])
+  } else if(primarySeriesList[['useEstimated']]){
+    lims <- calculateLims(primarySeriesList[['estimated']][['points']])
   } else {
     lims <- calculateLims(primarySeriesList[['uncorrected']][['points']])
   }
```
lindsayplatt, line 423: what is the scenario where uncorrected points are used? Just wondering if it's even relevant now.

R/uvhydrograph-render.R
```diff
@@ -327,7 +335,13 @@ createSecondaryPlot <- function(uvInfo, secondarySeriesList,
...
   invertPlot <- secondarySeriesList[['inverted']]
   
-  lims <- calculateLims(secondarySeriesList[['corrected']][['points']])
+  if(!isEmptyOrBlank(secondarySeriesList[['corrected']][['points']][['value']])){
+    lims <- calculateLims(secondarySeriesList[['corrected']][['points']])
+    corrEstPoints <- secondarySeriesList[['corrected']][['points']][['value']]
+  } else if(secondarySeriesList[['useEstimated']]){
+    lims <- calculateLims(secondarySeriesList[['estimated']][['points']])
+    corrEstPoints <- secondarySeriesList[['estimated']][['points']][['value']]
+  }
```
lindsayplatt: the other one has a final `else` for all the leftover cases. Still not sure if this is necessary or not.

R/uvhydrograph-render.R
```diff
@@ -327,7 +335,13 @@ createSecondaryPlot <- function(uvInfo, secondarySeriesList,
   
   invertPlot <- secondarySeriesList[['inverted']]
   
-  lims <- calculateLims(secondarySeriesList[['corrected']][['points']])
+  if(!isEmptyOrBlank(secondarySeriesList[['corrected']][['points']][['value']])){
+    lims <- calculateLims(secondarySeriesList[['corrected']][['points']])
+    corrEstPoints <- secondarySeriesList[['corrected']][['points']][['value']]
```
lindsayplatt: probably should rename this variable since it doesn't always refer to estimated - maybe `corrLimPoints`?

R/uvhydrograph-render.R
```diff
@@ -353,18 +367,25 @@ createSecondaryPlot <- function(uvInfo, secondarySeriesList,
...
-  doLog <- isLogged(secondarySeriesList[['corrected']][['points']], secondarySeriesList[['corrected']][['isVolumetricFlow']], excludeZeroNegativeFlag)
-  plot_object <- plotTimeSeries(plot_object, secondarySeriesList[['corrected']], "corrected", 
+  if (secondarySeriesList[['useEstimated']]) {
+  # Plot estimated corrected data, in place of non-estimated corrected data
+    doLog <- isLogged(secondarySeriesList[['estimated']][['points']], secondarySeriesList[['estimated']][['isVolumetricFlow']], excludeZeroNegativeFlag)
+    plot_object <- plotTimeSeries(plot_object, secondarySeriesList[['estimated']], "estimated", 
       timezone, getSecondaryPlotConfig, list(uvInfo[['label']], lims$ylim), excludeZeroNegativeFlag)
+  } else {
+  	#Plot estimated data if non-estimated corrected data exists
+  	if(!isEmptyOrBlank(secondarySeriesList[['estimated']]) && !isEmptyVar(secondarySeriesList[['estimated']][['points']])) {
```
lindsayplatt: so this `plotTimeSeries` call is happening for `secondarySeriesList[['estimated']]` no matter what. Seems like it can remain outside of the if-statement like it was before?

R/uvhydrograph-render.R
```diff
@@ -576,6 +602,10 @@ getPrimaryPlotConfig <- function(timeseries, name, label,
           lines = append(list(x=x, y=y, side=2, ylim=ylim, ylab=label, legend.name=paste(styles[['corr_UV_lbl']], label)), styles[['corr_UV_lines']]),
           view = list(side=2, log=doLog)
           ),
+      estimatedPrimary = list(
+          lines = append(list(x=x, y=y, side=2, ylim=ylim, ylab=label, legend.name=paste(styles[['est_UV_lbl']], label)), styles[['est_UV_lines']]),
+          view = list(side=2, log=doLog)
+          ),
```
lindsayplatt, line 608: why does this need to be added?

tests/testthat/test-uvhydrograph-render.R
```diff
@@ -29,6 +29,17 @@ test_that("uvhydrographPlot correctly skips rendering all if no primary series e
   expect_equal(length(renderList), 0)
 })
 
+test_that("uvhydrographPlot correctly renders if corrected data exists only as estimated data and not estimated data",{
+  library('jsonlite')
+  reportObject <- fromJSON(system.file('extdata','testsnippets','test-corrected-estimated-periods-uvhydro.json', package = 'repgen'))
+
+  renderList <- repgen:::uvhydrographPlot(reportObject)
+  browser()
```
lindsayplatt: probably don't need browser() here anymore :)

tests/testthat/test-uvhydrograph.R
```diff
@@ -24,7 +24,7 @@ test_that("uvhydrograph-data functions",{
   
   ###Secondary series list
   expect_is(secondarySeriesList,"list")
-  expect_equal(length(secondarySeriesList),4)
+  expect_equal(length(secondarySeriesList),5)
   expect_equal(length(secondarySeriesList$corrected),20)
   expect_equal(length(secondarySeriesList$uncorrected),20)
   expect_equal(length(secondarySeriesList$estimated),20)
```
lindsayplatt, line 30: Would be nice to have tests for `parsePimarySeriesList` and `parseSecondarySeriesList` directly

tests/testthat/test-uvhydrograph-render.R
kmschoep-usgs: whoops!

R/uvhydrograph-data.R
kmschoep-usgs: I'm not sure.  I left it in because it was there already. However, based on other observations that Laura and Chuck made when testing my first attempt, it appears that the y-axis limit calculation will need to be refactored to account for estimated points anyway.

R/uvhydrograph-render.R
kmschoep-usgs: You're right, I keep conflating "estimated" and "corrected", which is incorrect.

R/uvhydrograph-render.R
kmschoep-usgs: If I have it outside, then estimated would plot twice in the case that there's no non-estimated  corrected data:  once when called outside the if-statement, then again when/if it hits `useEstimated`.

R/uvhydrograph-render.R
kmschoep-usgs: This is so `ylab=label` can be added for estimated data.  One of the side effects of not having non-estimated data to plot is that there is no y-axis label on the plot.  More elegant solutions are welcome!

tests/testthat/test-uvhydrograph.R
kmschoep-usgs: Yes, the coveralls coverage report seems to indicate that as well ;).

R/uvhydrograph-data.R
lindsayplatt: I guess it should just be noted in that new issue that this may or may not be necessary to keep.

R/uvhydrograph-render.R
lindsayplatt: make a note of this in the new lims ticket. Same reason as https://github.com/USGS-R/repgen/pull/726/files#r126185926

R/uvhydrograph-render.R
lindsayplatt: I was thinking that L376-377 & L381-382 would be plucked out so just one call to plotTimeSeries was happening after these if statements.

R/uvhydrograph-render.R
lindsayplatt: Brainstormed for a bit, but can't think of a better way to do this

lindsayplatt (approved):

PR #730, opened 2017-07-11
AQCU-1200: change displayed qualifier value to 'code - display name' …

kmschoep-usgs:
…for TSS.

### DIFF
--- a/tests/testthat/test-timeseriessummary.R
+++ b/tests/testthat/test-timeseriessummary.R
@@ -92,6 +92,7 @@ test_that('parseTSSRealtedSeries properly retrieves the related upchain series',
 
 test_that('parseTSSQualifiers properly retrieves the qualifiers', {
   timezone <- "Etc/GMT+5"
+  browser()
   qualsJson <- fromJSON('{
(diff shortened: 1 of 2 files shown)


tests/testthat/test-timeseriessummary.R
```diff
@@ -92,6 +92,7 @@ test_that('parseTSSRealtedSeries properly retrieves the related upchain series',
 
 test_that('parseTSSQualifiers properly retrieves the qualifiers', {
   timezone <- "Etc/GMT+5"
+  browser()
```
zmoore-usgs: Should probably remove this guy

zmoore-usgs (approved):
Changes look good to me. Pulled and tested locally and it works as expected!

zmoore-usgs:
Guessing that drop might be caused by that browser call messing up tests following it?

tests/testthat/test-timeseriessummary.R
kmschoep-usgs: whoops, yes.

PR #731, opened 2017-07-12
AQCU-1196: TSS: Make Column headings sortable

zmoore-usgs:


mhines-usgs:
This works really well right after I render it in the browser, but after the files are saved locally, it doesn't seem to work anymore. It does retain the sort that you set before saving, though. That could be an improvement task since it's not explicitly called out in acceptance criteria. Not sure how hard that would be to handle?

mhines-usgs (approved):

zmoore-usgs:
Hmm I'll look a little bit into having it work after saving it

zmoore-usgs:
@mhines-usgs ^ That should hopefully fix the issue

mhines-usgs:
checking it out now!

mhines-usgs:
Excellent! :shipit:

mhines-usgs:
One other observation related to usability is that it's not immediately obvious that the columns are sortable until you click on them. Not sure that's changeable with the theme, though.

zmoore-usgs:
Yeah I've been messing around with that a bit to see if I can come up with something that implies they are sortable without implying any ordering since initially everything is sorted by start date and I don't want to imply that the other columns are ordered somehow. 

I think if I can set start date as the default selected/sorted column that might help, though it wouldn't necessarily imply that any other column but start date is sortable.

zmoore-usgs:
If I could get something similar to this icon that might work: http://fontawesome.io/icon/sort/

mhines-usgs:
Yeah, that would be perfect, if you can figure out how to get it to display. :/

mhines-usgs:
Looks great @zmoore-usgs !

PR #740, opened 2017-07-21
AQCU-1201: WIP consider the estimated points for secondary plot lims

mhines-usgs:


R/uvhydrograph-render.R
```diff
@@ -344,10 +344,18 @@ createSecondaryPlot <- function(uvInfo, secondarySeriesList,
   
   if(!isEmptyOrBlank(secondarySeriesList[['corrected']][['points']][['value']])){
     lims <- calculateLims(secondarySeriesList[['corrected']][['points']])
-    corrLimPoints <- secondarySeriesList[['corrected']][['points']][['value']]
```
mhines-usgs: These corrLimPoints weren't used, maybe that's part of the problem...  ? I removed them but perhaps adding them into logic below is the correct solution rather than what I've worked up here.

lindsayplatt: Hmmm I think all you need is to have estimated and corrected grouped together before the limits are calculated. Then you don't need to add anything.

```
allCorrectedData <- rbind(secondarySeriesList[['corrected']][['points']], secondarySeriesList[['estimated']][['points']])
if(!isEmptyOrBlank(allCorrectedData )){
    lims <- calculateLims(allCorrectedData )
    lims[['ylim']] <- bufferLims(primaryLims[['ylim']], secondarySeriesList[['uncorrected']][['points']][['value']])
  } else {
    lims <- calculateLims(secondarySeriesList[['uncorrected']][['points']])
  }
```

^ this copies what primary plot does. Since estimated would be used no matter what, the additional scenario for "useEstimated" would not be necessary.

lindsayplatt:
:crossed_fingers: c'mon checks!

mhines-usgs:
lol, it passed locally ;D i think we're good. thank you @lindsaycarr !

PR #741, opened 2017-07-24
AQCU-1199: modifying the format of grades AND NOTES before merging with qualifi…

mhines-usgs:
…ers because there was a column mismatch

### DIFF
--- a/R/timeseriessummary-data.R
+++ b/R/timeseriessummary-data.R
@@ -410,11 +410,16 @@ parseTSSGrades <- function(reportData, timezone){
   
   if(!isEmptyOrBlank(grades)){
     grades <- as.data.frame(grades, stringsAsFactors=FALSE)
-    colnames(grades)[which(colnames(grades) == 'code')] <- "value"
-    grades[['metaType']] <- paste0('Grade ', grades[['value']])
     grades <- grades[order(grades[['startDate']]),]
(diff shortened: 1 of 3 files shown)


zmoore-usgs:
Pulling this down to test now

mhines-usgs:
Thanks @zmoore-usgs I just noticed the test failure, and will check that out in a minute.

zmoore-usgs (approved):
Tested locally and it works well for me! Guessing the test failure is just needing to replace "code" with "value" in the test cases. Once that is passing I think this is good to go!

mhines-usgs:
@zmoore-usgs yep, just pushed that up! :D thanks for reviewing.

zmoore-usgs:
Just found a small issue, not sure if it's related to this ticket or not. Testing with timeseriessummary-example2.json it seems the note row in the metadata section is a bit malformed? Could just be bad test data though.

image

mhines-usgs:
Thanks, I am grabbing the JSON she sent in with her problem too to verify, and will check that out too.

mhines-usgs:
Is that a qualifier? (Note) or something different? Might also have to adjust notes...
Yeah I think it's probably the same issue, I can fix it while I'm in here..

zmoore-usgs:
Notes are their own thing separate from qualifiers and grades I believe.

PR #747, opened 2017-07-27
AQCU-1228/1223: changes to explanation on DC

mhines-usgs:
AQCU-1228:  Rename:
"Processor Types" --> "Dataset Processor Types"
"Primary Timeseries" --> "Selected Time Series"
"Upchain/downchain Timeseries" --> "Upchain/Downchain Time Series"
Add under the "Explanation" title add a note that "TS = Time Series" since "TS" is used in the explanation.

AQCU-1223: Dynamic dashed line

### DIFF
--- a/inst/templates/derivationchain/custom.js
+++ b/inst/templates/derivationchain/custom.js
@@ -373,7 +373,7 @@ function generateLegend(nodes) {
   $(legendContainer).html('<strong>Explanation</strong><br/><br/><span style="text-decoration:underline">Dataset Processor Types:</span> <br/>');
   
   for(var i = 0; i < nodes.length; i++){
-    if( flags[nodes[i].data["location"]]) continue;
+    if(flags[nodes[i].data["location"]]) continue;
       flags[nodes[i].data["location"]] = true;
       output.push(nodes[i].data["location"]);
       
--- a/inst/templates/derivationchain/custom.js
+++ b/inst/templates/derivationchain/custom.js
@@ -368,16 +368,15 @@ function generateLegend(nodes) {
   var visibleProcessors = [];
   var addExternalLine = false;
   var legendContainer = $("#legendContainer");
+  var flags = [];
+  var output = [];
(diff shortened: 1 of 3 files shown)


inst/templates/derivationchain/custom.js
```diff
@@ -368,8 +368,15 @@ function generateLegend(nodes) {
   var visibleProcessors = [];
   var addExternalLine = false;
   var legendContainer = $("#legendContainer");
-  $(legendContainer).html('<strong>Explanation</strong><br/><br/><span style="text-decoration:underline">Processor Types:</span> <br/>');
-
+  $(legendContainer).html('<strong>Explanation</strong><br/><br/><span style="text-decoration:underline">Dataset Processor Types:</span> <br/>');
+  
+  var flags = [], output = [], l = nodes.length, i;
```
mhines-usgs: this loops through the node locations and if there are different locations in the list, add to another array which we'll check later before adding that dashed line to the legend

inst/templates/derivationchain/custom.js
```diff
@@ -368,8 +368,15 @@ function generateLegend(nodes) {
...
+    for( i=0; i<l; i++) {
+      if( flags[nodes[i].data["location"]]) continue;
+      flags[nodes[i].data["location"]] = true;
+      output.push(nodes[i].data["location"]);
+    }
```
zmoore-usgs: I might be incorrectly understanding what this is doing but I think this process could mostly happen within the loop below this one?

If you create an array of added locations before the loop and add each node's location to that array if it's not already in the array and then after the loop you can have if(locations.length>1){addExternalLine = true;}

That way we won't be running through a loop of all of the nodes twice. The overall approach of your ticket seems good though!

mhines-usgs: Ok, I'll try that, sorry to flood you with PRs!

mhines-usgs:
Seems to work fine, thanks for the suggestion @zmoore-usgs !

PR #749, opened 2017-08-02
AQCU-1212 don't force gaps to noon + print zero times + tests

lindsayplatt:
DV Gaps in TSS were printing as "2010-10-01 12:00:00", and should now be printing as "2010-10-01 00:00:00".


mhines-usgs:
Pulling this to check it out now!

mhines-usgs:
Works great with a bunch of local test files I have, :+1:

PR #750, opened 2017-08-07
AQCU-1210: add advanced options configurations on report

mhines-usgs:
So that when users save them locally, they know what options were set in the advanced options to generate the reports


zmoore-usgs (approved):
Looks good to me!

PR #758, opened 2017-08-28
AQCU-1248: parameters passed into gsplot v 8 changed for error_bar

mhines-usgs:
And we didn't update uvhydro to reflect that. Changed now, and now error bars appear again.


mhines-usgs:
Need some tests updated too, so this may error. Working on those. FYI @dpattermann-usgs

dpattermann-usgs:
When the tests pass, it looks good to me. 👍

PR #760, opened 2017-09-21
AQCU-1286: TSS reworking parameter data output

mhines-usgs:
- relax requirement for USGSMultiPoint shift so it doesn't require endShiftPoints, only startShiftPoints, -
- remove type from correction amount(s) output when no parameters exist, just leave column empty


mhines-usgs:
Still waiting for some feedback on JIRA from Laura/Chuck about "missing data" -- but this stands alone from that.

R/timeseriessummary-data.R
```diff
@@ -583,7 +583,7 @@ unNestCorrectionParameters <- function(corrections, timezone) {
                                    "AdjustableTrim" = formatCorrectionsParamAdjustableTrim(UpperThresholdPoints, LowerThresholdPoints, timezone),
                                    "FillGaps" = formatCorrectionsParamFillGaps(ResamplePeriod, GapLimit),
                                    "Deviation" = formatCorrectionsParamDeviation(DeviationValue, DeviationType, WindowSizeInMinutes),
-                                   type)) %>% 
+                                   " ")) %>%
```
dpattermann-usgs: Does this never print the type then?

mhines-usgs: This ticket requests that we don't print anything in that column if it doesn't have any parameters associated. 
It was defaulting to print the type when there were no parameters for some of the correction types (DeleteRegion, CopyPaste, Freehand and Revert To Raw) but Laura/et al don't want anything there when there are no parameters. The type still prints in the CorrectionType column, previously it was duplicating like this screenshot
image

dpattermann-usgs: 👍

mhines-usgs:
Now with new correction type persistence gap fill

mhines-usgs:
Not sure what's going on with coveralls but I'm going to merge this...

PR #762, opened 2017-10-11
Aqcu 1226: Map API output to preferred nomenclature for Corrections section of report

mhines-usgs:


dpattermann-usgs (approved):
👍 This looks more like what Laura was expecting!

PR #764, opened 2017-10-19
AQCU-1278: the case of the missing TS

mhines-usgs:
allow dates that are greater than AND equal to the nodeStartDate


dpattermann-usgs (approved):
What a weird bug. Nice find! 👍

mhines-usgs:
I'm not sure this was the only problem after Laura has elaborated more. I may revert this.

PR #766, opened 2017-10-23
AQCU 1238 Change listing and calc of gaps for stat derived TS

dpattermann-usgs:
Added function that checks if the TS is stat derived and if it is, it makes the gap date ranges inclusive.


mhines-usgs:
seems to work as desired based on the json I pulled Laura described in the ticket. One thing I noticed is that the message "Gap dates are exclusive, therefore the gap period is inside the listed dates." should still be removed from the body.mustache file in /inst/templates/timeseriessummary.

you'll have to fiddle with the test there too, to make it work. it's missing some of the metadata needed now in that function

R/timeseriessummary-data.R
```diff
@@ -895,9 +895,11 @@ parseTSSProcessors <- function(reportData, timezone){
 #' @param gaps the gaps data to look at
 #'
 makeGapsInclusiveIfStatistic <- function(reportData, timezone, gaps){
+  tsMetadata <- parseTSSPrimaryTsMetadata(reportData)
+  processors <- parseTSSProcessors(reportData, timezone)
   
-  if(parseTSSPrimaryTsMetadata(reportData)[['timeSeriesType']]=='ProcessorDerived' 
-     && parseTSSProcessors(reportData,timezone)[['processorType']]=="statistics"){
+  if((!isEmptyOrBlank(tsMetadata) && tsMetadata[['timeSeriesType']]=='ProcessorDerived')
```
mhines-usgs: :+1: and gets past the test issue.

mhines-usgs (approved):

mhines-usgs:
just yank that line in the mustache template and you should be good!

dpattermann-usgs:
@mhines-usgs This changes the message based on whether it's stat derived or not. You should probably take a look to make sure the message makes sense and to make sure I didn't miss anything.

mhines-usgs:
OK, pulling to review!

mhines-usgs:
The message switch works well with a few different files I've got locally. I'm wondering if we need a message at all, though, for the stat-derived ts. Seems like Laura suggests this would be expected and maybe not need a note? May want to ask

Otherwise, I'd make sure to use the lingo/formatting from Laura's description, e.g. rather than "Gap dates are inclusive (because the series is Stat Derived), therefore the gap period is the listed dates."  >> "Gap dates are inclusive (because the series is stat-derived), therefore the gap period is the duration between the dates."

just my $0.02

mhines-usgs:
I see [one small thing](https://travis-ci.org/USGS-R/repgen/builds/292214079#L3713) that is probably the reason for the travis failure, but am not sure how I could push to this PR or if I can even..

1. add _isStatDerived=FALSE_ into the arguments for line 1151 in test-timeseriessummary.R 
2. minor addition to documentation for lines 506 and 898 for the new param in timeseriessummary-data.R

`#' @param isStatDerived boolean if the ts is stat-derived`

PR #768, opened 2017-11-20
Aqcu 1332: tweak advanced options logic

mhines-usgs:
to show whether or not discrete data are excluded


dpattermann-usgs (approved):
This also looks fine, although the Travis failure is odd.

mhines-usgs:
Yeah, I've been comparing the last log versus this one and it seems like it could be an issue with mran. If it doesn't self-resolve, I'll open an issue with them :/

mhines-usgs:
:clap: so it ended up being that there are a few places we reference mran, and they all need to be https now, their redirect from http must not work anymore.

PR #770, opened 2017-11-22
AQCU-1305: the legend wasn't always working correctly, fixed

mhines-usgs:
this resolves the issue where the legend wouldn't update when changing time periods. we had tried to simplify the looping to do both loops together but it didn't work out.

### DIFF
--- a/inst/templates/derivationchain/custom.js
+++ b/inst/templates/derivationchain/custom.js
@@ -368,15 +368,18 @@ function generateLegend(nodes) {
   var visibleProcessors = [];
   var addExternalLine = false;
   var legendContainer = $("#legendContainer");
-  var flags = [];
-  var output = [];
   $(legendContainer).html('<strong>Explanation</strong><br/><br/><span style="text-decoration:underline">Dataset Processor Types:</span> <br/>');
   
+  var flags = []; 
+  output = [];
+  l = nodes.length;
+  for(var i=0; i<l; i++) {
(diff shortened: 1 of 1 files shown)


inst/templates/derivationchain/custom.js
```diff
@@ -368,15 +368,18 @@ function generateLegend(nodes) {
   var visibleProcessors = [];
   var addExternalLine = false;
   var legendContainer = $("#legendContainer");
-  var flags = [];
-  var output = [];
   $(legendContainer).html('<strong>Explanation</strong><br/><br/><span style="text-decoration:underline">Dataset Processor Types:</span> <br/>');
   
+  var flags = []; 
+  output = [];
+  l = nodes.length;
```
dpattermann-usgs: Where is this variable used?

inst/templates/derivationchain/custom.js
```diff
@@ -368,15 +368,18 @@ function generateLegend(nodes) {
...
+  for(var i=0; i<l; i++) {
```
mhines-usgs: here, in the middle i<l -- i could just use nodes.length

dpattermann-usgs: Ugh, thought that was a 1. Could you change it to nodesLength maybe? Otherwise, this seems to make sense.

mhines-usgs: yup yup!

dpattermann-usgs (approved):

PR #772, opened 2017-12-07
removing the hack from script to remove "" from the list of imports, …

mhines-usgs:
…seems to be causing issues when run at CHS, where it removes dplyr from that list and that never gets installed

### DIFF
--- a/inst/extdata/installPackages.R
+++ b/inst/extdata/installPackages.R
@@ -34,7 +34,8 @@ repgenImports <- function (lib.loc = NULL, encoding = "") {
   }
 
   imports <- strsplit(unlist(desc$Imports), "[\n,]+")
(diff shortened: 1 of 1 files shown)


mhines-usgs:
@lindsaycarr and/or @dpattermann-usgs -- this seems to be causing an issue at CHS where they are using another version of R. I updated it so that the removing of the empty item in the list was cleaner so it wouldn't remove dplyr from the list, which is what was happening there :/

lindsayplatt:
Sorry I missed this - had a couple of things with due dates on my plate today and was heads down on that. 

Only thing I would say is that `x != ""` is more common than `!x == ""` but that's very much in the weeds. That old way is way to reliant on the order of things. Nice catch! And you removed one occurrence of the `### THIS IS A HACK` code :)

mhines-usgs:
Yes, those darn hacks. :)

PR #778, opened 2017-12-27
Aqcu 1319: 5YR GW Use multiple axes for different parameters

mhines-usgs:


mhines-usgs:
I'm working out an issue with labels on uvhydro, but I think a review would be helpful at any time @dpattermann-usgs & @lindsaycarr ! :thanks:

mhines-usgs:
@dpattermann-usgs things seem to work ok here, but i think my side logic for if sides should share sides versus get their own could be done better. if you have time to look at it with me, i'd love input. i'll think about a better solution too.

mhines-usgs:
@dpattermann-usgs if you or anyone has ideas, I'm stuck on how to get the calculated limits for each side that come back from the getSides function (in sides[["sideLims"]]) into the plot functions for each timeseries.

We used to plot them by timeseries, not by side. 

e.g. here:

https://github.com/USGS-R/repgen/pull/778/files#diff-4a05d16a572ce7bfddf1506c5b5469c1R110
https://github.com/USGS-R/repgen/pull/778/files#diff-4a05d16a572ce7bfddf1506c5b5469c1R115
https://github.com/USGS-R/repgen/pull/778/files#diff-4a05d16a572ce7bfddf1506c5b5469c1R120
https://github.com/USGS-R/repgen/pull/778/files#diff-4a05d16a572ce7bfddf1506c5b5469c1R125

into the ylim parameter

dpattermann-usgs (approved):
I've been trying to find ways to simplify the getSides function, but haven't been able to figure out how. I'm not an R master, but I think this is fine.

mhines-usgs:
Thanks.. it's not that much different than how UVHydro handles it but a little different because that doesn't have the same logic for the number of timeseries that can be involved. Just the primary, reference, and comparison.. But I agree, it's complicated to sort out in R

PR #784, opened 2018-03-15
Aqcu 1541 DV Hydro Report Transition

mhines-usgs:
yet to do:

- [x] fix broken tests
- [ ] test with more json examples


R/utils-fetch.R
```diff
@@ -182,6 +182,15 @@ fetchPrimarySeriesQualifiers <- function(reportObject){
   return(val)
 }
 
+#'Fetch Qualifier Metadata
```
mhines-usgs, line 196: When we merge TSS transition work in, this will be redundant/conflict.

zmoore-usgs (approved):
Looks good so far! Once you've been able to test with more finished JSON examples if this all still works I think it's probably good to go

PR #785, opened 2018-03-21
AQCU-1599: estimated data display issues for UV Hydro

mhines-usgs:
sort the estimatedPeriods by startDate before creating gaps for estimated periods, otherwise we lose data.


mhines-usgs:
There is still a little work on the JIRA to eliminate an unwanted gap, but merging this would help Chuck with some testing now. @zmoore-usgs mind taking a quick look so I can merge?

zmoore-usgs (approved):

PR #790, opened 2018-04-10
AQCU-1580: Add "period" to derivation chain popup info per node.

kmschoep-usgs:


man/makeGapsInclusiveIfStatistic.Rd
```diff
@@ -0,0 +1,20 @@
+% Generated by roxygen2: do not edit by hand
```
mhines-usgs, line 1: lol, i just keep ignoring this file in my Rstudio.. good for you.

kmschoep-usgs: I never know what to include in my commits for R, so went for it...

mhines-usgs (approved):
:+1: looks good to me

mhines-usgs:
This will actually be a good test to see if this makes its way out to reporting-test!

PR #792, opened 2018-04-17
AQCU-639: handle null labels coming in for callouts.

kmschoep-usgs:


mhines-usgs:
pulling this locally to check out

R/utils-limits.R
```diff
@@ -82,14 +82,14 @@ testCalloutsByView <- function(plot_obj, view_num, xlimits_real, width_char, xra
...
     
     if (!isEmptyOrBlank(callout_args$x)) {  
-      text_len <- nchar(callout_args$labels)
+      text_len <- ifelse(is.na(callout_args$labels),0,nchar(callout_args$labels))
       
       len <- ifelse(is.null(callout_args$length), 0.1, callout_args$length)
       
       xend <- len * xrange * cos(2*pi*(30/360))
       xnew <- callout_args$x + xend + (width_char * text_len) 
+      thing <- xlimits_real[2]
```
mhines-usgs: for debugging?

kmschoep-usgs: oops, yep.

R/utils-limits.R
```diff
@@ -82,8 +82,7 @@ testCalloutsByView <- function(plot_obj, view_num, xlimits_real, width_char, xra
     callout_args <- plot_obj[[view_num]][[i]]
     
     if (!isEmptyOrBlank(callout_args$x)) {  
-      text_len <- nchar(callout_args$labels)
-      
+      text_len <- ifelse(is.na(callout_args$labels),0,nchar(callout_args$labels))
```
lindsayplatt, line 85: don't need this to check `isEmptyOrBlank(callout_args$labels)` here?

lindsayplatt: I guess depends on if the value you are trying to catch is `NA` or `NULL`

kmschoep-usgs: We're trying to catch NA, but what I found is that isEmptyOrBlank only returns true if the entire vector is empty or blank.

lindsayplatt: Ahhhh I wonder if we should create another method of isEmptyOrBlank for vectors...what else is it not catching that it should?

mhines-usgs: Yeah I'm wondering if that's exactly the issue I'm having with another bug right now.. :grimacing:

PR #796, opened 2018-06-07
Aqcu 1351 extremes

mhines-usgs:
- apply qualifiers appropriately to points


mhines-usgs:
Will try having Travis use an older version of pandoc...

zmoore-usgs (approved):
:+1: Looks good to me!

PR #797, opened 2018-06-12
AQCU-1639: modify uvhydro, dvhydro & vdiagram to communicate whether publish is true or f…

mhines-usgs:
…alse to plots


R/utils-fetch.R
```diff
@@ -305,15 +305,6 @@ fetchQualifiers <- function(reportObject){
   return(val)
 }
 
-#'Fetch Qualifier Metadata (TSS)
```
mhines-usgs, line 308: This should have been removed when we merged the TSS work in. It became redundant. It's removed now :D

zmoore-usgs (approved):

PR #810, opened 2018-08-02
AQCU-1706: legacy branch update qualifier table creation logic/update tests to match

mhines-usgs:


R/extremes-data.R
```diff
@@ -105,9 +105,18 @@ extremesTable <- function(reportObject) {
 #' @return A vector of qualifiers.
 #' @importFrom dplyr mutate
 extremesQualifiersTable <- function(reportObject, table, primaryHeaderTerm, upchainHeaderTerm) {
-  #Construct List of all qualifiers
+  #get all unique qualifier bits
   qualifiersList <- list(data.frame(reportObject$dv$qualifiers), data.frame(reportObject$upchain$qualifiers), data.frame(reportObject$primary$qualifiers))
-  qualifiersList <- Reduce(function(...) merge(..., all=T), qualifiersList)
```
mhines-usgs, line 110: according to searches I did, reduce/merge only works with 2 objects, and we can/often have 3 here. so i just rewrote it to work with as many as needed.

mhines-usgs: https://stackoverflow.com/questions/48799959/r-reduce-with-merge-and-more-than-2-suffixes-or-how-to-merge-multiple-datafr

zmoore-usgs: Ah, good to know!

zmoore-usgs (approved):

PR #811, opened 2018-08-10
AQCU-1644: update mran snapshot date and travis R version

mhines-usgs:


.travis.yml
```diff
@@ -12,10 +12,10 @@ install:
...
+  - Rscript -e 'install.packages(c("bitops","Rcpp","digest","magrittr","stringi","stringr","yaml","evaluate","formatR","highr","R6","assertthat","lazyeval","DBI","jsonlite","brew","mime","curl","caTools","openssl","memoise","whisker","rstudioapi","git2r","withr","markdown","knitr","htmltools", "rmarkdown","RCurl","htmlTable","roxygen2","dplyr","lubridate", "readr", "base64enc", "whisker"), repos="https://mran.microsoft.com/snapshot/2018-08-09", type="source")'
   
   #Install a different version of HTTR
-  - Rscript -e 'install.packages("httr", repos = "https://mran.microsoft.com/snapshot/2016-01-27",type="source")'
+  - Rscript -e 'install.packages("httr", repos = "https://mran.microsoft.com/snapshot/2018-08-09",type="source")'
```
zmoore-usgs, line 18: I feel like I remember this being intentional because the newer version changed the case of some method which broke something else somewhere... But maybe that's all fixed now?

mhines-usgs: I agree, but I'm not sure we need to be tied down any more, but I guess we'll see. Unless you can run this locally. I still don't have all the auth stuff figured out so can't truly test this outside of RStudio

zmoore-usgs: Yeah I'm fine with going for the merge and seeing how this behaves on Test. Once I get MLR-288 done hopefully it'll be much easier to test stuff using Water Auth locally

PR #813, opened 2018-09-13
AQCU-1706: remove space between qualifiers so the comparison function

kmschoep-usgs:
can find them all.


mhines-usgs:
works, but tests need updating to remove spaces I added when I made it pretty :/

kmschoep-usgs:
a real improvement will have to wait.  I was quickly out of my R depth when trying to make this more robust.

mhines-usgs (approved):

PR #815, opened 2018-09-25
AQCU-1642: update isEmptyOrBlank to handle vectors properly

mhines-usgs:
It's kind of ugly but it's working.. looking at broken tests now.

My intent was to make sure the entire vector is looked at but still return a TRUE if any item in the vector is empty or blank. I don't know if that's dangerous or not yet. Once I get into the tests I can revise that theory...

### DIFF
--- a/tests/testthat/test-utils-validation.R
+++ b/tests/testthat/test-utils-validation.R
@@ -276,4 +276,41 @@ test_that('fieldExists works as expected', {
   expect_false(repgen:::fieldExists(data, "field3"))    
 })
 
+test_that('isEmptyOrBlankVectors handles single values', {
+  singleValue <- "temp"
+  emptyValue <- ""
+  naValue <- NA
+  
+  expect_false(repgen:::isEmptyOrBlankVectors(singleValue))
+  expect_true(repgen:::isEmptyOrBlankVectors(emptyValue))
+  expect_true(repgen:::isEmptyOrBlankVectors(naValue))
+})
+
+test_that('isEmptyOrBlankVectors handles lists and returns a response for each item in list', {
+  coupleOfValues <- c("a","b","c")
+  response <- repgen:::isEmptyOrBlankVectors(coupleOfValues)
+  expect_equal(response, c(FALSE,FALSE,FALSE))
+  
+  coupleOfDifferentValues <- c("a","","c")
+  response <- repgen:::isEmptyOrBlankVectors(coupleOfDifferentValues)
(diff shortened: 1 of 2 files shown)


mhines-usgs:
example of tests I wrote locally that are handled properly

```
a<-c("this","that","this")
b<-c("this","","that")
c<-"this"
emp.data <- data.frame( emp_name = c("Rick","Dan","Michelle","Ryan","Gary"))
emp_name <- c("Rick","Dan","Michelle","Ryan",NA)
repgen:::isEmptyOrBlank(a)
[1] FALSE
repgen:::isEmptyOrBlank(b)
[1] TRUE
repgen:::isEmptyOrBlank(c)
[1] FALSE
 repgen:::isEmptyOrBlank(emp_name)
[1] TRUE
repgen:::isEmptyOrBlank(emp.data)
[1] FALSE
```

kmschoep-usgs:
I like this!  It's flexible enough so that functions using it can then do its own evaluation of "are some/all/none empty or blank" and handle appropriately.

tests/testthat/test-utils-validation.R
```diff
@@ -276,4 +276,53 @@ test_that('fieldExists works as expected', {
...
+  response <- repgen:::isEmptyOrBlankVectors(exampleDf)
+  expect_equal(response[,1],as.logical(c(FALSE,FALSE,FALSE,FALSE,FALSE,TRUE)))
+  expect_equal(response[,2],as.logical(c(FALSE,FALSE,TRUE,FALSE,TRUE,FALSE)))
+  expect_equal(response[,3],as.logical(c(FALSE,FALSE,FALSE,FALSE,FALSE,TRUE)))
+  expect_s3_class(response,"data.frame")
+  expect_equal(dim(exampleDf),dim(response))
+})
+
+test_that('isEmptyOrBlankVectors handles bug case found and reported in AQCU-1642', {
+  val1 <- c(NA,NULL,1:3)
```
mhines-usgs: this test just jumped out at me.. NULL doesn't seem to be handled right by this function at all? I'll go poke it a little more...

mhines-usgs:
Yeah, I only wonder about the last test there - why is NULL seemingly invisible to this function?

tests/testthat/test-utils-validation.R
mhines-usgs: Seems like NULL values are not allowed in vectors how I've created them here. I'll remove this since it's a little misleading? https://www.r-bloggers.com/r-na-vs-null/

PR #816, opened 2018-09-28
AQCU-1543: transition 5yr; update dvhydro specific functions to be fo…

mhines-usgs:
…r both reports


mhines-usgs:
- [x] seems like uvhydro is using readTimeSeries so still need a temp function to accommodate that split
- [x] also getting weird test failure for five year I'll figure out Monday as well.. test-fiveyeargwsum.R line 49 - test expects red but the value returns blue. a mix up in the min/max logic?

tests/testthat/test-utils-fetch.R
```diff
@@ -409,30 +400,28 @@ test_that("fetchPrimarySeriesApprovals properly retrieves the primary series app
...
   }')
   
-  primary <- repgen:::fetchPrimarySeriesQualifiers(approvals)
+  primary <- repgen:::fetchPrimarySeriesQualifiers(qualifiers)
   
-  expect_is(approvals, 'list')
-  expect_equal(approvals[[1]][['code']], "E")
-  expect_equal(approvals[[1]][['displayName']], "Flow at station affected by ice")
+  expect_is(qualifiers, 'list')
+  expect_equal(qualifiers[[1]][['identifier']], "ESTIMATED")
```
mhines-usgs, line 423: had to change these tests because the data for code and displayName was no longer a part of the primarySeriesQualifiers fetch when those data moved to qualifierMetadata

zmoore-usgs (approved):
This all looks good to me! You're a wizard for keeping all of this stuff straight 😄✨

PR #818, opened 2018-10-01
Aqcu 1544

kmschoep-usgs:


R/sensorreading-data.R
```diff
@@ -63,16 +67,29 @@ formatSensorData <- function(readings, columnNames, includeComments){
...
+    timeFormatted <- timeFormatting(displayTime, "%m/%d/%Y", " ")
+    timeFormattedCorrected <- timeFormatting(nearestCorrectedTime, "%m/%d/%Y", " ")
 
     rec <- getRecorderWithinUncertainty(listElements[["uncertainty"]], listElements[["value"]], listElements[["recorderValue"]])
     ind <- getIndicatedCorrection(listElements[["recorderValue"]], listElements[["value"]])
     app <- getAppliedCorrection(listElements[["nearestRawValue"]], listElements[["nearestCorrectedValue"]])
     corr <- getCorrectedRef(listElements[["value"]], listElements[["nearestCorrectedValue"]], listElements[["uncertainty"]])
     
-    qual <- formatQualifiersStringList(as.data.frame(listElements[["qualifiers"]]))
+    qualifiers <- tryCatch({
```
zmoore-usgs: I think we've been putting try/catch blocks like this into `utils-parse.r` as "parse" functions. So this could be `parseSRSQualifiers`

R/utils-time.R
```diff
@@ -34,6 +35,12 @@ flexibleTimeParse <- function(x, timezone, shiftTimeToNoon = TRUE) {
     time[which.still.NA] <- time[which.still.NA] + hours(12)
   }
   
+  #Apply UTC offset
+  if (applyOffset){
```
zmoore-usgs: I was under the impression that `parse_date_time(x,format_str, tz=timezone,quiet=TRUE)` would parse the times into the timezone specified in the `timezone` argument. So even if the times were provided in UTC if a CST timezone was provided they would be parsed to CST and have the offset applied there automatically. 

Is that not doing that?

R/utils-read.R
```diff
@@ -1233,6 +1233,36 @@ readExcludedControlConditions <- function(reportObject){
...
+	qualifiers <- qualifiersList[[1]]
+	returnList <- list()
+	
+	if(validateFetchedData(qualifiers, 'Qualifiers', requiredFields, stopEmpty=FALSE)){
+		returnList <- qualifiers
+		returnList[['startTime']] <- flexibleTimeParse(returnList[['startTime']], timezone)
+		returnList[['endTime']] <- flexibleTimeParse(returnList[['endTime']], timezone)
+	}
+	
+	#qualifierMetadata <- fetchQualifierMetadata(reportObject)
```
zmoore-usgs: Once it's verified that this line isn't necessary we could probably remove it

mhines-usgs: delete?

R/utils-time.R
kmschoep-usgs: I should probably rename this variable, since it's not really "apply offset" but more "pull out offset for display"

mhines-usgs (approved):

R/utils-time.R
zmoore-usgs: Ahh I see what this is doing now. Gotcha, okay!

Wondering if perhaps that might be better to have in a separate function like `formatTimeForDisplayWithoutOffset` or something?

tests/testthat/test-sensorreadingsummary.R
```diff
@@ -59,53 +114,48 @@ test_that("get unique qualifiers", {
...
+        	"code": "E",
+      	 	"displayName": "Estimated"
+      	}
+			}
+    }
   }')
-   
-  quals <- as.data.frame(reportObject$readings$qualifiers)
+
+  quals <- bind_rows(repgen:::fetchQualifierMetadata(reportObject)[[1]])
```
mhines-usgs, line 158: fancy

mhines-usgs:
This looks good to me (with additional changes just pushed)

PR #819, opened 2018-10-09
AQCU-1733: remove unnecessary code actually impacting whether or not …

mhines-usgs:
…approval bars drew when multiple time series are plotted

And adjust logic of the R markdown to only display the header details when appropriate (excluding discrete and min max)


isuftin:
Weird formatting not in line with the rest of json document. Not a big deal

R/fiveyeargwsum-render.R
```diff
@@ -102,19 +102,6 @@ createfiveyeargwsumPlot <- function(reportObject){
 
   plot_object <- 
     XAxisLabels(plot_object, month_label, month_label_location, date_seq_yr + months(6))
-  
-  if(!isEmptyOrBlank(stat2TimeSeries)) {
```
zmoore-usgs, line 106: Is this stuff handled elsewhere now? Sorry, been a long time since I've looked at this code - just want to make sure I'm understanding what was removed here

mhines-usgs: The plotting of the data is handled by calls below (now lines 106/111/116/121). These lines were setting up the sides and axes but they don't seem to be necessary at all - and were actually shifting the lims incorrectly to push the approval bars below the lower lims.  I noticed because there wasn't a similar line related to the primary time series/stat1TimeSeries, which was displaying approvals properly its own.. so removed these and then magically they all work again..

mhines-usgs:
@isuftin I think I fixed it, but if you still see something funky let me know

R/fiveyeargwsum-render.R
zmoore-usgs: Gotcha, okay makes sense. The only thing I can think of with these is that they might have been for implementing the 3rd y-axis stuff when the additional stat derived series are on other parameters? But as long as that still works I think this looks good :+1:

mhines-usgs: yeah, i think it may have been necessary at some point, maybe prior to a gsplot change? but now it isn't. but here's a screenshot of this code with 4 diff params

image

zmoore-usgs: Awesome! 🎨

zmoore-usgs (approved):

PR #820, opened 2018-10-10
AQCU-1678: rename conditionaldata mapping

mhines-usgs:
prior to this change, caused conditionalFill processor entries in the DC plot to be empty and the legend said 'undefined'


mhines-usgs:
after
image

zmoore-usgs (approved):

PR #828, opened 2018-10-22
WIP Aqcu 1542 - transition extremes to refactored json 

mhines-usgs:


mhines-usgs:
if https://github.com/USGS-R/repgen/pull/829 works, I'll merge that and hopefully this PR will also succeed with Travis

PR #832, opened 2018-10-25
Aqcu 1548 Uvydro refactor

mhines-usgs:


mhines-usgs:
Seems like repgen's Travis is failing because https://github.com/r-lib/devtools/issues/1906#issuecomment-433477711

zmoore-usgs (approved):
Looks good! Nice to see those report-specific methods continuing to disappear 💥

PR #835, opened 2018-10-31
Aqcu 1608- Extremes fixes after refactor

mhines-usgs:
Parse date/times before comparison
Fix tests/test data
Update documentation


R/extremes-data.R
```diff
@@ -501,4 +507,66 @@ completeQualifiers <- function(reportObject) {
...
   
   return(consolidated)
+}
+
+#' Translate date/times
+#' @description Add date/times that is in a format we can use for comparison
+#' @param series The series to translate into format that can be used for comparison
+#' @param timezone The timezone for the data
+#' @return series with new fields for comparison
+translateDateTimes <- function(series, timezone) {
```
kmschoep-usgs, line 517: would this be easier if the dates came through with offsets instead of as instants?  or do DVs have to be handled special regardless?

mhines-usgs: i just need to tell flexibletimeparse whether or not to attach an offset, and we don't want an offset on the daily values, i don't know if any other way would be particularly easier?

kmschoep-usgs: If it's working we can leave it and I'll just live with the guilt.

kmschoep-usgs (approved):

PR #837, opened 2018-11-06
Aqcu 1547 VDiagram refactor

mhines-usgs:


R/vdiagram-data.R
```diff
@@ -70,11 +70,8 @@ createControlConditionsString <- function(controlConditions){
   returnString <- ""
   
   if(!isEmptyOrBlank(controlConditions)){
-    displayValues <- controlConditions[['name']]
-
```
mhines-usgs: hopefully they don't care about this, but if they do, we'll have to get the prettier name back in there maybe

mhines-usgs: actually maybe I can modify this to split on the capital letter ...

tests/testthat/test-utils-parse.R
```diff
@@ -359,7 +359,7 @@ test_that('parseExcludedControlConditions properly retrieves the excluded contro
   
   conditions <- repgen:::parseExcludedControlConditions(controlConditionJSON)
   
-  expect_equal(repgen:::parseExcludedControlConditions(c()), NULL)
+  expect_equal(repgen:::parseExcludedControlConditions(), NULL)
```
mhines-usgs: ah, figured this out. when I changed the utils-read function readExcludedControlConditions() to not use validation since the new excludeConditions data have no required fields (just a char string), the function returns something other than NULL. I changed so that I still use the validateFetchedData function (with no required fields) and this test can be changed back. 😃

kmschoep-usgs (approved):

zmoore-usgs (approved):

PR #840, opened 2018-11-13
Aqcu 1547 revision

mhines-usgs:
handle refactor's lack of arbitrary shiftNumber: 0 in measurements


zmoore-usgs (approved):
Seems like a nice quick fix

PR #852, opened 2019-04-05
Merge docker and code repos.

zmoore-usgs:


Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM cidasdpdasartip.cr.usgs.gov:8447/aqcu/rserve:latest
+
+ENV GSPLOT_VERSION_DEFAULT=0.8.1
+
+COPY inst/extdata/installPackages.R /tmp/install/installPackages.R
```
mhines-usgs, line 5: so we dont have to download it again?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY R /tmp/install/repgen/R
+COPY tests /tmp/install/repgen/test
+COPY .Rbuildignore /tmp/install/repgen/.Rbuildignore
+COPY LICENSE /tmp/install/repgen/LICENSE
+COPY NAMESPACE /tmp/install/repgen/NAMESPACE
+
+# Build and install Repgen Package
+RUN cd /tmp/install/ && R CMD build repgen && \
+  mv /tmp/install/repgen_*.tar.gz /tmp/install/repgen.tar.gz && \
+  Rscript -e "install.packages(\"/tmp/install/repgen.tar.gz\", repos = NULL, type=\"source\")" && \
```
mhines-usgs, line 34: might need to make sure to do the dependencies=F again here, otherwise we'll be getting uncontrolled new dependency versions?

mhines-usgs: Unless I'm not understanding how things have changed from the other Dockerfile

mhines-usgs (approved):
looks good, just a q about how the repgen dependencies are handled, not sure if it keeps the versions we're trying to lock down to, or installs the newest when it gets to that point

mhines-usgs:
Also, one more Q, we're not controlling the repgen version anymore, this will just be latest/master?

Dockerfile
zmoore-usgs: From what I could tell this same script is used for both GSPlot and Repgen, it just reads the packages to install from the DESCRIPTION file in whatever directory it is run from. So instead of downloading this GitHub like we were before it's just pulling it directly from the Repgen source code already on the machine.

zmoore-usgs:
Yeah, so like the other merged repos this just installs Repgen from whatever sources are downloaded with the Dockerfile. So if you clone this down to your machine and build it'll be building from master. if we cut a release and make new changes then to build the old release you'd checkout that tag locally and then just run the docker build since it builds from the current checked-out sources.

If we need to rebuild an old release from before this merge then we'll need to use the old Dockerfile that pulls the code from GitHub.

Dockerfile
zmoore-usgs: Ah, I wasn't sure if that was available for the `install.packages()` command (we were using `install_gitub()` before), but I can try adding it back!

mhines-usgs: Ah, you might be good, because https://www.rdocumentation.org/packages/utils/versions/3.5.3/topics/install.packages says dependencies = whatever is "Not used if repos = NULL. "

so, I guess just make sure that it has the locked-down versions we're hoping for?

mhines-usgs: you can get into R on the container by typing 'R' as the root user on the container.. and then hit enter, and then do a packageVersion("packagename")
checked test and dplyr is 0.7.6 for example, which matches the dplyr snapshot on the mran date https://cran.microsoft.com/snapshot/2018-08-09/

kmschoep-usgs (approved):

zmoore-usgs:
Going to go ahead and merge since Travis seems to give up pulling our Rserve base image. Will look into moving that from our Artifactory to our DockerHub account

PR #854, opened 2019-11-04
Updated Dockerfile to allow building on network. Tweaked extremes…

zmoore-usgs:
… report numeric --> string conversion to include the proper number of decimal places without any trailing 0s.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,20 +4,37 @@ ENV GSPLOT_VERSION_DEFAULT=0.8.1
 
 COPY inst/extdata/installPackages.R /tmp/install/installPackages.R
 
-# Get GSPlot dependency installs
-RUN mkdir -p /tmp/install/gsplot_description_dir && \
-  wget -O /tmp/install/gsplot_description_dir/DESCRIPTION https://raw.githubusercontent.com/USGS-R/gsplot/v${GSPLOT_VERSION:-$GSPLOT_VERSION_DEFAULT}/DESCRIPTION
+RUN mkdir -p /tmp/install/gsplot_description_dir
+RUN mkdir -p /tmp/install/gsplot_source_dir
 
-# Install GSplot
+# Get GSPlot Sources
+RUN if getent ahosts "sslhelp.doi.net" > /dev/null 2>&1; then \
+		wget 'http://sslhelp.doi.net/docs/DOIRootCA2.cer' && \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -32,4 +49,5 @@ COPY NAMESPACE /tmp/install/repgen/NAMESPACE
 RUN cd /tmp/install/ && R CMD build repgen && \
   mv /tmp/install/repgen_*.tar.gz /tmp/install/repgen.tar.gz && \
   Rscript -e "install.packages(\"/tmp/install/repgen.tar.gz\", repos = NULL, type=\"source\")" && \
+  ls /opt/rserve/R_libs | if [ $(grep -c repgen) -eq 0 ]; then echo "Repgen build failed" && exit 1; fi && \
```
zmoore-usgs, line 52: Previously the build would still continue if the Repgen build/install failed because Rscript always returns exit code 0, even if the script it ran exited with an error. This line checks that the package was _actually_ installed before continuing.

Dockerfile
```diff
@@ -4,21 +4,38 @@ ENV GSPLOT_VERSION_DEFAULT=0.8.1
...
   cd /tmp/install/gsplot_description_dir && \
-  Rscript /tmp/install/installPackages.R && \
-  Rscript -e "library(devtools);install_url('https://github.com/USGS-R/gsplot/archive/v${GSPLOT_VERSION:-$GSPLOT_VERSION_DEFAULT}.zip', dependencies = F)"
+  Rscript /tmp/install/installPackages.R
 
 # Install Repgen Dependencies
 COPY DESCRIPTION /tmp/install/repgen/DESCRIPTION
 RUN cd /tmp/install/repgen && Rscript /tmp/install/installPackages.R
 
+# Build and install GSPlot Package
```
zmoore-usgs, line 30: Replaced the `"...install_url(...)..."` version of the GSPlot install with a manual version that works the same way because I couldn't figure out how to inject the cert properly for the R `install_url` command. This accomplishes the same thing as that command does, just externally from R (so that I could use the root cert when needed).

R/extremes-data.R
```diff
@@ -466,8 +466,10 @@ applyQualifiersToValues <- function(points, qualifiers) {
     pointsWithQs$quals <- NULL
     points <- pointsWithQs
   }
-  
-  points$value <- as.character(points$value)
+
+  # Print values with proper number of decimals and no trailing zeros
+  valString <- sprintf("%.14f", points$value)
+  points$value <- str_replace_all(valString, "([0-9].*)(\\.[0-9]+?)0*?$", "\\1\\2")
```
zmoore-usgs: Regex that identifies trailing zeros after a decimal point and removes them.

kmschoep-usgs (approved):

PR #857, opened 2020-02-04
use a flag to determine whether the message for multiple extremes in

kmschoep-usgs:
same dataset occurs.  Fix/add tests.


kmschoep-usgs:
for some reason my roxygen doesn't like to redo the documentation.

zmoore-usgs (approved):
Looks good to me!

PR #858, opened 2020-02-06
fix bug where I look for the multipleMinFlag and MultipleMaxFlag

kmschoep-usgs:
attributes in the wrong place in the reportObject json.


zmoore-usgs (approved):
:+1:

